PR #2, opened 2015-07-04
add basic totp junk. definitely WIP

voldemortensen:
George, is this a start to what you're looking for?


georgestephanis:
Yup, a start.  We should ideally be able to generate the QR codes for Google Authenticator / Authy / etc to scan and add.

I'm also okay architecturally with not including backup codes in each submodule -- so you wouldn't need to include that -- and actually having another provider that is a 'backup codes' provider.

PR #4, opened 2015-07-10
Late-escaping of output, removal of $_REQUEST

stevegrunwell:
We're looking to use this on a platform that champions "escape all the things!", so I've gone through and added late-escaping wherever it was missing, following the [WordPress VIP standards](https://vip.wordpress.com/documentation/code-review-what-we-look-for/) (as they tend to know a thing or two about securing WordPress 😉).

This PR also replaces instances of `$_REQUEST` with `$_POST` or `$_GET`, based on how the data is being received.

Thank you for putting this together!

### DIFF
--- a/class.application-passwords-list-table.php
+++ b/class.application-passwords-list-table.php
@@ -8,10 +8,10 @@ class Application_Passwords_List_Table extends WP_List_Table {
 
 	function get_columns() {
 		return array(
-			'name'      => esc_html__( 'Name', 'two-factor' ),
-			'created'   => esc_html__( 'Created', 'two-factor' ),
-			'last_used' => esc_html__( 'Last Used', 'two-factor' ),
-			'last_ip'   => esc_html__( 'Last IP', 'two-factor' ),
+			'name'      => wp_strip_all_tags( __( 'Name', 'two-factor' ) ),
+			'created'   => wp_strip_all_tags( __( 'Created', 'two-factor' ) ),
+			'last_used' => wp_strip_all_tags( __( 'Last Used', 'two-factor' ) ),
+			'last_ip'   => wp_strip_all_tags( __( 'Last IP', 'two-factor' ) ),
 		);
 	}
 
--- a/class.application-passwords-list-table.php
+++ b/class.application-passwords-list-table.php
@@ -1,21 +1,41 @@
 <?php
-
+// Load the parent class if it doesn't exist.
 if ( ! class_exists( 'WP_List_Table' ) ) {
 	require_once ABSPATH . 'wp-admin/includes/class-wp-list-table.php';
 }
 
+/**
+ * Class for displaying the list of application password items.
+ *
+ * @since 0.1-dev
+ * @access private
+ *
+ * @package Two_Factor
+ */
 class Application_Passwords_List_Table extends WP_List_Table {
 
-	function get_columns() {
+	/**
+	 * Get a list of columns.
+	 *
+	 * @since 0.1-dev
+	 *
+	 * @return array
+	 */
+	public function get_columns() {
 		return array(
-			'name'      => wp_strip_all_tags( __( 'Name', 'two-factor' ) ),
-			'created'   => wp_strip_all_tags( __( 'Created', 'two-factor' ) ),
-			'last_used' => wp_strip_all_tags( __( 'Last Used', 'two-factor' ) ),
-			'last_ip'   => wp_strip_all_tags( __( 'Last IP', 'two-factor' ) ),
+			'name'      => wp_strip_all_tags( __( 'Name' ) ),
+			'created'   => wp_strip_all_tags( __( 'Created' ) ),
+			'last_used' => wp_strip_all_tags( __( 'Last Used' ) ),
+			'last_ip'   => wp_strip_all_tags( __( 'Last IP' ) ),
 		);
 	}
 
-	function prepare_items() {
+	/**
+	 * Prepares the list of items for displaying.
+	 *
+	 * @since 0.1-dev
+	 */
+	public function prepare_items() {
 		$columns  = $this->get_columns();
 		$hidden   = array();
 		$sortable = array();
@@ -23,8 +43,17 @@ class Application_Passwords_List_Table extends WP_List_Table {
 		$this->_column_headers = array( $columns, $hidden, $sortable, $primary );
 	}
 
-	function column_default( $item, $column_name ) {
-		switch( $column_name ) {
+	/**
+	 * Generates content for a single row of the table
+	 *
+	 * @since 0.1-dev
+	 * @access protected
+	 *
+	 * @param object $item The current item.
+	 * @param string $column_name The current column name.
+	 */
+	protected function column_default( $item, $column_name ) {
+		switch ( $column_name ) {
 			case 'name':
 				$actions = array(
 					'delete' => Application_Passwords::delete_link( $item ),
@@ -32,17 +61,17 @@ class Application_Passwords_List_Table extends WP_List_Table {
 				return esc_html( $item['name'] ) . self::row_actions( $actions );
 			case 'created':
 				if ( empty( $item['created'] ) ) {
-					return esc_html__( 'Unknown', 'two-factor' );
+					return esc_html__( 'Unknown' );
 				}
 				return date( get_option( 'date_format', 'r' ), $item['created'] );
 			case 'last_used':
 				if ( empty( $item['last_used'] ) ) {
-					return esc_html__( 'Never', 'two-factor' );
+					return esc_html__( 'Never' );
 				}
 				return date( get_option( 'date_format', 'r' ), $item['last_used'] );
 			case 'last_ip':
 				if ( empty( $item['last_ip'] ) ) {
-					return esc_html__( 'Never Used', 'two-factor' );
+					return esc_html__( 'Never Used' );
 				}
 				return $item['last_ip'];
 			default:
@@ -51,7 +80,12 @@ class Application_Passwords_List_Table extends WP_List_Table {
 	}
 
 	/**
-	 * Pull into the child class to prevent conflicting nonces.
+	 * Generates custom table navigation to prevent conflicting nonces.
+	 *
+	 * @since 0.1-dev
+	 * @access protected
+	 *
+	 * @param string $which The location of the bulk actions: 'top' or 'bottom'.
 	 */
 	protected function display_tablenav( $which ) {
 		?>
@@ -70,6 +104,13 @@ class Application_Passwords_List_Table extends WP_List_Table {
 		<?php
 	}
 
+	/**
+	 * Generates content for a single row of the table.
+	 *
+	 * @since 0.1-dev
+	 *
+	 * @param object $item The current item.
+	 */
 	public function single_row( $item ) {
 		echo '<tr data-slug="' . esc_attr( Application_Passwords::password_unique_slug( $item ) ) . '">';
 		$this->single_row_columns( $item );
--- a/class.application-passwords-list-table.php
+++ b/class.application-passwords-list-table.php
@@ -8,10 +8,10 @@ class Application_Passwords_List_Table extends WP_List_Table {
 
 	function get_columns() {
 		return array(
-			'name'      => __( 'Name', 'two-factor' ),
-			'created'   => __( 'Created', 'two-factor' ),
-			'last_used' => __( 'Last Used', 'two-factor' ),
-			'last_ip'   => __( 'Last IP', 'two-factor' ),
+			'name'      => esc_html__( 'Name', 'two-factor' ),
+			'created'   => esc_html__( 'Created', 'two-factor' ),
+			'last_used' => esc_html__( 'Last Used', 'two-factor' ),
+			'last_ip'   => esc_html__( 'Last IP', 'two-factor' ),
 		);
 	}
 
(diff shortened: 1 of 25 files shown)


stevegrunwell:
Hmm, seem to be encountering an infinite login loop when I hit mysite.com/wp-login.php directly; will troubleshoot and have a patch shortly, as it may be connected to my `$_REQUEST` adjustments.

stevegrunwell:
That's exactly what it was, the 'redirect_to' value was getting lost, resulting in an infinite login loop.

providers/class.two-factor-email.php
```diff
@@ -64,11 +64,11 @@ function authentication_page( $user ) {
 			}, 200);
 		</script>
 		<?php
-		submit_button( __( 'Log In', 'two-factor' ) );
+		submit_button( esc_html__( 'Log In', 'two-factor' ) );
 	}
 
 	function validate_authentication( $user ) {
-		return $this->validate_token( $user->ID, $_REQUEST['two-factor-email-code'] );
+		return $this->validate_token( $user->ID, $_POST['two-factor-email-code'] );
```
georgestephanis: This I had intentionally left as $_REQUEST in case we later modify the emails to include a link they can follow with the $_GET argument pre-filled.

providers/class.two-factor-email.php
```diff
@@ -64,11 +64,11 @@ function authentication_page( $user ) {
 			}, 200);
 		</script>
 		<?php
-		submit_button( __( 'Log In', 'two-factor' ) );
+		submit_button( esc_html__( 'Log In', 'two-factor' ) );
```
georgestephanis: You're escaping this too early.  The string should be passed in to `submit_button()` unescaped, as it will escape it in a contextually correct way when printing.

https://github.com/WordPress/WordPress/blob/c6cff377e7aa9cf566a7511e54a470671cf5088a/wp-admin/includes/template.php#L1956

providers/class.two-factor-email.php
```diff
@@ -51,7 +51,7 @@ function authentication_page( $user ) {
 		<p><?php esc_html_e( 'A verification code has been sent to the email address associated with your account.', 'two-factor' ); ?></p>
 		<p>
 			<label for="authcode"><?php esc_html_e( 'Verification Code:' ); ?></label>
-			<input type="tel" name="two-factor-email-code" id="authcode" class="input" value="" size="20" pattern="[0-9]*" />
+			<input type="text" name="two-factor-email-code" id="authcode" class="input" value="" size="20" pattern="[0-9]*" />
```
georgestephanis: Leaving this as `tel` results in a numpad input on mobile.  If you're going to change the input type, please change to an input type that keeps this ux.

providers/class.two-factor-email.php
```diff
@@ -39,8 +39,8 @@ function delete_token( $user_id ) {
 	function generate_and_email_token( $user ) {
 		$token = $this->generate_token( $user->ID );
 
-		$subject = sprintf( __( 'Your login confirmation code for %s', 'two-factor' ), get_bloginfo( 'name' ) );
-		$message = sprintf( __( 'Enter %s to log in.', 'two-factor' ), $token );
+		$subject = sprintf( esc_html__( 'Your login confirmation code for %s', 'two-factor' ), get_bloginfo( 'name' ) );
+		$message = sprintf( esc_html__( 'Enter %s to log in.', 'two-factor' ), $token );
```
georgestephanis: `wp_mail()` by default does not use an HTML content type, it sends as plain text.  As such, escaping it with an HTML context would be inaccurate and potentially break things.

providers/class.two-factor-email.php
```diff
@@ -39,8 +39,8 @@ function delete_token( $user_id ) {
 	function generate_and_email_token( $user ) {
 		$token = $this->generate_token( $user->ID );
 
-		$subject = sprintf( __( 'Your login confirmation code for %s', 'two-factor' ), get_bloginfo( 'name' ) );
-		$message = sprintf( __( 'Enter %s to log in.', 'two-factor' ), $token );
+		$subject = sprintf( esc_html__( 'Your login confirmation code for %s', 'two-factor' ), get_bloginfo( 'name' ) );
```
georgestephanis: Subject lines don't normally support html, so content shouldn't be escaped as such?

providers/class.two-factor-dummy.php
```diff
@@ -20,7 +20,7 @@ function authentication_page( $user ) {
 		?>
 		<p><?php esc_html_e( 'Are you really you?', 'two-factor' ); ?></p>
 		<?php
-		submit_button( __( 'Yup.', 'two-factor' ) );
+		submit_button( esc_html__( 'Yup.', 'two-factor' ) );
```
georgestephanis: As noted below -- https://github.com/georgestephanis/two-factor/pull/4/files#r34511775 -- this shouldn't be escaped here.

class.application-passwords-list-table.php
```diff
@@ -8,10 +8,10 @@ class Application_Passwords_List_Table extends WP_List_Table {
...
 	function get_columns() {
 		return array(
-			'name'      => __( 'Name', 'two-factor' ),
-			'created'   => __( 'Created', 'two-factor' ),
-			'last_used' => __( 'Last Used', 'two-factor' ),
-			'last_ip'   => __( 'Last IP', 'two-factor' ),
+			'name'      => esc_html__( 'Name', 'two-factor' ),
+			'created'   => esc_html__( 'Created', 'two-factor' ),
+			'last_used' => esc_html__( 'Last Used', 'two-factor' ),
+			'last_ip'   => esc_html__( 'Last IP', 'two-factor' ),
```
georgestephanis: These columns probably shouldn't be escaped here, as you're providing the content, that the list table class can display in many different methods -- for example, as an attribute like so:

https://github.com/WordPress/WordPress/blob/b72f822a69ad634488b5a57aea99f0a78e42f777/wp-admin/includes/class-wp-list-table.php#L1213

Maybe calling `wp_strip_all_tags()` on it, or `wp_kses()` -- but escaping without knowing the usage context here feels wrong.

class.application-passwords-list-table.php
```diff
@@ -71,7 +71,7 @@ protected function display_tablenav( $which ) {
 	}
 
 	public function single_row( $item ) {
-		echo '<tr data-slug="' . Application_Passwords::password_unique_slug( $item ) . '">';
+		echo '<tr data-slug="' . esc_attr( Application_Passwords::password_unique_slug( $item ) ) . '">';
```
georgestephanis: Good catch.  I don't know if it will ever be useful in practice, as the content will only ever be a substr of a md5 hash, but it's still a good practice to be in.

class.application-passwords.php
```diff
@@ -86,7 +86,11 @@ public static function show_user_profile( $user ) {
 
 			<?php if ( $new_password ) : ?>
 			<p class="new-application-password">
-				<?php printf( __( 'Your new password for <strong>%s</strong> is <kbd>%s</kbd>.', 'two-factor' ), esc_html( $new_password_name ), self::chunk_password( $new_password ) ); ?>
+				<?php printf(
+					esc_html_x( 'Your new password for %1$s is %2$s.', 'application, password', 'two-factor' ),
+					'<strong>' . esc_html( $new_password_name ) . '</strong>',
+					'<kbd>' . self::chunk_password( $new_password ) . '</kbd>'
```
georgestephanis: Since we're doing it anyways, should we also pass an `esc_html()` around the result in this row?

class.application-passwords.php
```diff
@@ -119,8 +123,8 @@ public static function catch_submission( $user_id ) {
 
 	public static function catch_delete_application_password() {
 		$user_id = get_current_user_id();
-		if ( ! empty( $_REQUEST['delete_application_password'] ) ) {
-			$slug = $_REQUEST['delete_application_password'];
+		if ( ! empty( $_GET['delete_application_password'] ) ) {
```
georgestephanis: Depending on how the AJAX processing endpoints are written, this may be better left as $_REQUEST to let one endpoint do both.  Maybe.  I'd need to see it built out first.

class.application-passwords.php
```diff
@@ -172,7 +176,7 @@ public static function delete_link( $item ) {
 		$slug = self::password_unique_slug( $item );
 		$delete_link = add_query_arg( 'delete_application_password', $slug );
 		$delete_link = wp_nonce_url( $delete_link, "delete_application_password-{$slug}", '_nonce_delete_application_password' );
-		return sprintf( '<a href="%1$s">%2$s</a>', esc_url( $delete_link ), esc_html__( 'Delete', 'two-factor' ) );
+		return sprintf( '<a href="%s">%s</a>', esc_url( $delete_link ), esc_html__( 'Delete', 'two-factor' ) );
```
georgestephanis: No, the numbering was intentional.  Should the structure ever get changed around, it helps to make the sequence of arguments obvious.

class.two-factor-core.php
```diff
@@ -122,10 +122,10 @@ function show_two_factor_login( $user ) {
 
 		$login_nonce = $this->create_login_nonce( $user->ID );
 		if ( ! $login_nonce ) {
-			wp_die( __( 'Could not save login nonce.', 'two-factor' ) );
+			wp_die( esc_html__( 'Could not save login nonce.', 'two-factor' ) );
 		}
 
-		$redirect_to = isset( $_REQUEST['redirect_to'] ) ? $_REQUEST['redirect_to'] : $_SERVER['REQUEST_URI'];
+		$redirect_to = isset( $_POST['redirect_to'] ) ? $_POST['redirect_to'] : $_SERVER['REQUEST_URI'];
```
georgestephanis: Leave this as $_REQUEST -- it's necessary if switching to a different provider on a branch I'm working on.

class.two-factor-core.php
```diff
@@ -134,7 +134,7 @@ function login_html( $user, $login_nonce, $redirect_to, $error_msg = '', $login_
 		$provider = $this->get_provider_for_user( $user->ID );
 
 		$rememberme = 0;
-		if ( isset ( $_REQUEST[ 'rememberme' ] ) && $_REQUEST[ 'rememberme' ] ) {
+		if ( isset ( $_POST[ 'rememberme' ] ) && $_POST[ 'rememberme' ] ) {
```
georgestephanis: Leave this as $_REQUEST -- it's necessary if switching to a different provider on a branch I'm working on.

class.two-factor-core.php
```diff
@@ -148,14 +148,14 @@ function login_html( $user, $login_nonce, $redirect_to, $error_msg = '', $login_
...
-				<input type="hidden" name="redirect_to" id="redirect_to" value="<?php echo esc_attr( $redirect_to ) ?>"/>
+				<input type="hidden" name="redirect_to" id="redirect_to" value="<?php echo esc_url( $redirect_to ) ?>"/>
 				<input type="hidden" name="rememberme" id="rememberme" value="<?php echo esc_attr( $rememberme ) ?>"/>
 
 				<?php $provider->authentication_page( $user ); ?>
 
 		</form>
 
-		<p id="backtoblog"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" title="<?php esc_attr_e( 'Are you lost?' ); ?>"><?php printf( __( '&larr; Back to %s' ), get_bloginfo( 'title', 'display' ) ); ?></a></p>
+		<p id="backtoblog"><a href="<?php echo esc_url( home_url( '/' ) ); ?>" title="<?php esc_attr_e( 'Are you lost?' ); ?>"><?php printf( esc_html__( '&larr; Back to %s', 'two-factor' ), get_bloginfo( 'title', 'display' ) ); ?></a></p>
```
georgestephanis: This was intentionally using the core string.

https://github.com/WordPress/WordPress/blob/04793dce341e474dc16429150ee2a318df1ad305/wp-login.php#L219

georgestephanis:
I skipped a few remarks at the end between $_REQUEST and $_POST -- but I'd prefer if for the moment anyways, we could leave all of those as $_REQUEST until I get the backup methods branch sorted.

providers/class.two-factor-email.php
stevegrunwell: Using `input[type="text"]` with the numeric pattern [should be enough to trigger the numeric keypad](http://bradfrost.com/blog/mobile/better-numerical-inputs-for-mobile-forms/) (at the bottom of the article Brad points out the pattern attribute is enough to trigger the behavior), at least on iOS, though it would be beneficial to [do further device testing using the "Text pattern=[0-9]" input here](http://bradfrost.com/demo/tel/).

Screenshot of a text input with the pattern attribute in Xcode's iOS simulator (iPhone 6, iOS 8.2):

ios simulator screen shot jul 13 2015 5 27 45 pm

class.application-passwords-list-table.php
stevegrunwell: Ah, good call.

stevegrunwell:
@georgestephanis I've gone through and addressed your feedback on the PR, thank you for being so thorough! I've also reverted the `$_REQUEST`-related changes, but opened #6 so it doesn't get lost in the historical shuffle.

class.two-factor-core.php
```diff
@@ -217,7 +217,7 @@ function login_form_twostep() {
 				return;
 			}
 
-			$this->login_html( $user, $login_nonce, $_REQUEST['redirect_to'], __( 'ERROR: Invalid verification code.', 'two-factor' ) );
+			$this->login_html( $user, $login_nonce, $_REQUEST['redirect_to'], esc_html__( 'ERROR: Invalid verification code.', 'two-factor' ) );
```
georgestephanis: This is escaped inside `login_html()` here: https://github.com/georgestephanis/two-factor/blob/41bed06cc04ec0683ce34ac826cc8a29dd2e5790/class.two-factor-core.php#L144 -- doesn't need double escaping.

stevegrunwell:
Can you tell I've been burned by under-escaping before? ;)

georgestephanis:
:+1:

valendesigns:
I'm not sure if this PR is still needed anymore. There are aspects of it that made it into Master already. @stevegrunwell would you like to fix the merge conflicts so we can see what's different, or should we close the PR?

stevegrunwell:
@valendesigns It looks like the only thing left in the PR that hasn't been merged is converting the two-factor-input-code input from `type="tel"` to `type="text"`, as George and I [discussed earlier in the thread](#diff-for-comment-34511873). I'd say it's safe to close the PR :)

stevenkword:
We'll want to make sure to include the backup codes provider if we take another look at `$_REQUEST`.

PR #5, opened 2015-07-13
Add the 'two_factor_provider_for_user' filter

stevegrunwell:
This filter, added to `Two_Factor_Core::get_provider_for_user()` serves two purposes:

1. Enables developers to set a default two-factor provider (e.g. "if the user hasn't explicitly made a decision default to `$provider`")
2. Enables developers to **force** a specific method when necessary.

### DIFF
--- a/class.two-factor-core.php
+++ b/class.two-factor-core.php
@@ -88,7 +88,7 @@ class Two_Factor_Core {
 		 * @param string $provider The provider currently being used.
 		 * @param int    $user_id  The user ID.
 		 */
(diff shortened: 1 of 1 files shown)


georgestephanis:
A lot of the ambiguity here will be resolved as soon as the https://github.com/georgestephanis/two-factor/tree/add/fallback-methods branch merges in.  But change it to `two_factor_primary_provider_for_user` and I'll merge it in to that branch.

stevegrunwell:
The filter name has been updated.

georgestephanis:
Merged into the `add/fallback-methods` branch, it will merge to master with that.

PR #8, opened 2015-07-17
Code Standards & Testing Tools

valendesigns:
I've added Sniffs, PHPUnit, Travis CI, & Coveralls. I fixed all the build errors and added documentation, please review the docs from the perspective of someone who knows the code before merging. 

Additionally, I added a couple basic unit tests to kick things off. Let me know if you have any questions or want me to change something. You should probably set things up for this repo over at travis-ci.org and coveralls.io before merging. 

I would suggest doing an audit on all the methods in `Application_Passwords`. I don't believe they all need to be `static`. As well, we should be standardizing the class names by prefixing all of them with `Two_Factor_` so they can easily be changed to `WP_` in the future.

Something that stands out as being odd is the way your verifying nonces. I had to add an exclusion so the coding standards didn't complain about it.


valendesigns:
@georgestephanis I've merged master and fix the build. Please review the PR and merge. I can merge it if you want but I think it would make sense that you look over it first. Also, have you activated Travis CI?

georgestephanis:
Looks reasonable.  A couple minor whitespace issues, but I'll tidy those up after merging.  Nothing to fret about.

valendesigns:
You didn't turn travis on before merging.

PR #14, opened 2015-07-31
Adds Backup Codes Provider

stevenkword:
Adds the backup-codes provider to the list of potential methods.  Verification codes are single use and stored in the database as an array of hashed string.  Let me know if you have any questions.

### DIFF
--- a/providers/class.two-factor-backup-codes.php
+++ b/providers/class.two-factor-backup-codes.php
@@ -31,7 +31,7 @@ class Two_Factor_Backup_Codes extends Two_Factor_Provider {
 		}
 		// Only show this if the provider is enabled
 		$enabled_providers = get_user_meta( $user_id, Two_Factor_Core::ENABLED_PROVIDERS_USER_META_KEY, true );
-		if( ! in_array( 'Two_Factor_Backup_Codes', $enabled_providers ) ) {
+		if( ! in_array( __CLASS__, $enabled_providers ) ) {
 			return;
 		}
 
--- a/providers/class.two-factor-backup-codes.php
+++ b/providers/class.two-factor-backup-codes.php
@@ -59,7 +59,7 @@ class Two_Factor_Backup_Codes extends Two_Factor_Provider {
 
 		// Exit if we are not out of codes.
 		if ( 0 < self::codes_remaining_for_user( $user ) ) {
-			return;
+		//	return;
 		}
 
 		// Only show when the provider is enabled.
@@ -70,7 +70,7 @@ class Two_Factor_Backup_Codes extends Two_Factor_Provider {
 		<div class="error">
 			<p>
 				<span><?php esc_html_e( 'Two-Factor: You are out of backup codes and need to ' ); ?><span>
-				<a href="<?php echo esc_url( get_edit_user_link( $user->ID ) ); ?>#two-factor-backup-codes">regenerate!</a>
+				<a href="<?php echo esc_url( get_edit_user_link( $user->ID ) . '#two-factor-backup-codes' ); ?>">regenerate!</a>
 			</p>
 		</div>
 		<?php
--- a/providers/class.two-factor-backup-codes.php
+++ b/providers/class.two-factor-backup-codes.php
@@ -144,21 +144,19 @@ class Two_Factor_Backup_Codes extends Two_Factor_Provider {
 						url: ajaxurl,
 						data: {
 							action: 'two_factor_backup_codes_generate',
+							user_id: '<?php echo esc_js( $user_id ); ?>',
 							nonce: '<?php echo esc_js( $ajax_nonce ); ?>'
 						},
 						dataType: 'JSON',
 						success: function( data ) {
 							$( '.two-factor-backup-codes-wrapper' ).show();
 							$( '.two-factor-backup-codes-unused-codes' ).html( '' );
-							$.each( data, function( key, val ) {
+
+							$.each( data.data, function( key, val ) {
 								$( '.two-factor-backup-codes-unused-codes' ).append( '<li>' + val + '</li>' );
 							} );
 							// Update counter.
 							$( '.two-factor-backup-codes-count' ).html( data.length );
-						},
-						error: function( errorThrown ) {
-							alert( 'error' ); // @todo: remove for production
-							console.log( errorThrown );
 						}
 					} );
 				} );
@@ -198,10 +196,9 @@ class Two_Factor_Backup_Codes extends Two_Factor_Provider {
 	 */
 	public function ajax_generate_json() {
 		check_ajax_referer( 'two-factor-backup-codes-generate-json', 'nonce' );
-
-		$user_id = get_current_user_id();
+		$user_id = sanitize_text_field( $_REQUEST['user_id'] );
 		$codes = $this->generate_codes( $user_id );
-		wp_send_json( $codes );
+		wp_send_json_success( $codes );
 	}
 
 	/**
--- a/providers/class.two-factor-backup-codes.php
+++ b/providers/class.two-factor-backup-codes.php
@@ -1,0 +1,218 @@
+<?php
+
+class Two_Factor_Backup_Codes extends Two_Factor_Provider {
+
+	const DEBUG = true;
+	const BACKUP_CODES_META_KEY = '_two_factor_backup_codes';
+	const BACKUP_CODES_DEBUG_META_KEY = '_two_factor_backup_codes_debug';
+	const NUMBER_OF_CODES = 10;
+
+	static function get_instance() {
+		static $instance;
+		$class = __CLASS__;
+		if ( ! is_a( $instance, $class ) ) {
+			$instance = new $class;
+		}
+		return $instance;
+	}
+
+	public static function add_hooks() {
+		$user_id = get_current_user_id();
+		$primary_provider = get_user_meta( $user_id, Two_Factor_Core::PROVIDER_USER_META_KEY, true );
+
+		// Only do these things when Backup Codes are selected as the primary provider.
+		if( 'Two_Factor_Backup_Codes' != $primary_provider ) {
+			return;
+		}
+
+		add_action( 'admin_notices', array( __CLASS__, 'admin_notices' ) );
+		add_action( 'user_two_factor_options', array( __CLASS__, 'user_two_factor_options' ) );
+
+		add_action( 'wp_ajax_two_factor_backup_codes_generate', array( __CLASS__, 'ajax_two_factor_backup_codes_generate' ) );
+	}
+
+	// @todo add nonce
+	public static function ajax_two_factor_backup_codes_generate() {
+		check_ajax_referer( 'two-factor-backup-codes-generate', 'nonce' );
+
+		$user_id = get_current_user_id();
+		$codes = self::get_instance()->generate_codes( $user_id );
+		$json_codes = json_encode( $codes );
+		echo $json_codes;
+		die(0);
+	}
+
+	public static function admin_notices() {
+		$user_id = get_current_user_id();
+
+		// Only show this notice if we are out of backup codes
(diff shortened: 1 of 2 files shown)


valendesigns:
This should be merged after #8

valendesigns:
OK, I've cleaned up the PR and we now have a passing build. Should I merge it or is there still work to be done?

providers/class.two-factor-backup-codes.php
```diff
@@ -0,0 +1,274 @@
...
+			return;
+		}
+
+		// Only show when the provider is enabled.
+		if ( false === Two_Factor_Backup_Codes::get_instance()->is_enabled( $user_id ) ) {
+			return;
+		}
+		?>
+		<div class="error">
+			<p><?php echo wp_kses( sprintf( __( 'Two-Factor: You are out of backup codes and need to <a href="%1$s#two-factor-backup-codes" >regenerate</a>!' ), esc_url( get_edit_user_link( $user_id ) ) ), array( 'a' => array( 'href' => array() ) ) ); ?></p>
```
georgestephanis: wp_kses'ing a translation feels a bit of overkill to me, and far and beyond what Core ever does with translations.

providers/class.two-factor-backup-codes.php
```diff
@@ -0,0 +1,274 @@
...
+
+	/**
+	 * Verify the provider is enabled.
+	 *
+	 * @since 0.1-dev
+	 *
+	 * @param int $user_id The logged-in user ID.
+	 */
+	public function is_enabled( $user_id ) {
+		$enabled_providers = (array) get_user_meta( $user_id, Two_Factor_Core::ENABLED_PROVIDERS_USER_META_KEY, true );
```
georgestephanis: This feels like it'd be better done as a call to `Two_Factor_Core::get_enabled_providers_for_user();` instead of searching for the meta directly.

https://github.com/georgestephanis/two-factor/blob/d2bfdfc66e0d5f266e2d6e4d41974292278f7e8e/class.two-factor-core.php#L88-L107

providers/class.two-factor-backup-codes.php
```diff
@@ -0,0 +1,274 @@
...
+	public static function admin_notices() {
+		// Only show this notice if we are out of backup codes.
+		$user_id = get_current_user_id();
+		$backup_codes = get_user_meta( $user_id, self::BACKUP_CODES_META_KEY, true );
+		if ( ! empty( $backup_codes ) ) {
+			return;
+		}
+
+		// Only show when the provider is enabled.
+		if ( false === Two_Factor_Backup_Codes::get_instance()->is_enabled( $user_id ) ) {
```
georgestephanis: Why are we instantiating this class just to call what could just as easily be a static method?

I'd sooner see `is_enabled()` made static, so we can just call `self::is_enabled()`

providers/class.two-factor-backup-codes.php
```diff
@@ -0,0 +1,274 @@
...
+	/**
+	 * Whether this Two Factor provider is configured and available for the user specified.
+	 *
+	 * @since 0.1-dev
+	 *
+	 * @param WP_User $user WP_User object of the logged-in user.
+	 * @return boolean
+	 */
+	public function is_available_for_user( $user ) {
+		return true;
```
georgestephanis: This should be fleshed out to return whether the specified user has any unused codes saved in the database.

providers/class.two-factor-backup-codes.php
```diff
@@ -0,0 +1,274 @@
...
+			<p class="description"><?php esc_html_e( "Write 'em down y'all!" ); ?></p>
+		</div>
+		<script type="text/javascript">
+			// @todo: move this into a JS file & internationalize the count strings
+			jQuery( document ).ready( function( $ ) {
+				$( '.button-two-factor-backup-codes-generate' ).click( function() {
+					jQuery.ajax( {
+						url: ajaxurl,
+						data: {
+							action: 'two_factor_backup_codes_generate',
```
georgestephanis: This should also pass in the user id and have the nonce salted with the user id.  Currently, the ajax response method ( `self::ajax_generate_json()` ) assumes user id === `get_current_user_id()` -- but that may not be the case, for example, if an administrator needs to generate a single code for a contributor that needs to authenticate.

providers/class.two-factor-backup-codes.php
```diff
@@ -0,0 +1,274 @@
...
+	 * Generates a JSON object of backup codes.
+	 *
+	 * @since 0.1-dev
+	 */
+	public function ajax_generate_json() {
+		check_ajax_referer( 'two-factor-backup-codes-generate-json', 'nonce' );
+
+		$user_id = get_current_user_id();
+		$codes = $this->generate_codes( $user_id );
+		wp_send_json( $codes );
```
georgestephanis: Should this use `wp_send_json_success()` instead?

providers/class.two-factor-backup-codes.php
```diff
@@ -0,0 +1,274 @@
...
+
+	/**
+	 * Generates a JSON object of backup codes.
+	 *
+	 * @since 0.1-dev
+	 */
+	public function ajax_generate_json() {
+		check_ajax_referer( 'two-factor-backup-codes-generate-json', 'nonce' );
+
+		$user_id = get_current_user_id();
```
georgestephanis: As noted above, this should take the `$user_id` as an argument, should the current user have the edit other users cap and can generate tokens for them as well.

providers/class.two-factor-backup-codes.php
```diff
@@ -0,0 +1,274 @@
...
+	 *
+	 * @since 0.1-dev
+	 *
+	 * @param int $user_id The logged-in user ID.
+	 */
+	public function generate_codes( $user_id ) {
+		$codes = array();
+		$codes_hashed = array();
+
+		for ( $i = 0; $i < self::NUMBER_OF_CODES; $i++ ) {
```
georgestephanis: I almost wonder if it may be tidier to do an `array_fill()` and then `array_map()` to populate the array instead of a for loop.  Probably fine to leave as is.

providers/class.two-factor-backup-codes.php
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	/**
+	 * Generates backup codes & updates the user meta.
+	 *
+	 * @since 0.1-dev
+	 *
+	 * @param int $user_id The logged-in user ID.
+	 */
+	public function generate_codes( $user_id ) {
```
georgestephanis: I'd sooner see this function accept a second `$args` param with the number of codes requested, and whether to append to or replace existing codes, to keep it flexible for generating single serve codes down the road, should an admin need to make one on the fly for a user.

providers/class.two-factor-backup-codes.php
```diff
@@ -0,0 +1,274 @@
...
+	 * @param int $user_id The logged-in user ID.
+	 * @param int $code    The backup code.
+	 * @return boolean
+	 */
+	public function validate_code( $user_id, $code ) {
+		$backup_codes = get_user_meta( $user_id, self::BACKUP_CODES_META_KEY, true );
+
+		foreach ( $backup_codes as $code_index => $backup_code ) {
+			if ( wp_check_password( $code, $backup_code, $user_id ) ) {
+				$this->delete_code( $user_id, $code_index );
```
georgestephanis: I'd sooner see this delete by the `$backup_code` hashed version than array index.  Array index feels like a relatively arbitrary index to use.

georgestephanis:
Just dropped all my thoughts on -- most are trivial and I'm totally fine if y'all decide no, my only really serious concern is the ability to let an admin generate codes for another user that they have editing privileges on.

providers/class.two-factor-backup-codes.php
stevenkword: Good call, I'll make this change.

stevenkword:
@valendesigns Would you mind taking a look at the tests that are failing here and offer some guidance? I see its being caused by JSHint, but I'm not sure where.

valendesigns:
@stevenkword If you take a look at https://travis-ci.org/georgestephanis/two-factor/jobs/76230136 you can see that it's failing for mostly coding standards issues like docs and some spacing issues. Take care of those and I'll sort the escaping ones out tomorrow.

stevenkword:
I think I've sorted it out -- thanks!

valendesigns:
Awesome, good deal! I'll take a look tomorrow and give feedback or make any mods if needed.

stevenkword:
Made a few small updates, but I'm going to try to stay hands-off until you guys have a chance to take another pass.

valendesigns:
I'll have some time later this afternoon to do a pass on it. How far along is the PR, where do you thing we are in terms of being functionally complete? Are there outstanding issues that need to be addressed and if so what are they?

stevenkword:
Functionally, the main piece missing here is going to be some mechanism for implementing a "print these codes" modal / new page opposed to the simple AJAX response we are displaying in the user options now. However, that piece is largely a UX issue.

Regarding bugs, I'm currently aware of two -- one of which relates to all providers.

1.) When you prompted to enter a code during login, if the code is invalid and the error is triggered, the page reloads with the default provider shown instead of the one you used to submit the request.

2.) There is still a singular / plural issue that happens after the AJAX request comes back during the generate process of user options.  Currently only the count updates and not the related text strings.

Other that that I'm starting to feel pretty good about it.

valendesigns:
Point 1 seems out of scope for this PR, so we'll leave it for now. Though an issue should be created to track it's progress.

To address point 2, we could pass the strings via JSON or create an i18n JS object, or really any number of things here to solve that. Which might actually make escaping a lot simpler and less need for all the markup. I notice it as well when I was doing WPCS changes a few weeks ago. I'll tackle that in my next pass.

valendesigns:
The is one thing I see that may need some modification is the string you see when you generate new codes. It's not bad, just a bit informal for a very important aspect of this provider. We should probably change it to truly depict the importance of storing the backup codes for later use.

valendesigns:
@stevenkword & @georgestephanis Any last minute objections before I merge?

providers/class.two-factor-backup-codes.php
```diff
@@ -0,0 +1,297 @@
...
+	 */
+	public static function admin_notices() {
+		$user = wp_get_current_user();
+
+		// Exit if the provider is not enabled.
+		if ( ! in_array( __CLASS__, Two_Factor_Core::get_enabled_providers_for_user( $user->ID ) ) ) {
+			return;
+		}
+
+		// Exit if we are not out of codes.
```
georgestephanis: return if we're not out of codes, exit is a dangerous word to use.

providers/class.two-factor-backup-codes.php
```diff
@@ -0,0 +1,297 @@
...
+			<span class="two-factor-backup-codes-count"><?php echo esc_html( sprintf( _n( '%s unused code remaining.', '%s unused codes remaining.', $count ), $count ) ); ?></span>
+		</p>
+		<div class="two-factor-backup-codes-wrapper" style="display:none;">
+			<ol class="two-factor-backup-codes-unused-codes"></ol>
+			<p class="description"><?php esc_html_e( "Write 'em down y'all!" ); ?></p>
+		</div>
+		<script type="text/javascript">
+			jQuery( document ).ready( function( $ ) {
+				$( '.button-two-factor-backup-codes-generate' ).click( function() {
+					$.ajax( {
```
georgestephanis: Should this be firing a POST request?  I think it's currently a GET request.

georgestephanis:
Only utterly trivial stuff, hardly worth even thinking about.

I'm fine ignoring it and just merging.  :+1:

providers/class.two-factor-backup-codes.php
valendesigns: Ajax defaults to GET. Are you saying you want to change it to POST?

PR #15, opened 2015-08-04
Update plugin loading for phpunit

scotchfield:
Hi! I was trying to get the unit testing up and running, but hit a couple of snags. The main was was the use of is_plugin_active, which under my testing environment here wasn't registering two-factor/two-factor.php. I changed bootstrap to require the code and to set the active_plugins option. Does that seem fine? I wasn't able to find a spot where $GLOBALS['wp_tests_options']['active_plugins'] was actually used to load the plugins and update the option (maybe I missed it though).
Thanks! Looking forward to writing some tests. :)

### DIFF
--- a/tests/phpunit/includes/bootstrap.php
+++ b/tests/phpunit/includes/bootstrap.php
@@ -20,9 +20,20 @@ if ( ! file_exists( $_tests_dir . '/includes/' ) ) {
 
 require_once $_tests_dir . '/includes/functions.php';
 
-// Activate the plugins.
-if ( defined( 'WP_TEST_ACTIVATED_PLUGINS' ) ) {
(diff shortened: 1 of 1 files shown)


georgestephanis:
@valendesigns -- did you find anything when looking into attempting to duplicate this?

valendesigns:
Sorry, been busy. I'll look at it this evening.

scotchfield:
@georgestephanis @valendesigns Thanks for your time, you two. Let me know what you think.

valendesigns:
Unit test work for me with or without this PR merged. So if it fixes it for you, then I'm fine with that. I'll have to verify that the Travis still work after the merge and adjust if they don't.

valendesigns:
And everything is still functioning with Travis. Thanks @scotchfield

scotchfield:
❤️️
More unit tests on the way. Thanks!

PR #16, opened 2015-08-09
Introduce a 2FA column on the Users admin listing screen

johnbillion:
Introduces a column on the Users admin listing screen which displays each user's 2FA status.

It displays either `Disabled` for users with no 2FA enabled, or it displays their primary 2FA provider name.


georgestephanis:
My only real thought here is whether it would be better to just say yes/no, versus listing what their primary method is?  As Primary Method means very little, it's trivial to use any of their other enabled and configured methods.

valendesigns:
We could use Disabled/Enabled instead of yes/no.

PR #17, opened 2015-08-09
Don't escape the `<kbd>` tags used in the `Two_Factor_Email` provider

johnbillion:


aaroncampbell:
Sorry, I didn't see this when I sent in #18. I think removing the `<kbd>` tags is preferable (I didn't really see that much added benefit from them), but I'm fine either way.

georgestephanis:
Closing in favor of #18 -- there's not really a noticeable win imho for keeping them in.

PR #19, opened 2015-08-12
Add email provider tests, some core tests, and small bug fixes

scotchfield:
I started by adding test coverage for the email provider and a few basic tests for the core class. In addition, I added a few safety checks in email. Also using the mockmailer so email output can be tested (used in one test at the moment). Please let me know what you think.
Thanks!

### DIFF
--- a/providers/class.two-factor-email.php
+++ b/providers/class.two-factor-email.php
@@ -113,6 +113,10 @@ class Two_Factor_Email extends Two_Factor_Provider {
 	 * @param WP_User $user WP_User object of the logged-in user.
 	 */
 	public function authentication_page( $user ) {
+		if ( ! $user ) {
+			return;
+		}
+
 		$this->generate_and_email_token( $user );
 		require_once( ABSPATH .  '/wp-admin/includes/template.php' );
 		?>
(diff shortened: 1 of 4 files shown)


scotchfield:
I don't see these errors locally, is there something in the Travis script that's causing this?

valendesigns:
I'll test it out on my setup later tonight.

scotchfield:
It looks like an older version of phpunit that predates assertNotFalse. I can rewrite, or is it possible to move to a newer version on travis?

valendesigns:
It's what is on Travis. I would just rewrite the test. It's simple enough.

scotchfield:
Will do. Thanks!

valendesigns:
Looks like you need to define `SERVER_NAME`, as well.

scotchfield:
This is a strange one. In WP_UnitTestCase, there's a hook to wp_mail to set the server name in setUp (example.com) if it's not set already and to tear it down afterwards (set_wp_mail_globals). I can override, but it kind of stinks not knowing why that's not being called in the first place. Just going to test a few things locally. Sorry about the fails!

valendesigns:
I don't believe we need to `setUp` and `tearDown` before each test just to change server globals. Couldn't we use `setUpBeforeClass` & `tearDownAfterClass` instead?

scotchfield:
That's a really good observation, thanks! I'll push a change now.

georgestephanis:
@valendesigns Is this good to merge?

georgestephanis:
Reviewed, looks good to me and all checks pass.  Merging.

Issue #20, opened 2015-08-20
PHP Notice when enabling


mauteri:
Notice: Undefined index: _two_factor_enabled_providers in /vagrant/wordpress/wp-content/plugins/two-factor/class.two-factor-core.php on line 515

PR #21, opened 2015-08-20
PHP Notice when enabling #20

mauteri:


georgestephanis:
@mauteri Thanks for the suggestion -- rather than complicating the processing script with a ternary (I personally feel that complicated ternaries make code hard to read) -- how would you feel about just adding in a dummy input field so it is always passed, akin to #23 ?

mauteri:
Cool, that works too :-)

PR #22, opened 2015-08-20
Add FIDO U2F as a provider

shield-9:
Tested with:
* YubiKey NEO
* YubiKey Edge
* FIDO U2F Security Key (Yubico)
* FIDO U2F Security Key (Plug-up)


includes/Google/u2f-api.js
```diff
@@ -0,0 +1,361 @@
+// Copyright 2014 Google Inc. All rights reserved
+//
+// Use of this source code is governed by a BSD-style
```
georgestephanis: We'll need to read the source of the license and confirm that it's compatible to use in a GPL project such as WordPress (I think we can, just want to confirm)

includes/Google/u2f-api.js
```diff
@@ -0,0 +1,361 @@
...
+ * @type {Object}
+ */
+var u2f = u2f || {};
+
+/**
+ * The U2F extension id
+ * @type {string}
+ * @const
+ */
+u2f.EXTENSION_ID = 'kmendfapggjehodndflmmgagdbamhnfd';
```
georgestephanis: I believe this is referring to the extension in Chrome that enables U2F support, it's not some sort of api key that shouldn't be distributed.

includes/Yubico/U2F.php
```diff
@@ -0,0 +1,427 @@
...
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
```
georgestephanis: I believe that these licensing terms are fine to include in a GPL project, but would like someone to confirm.

includes/Yubico/U2F.php
```diff
@@ -0,0 +1,427 @@
...
+  private $attestDir;
+
+  /** @internal */
+  private static $FIXCERTS = array(
+    '349bca1031f8c82c4ceca38b9cebf1a69df9fb3b94eed99eb3fb9aa3822d26e8',
+    'dd574527df608e47ae45fbba75a2afdd5c20fd94a02419381813cd55a2a3398f',
+    '1d8764f0f7cd1352df6150045c8f638e517270e8b5dda1c63ade9c2280240cae',
+    'd0edc9a91a1677435a953390865d208c55b3183c6759c9b5a7ff494c322558eb',
+    '6073c436dcd064a48127ddbf6032ac1a66fd59a0c24434f070d4e564c124c897',
+    'ca993121846c464d666096d35f13bf44c1b05af205f9b4a1e00cf6cc10c5e511');
```
georgestephanis: Are these salts that would be better stored elsewhere, and not hardcoded in a library?

two-factor.php
```diff
@@ -24,6 +24,43 @@
...
+						'<a href="' . esc_url( 'http://php.net/supported-versions.php' ) . '">',
+						'</a>'
+					);
+				?>
+			</div>
+			<?php
+		}
+	}
+
+	add_action( 'admin_notices', 'upgrade_php_nag' );
```
georgestephanis: I'm not sure if we should add a nag, or just silently not include it.  I'm thinking perhaps not include the nag, and just silently drop the feature because Core won't keep it if it gets merged in.

two-factor.php
```diff
@@ -24,6 +24,43 @@
...
+	 *
+	 * @param array $providers Array of providers.
+	 * @return array Array of providers.
+	 */
+	function remove_fido_u2f_support( $providers ) {
+		unset( $providers['Two_Factor_FIDO_U2F'] );
+		return $providers;
+	}
+
+	add_filter( 'two_factor_providers', 'remove_fido_u2f_support' );
```
georgestephanis: Rather than doing this via a filter, I'd personally rather remove it right in the Core class here:

https://github.com/georgestephanis/two-factor/blob/d2bfdfc66e0d5f266e2d6e4d41974292278f7e8e/class.two-factor-core.php#L59

providers/class.two-factor-fido-u2f.php
```diff
@@ -55,7 +147,190 @@ public function validate_authentication( $user ) {}
...
+		}
+
+		$meta_ids = $wpdb->get_col( $query );
+		if ( ! count( $meta_ids ) ) {
+			return false;
+		}
+
+		$query = "DELETE FROM $table WHERE umeta_id IN( " . implode( ',', $meta_ids ) . ' )';
+
+		$count = $wpdb->query( $query );
```
georgestephanis: This could be done with the core function `delete_metadata_by_mid()` -- with the added benefit of handling the WordPress Object Cache.

https://github.com/WordPress/WordPress/blob/d65112d382c61aa46cc5ec785fd1fdcbad5690be/wp-includes/meta.php#L689-L774

providers/class.two-factor-fido-u2f.php
```diff
@@ -55,7 +147,190 @@ public function validate_authentication( $user ) {}
...
+		if ( ! is_numeric( $user_id ) ) {
+			return false;
+		}
+
+		$user_id = absint( $user_id );
+		if ( ! $user_id ) {
+			return false;
+		}
+
+		$table = $wpdb->usermeta;
```
georgestephanis: Why is this function doing direct database queries, instead of using `delete_user_meta()` ?

https://github.com/WordPress/WordPress/blob/ce557062f4123d8513378cf415b4e8b612c33ccc/wp-includes/user.php#L1304-L1321

`delete_user_meta()` can delete only where the meta value matches the third passed in argument, am I missing something here?

providers/class.two-factor-fido-u2f.php
```diff
@@ -55,7 +147,190 @@ public function validate_authentication( $user ) {}
...
+
+		return true;
+	}
+
+	/**
+	 * Detect browser support for FIDO U2F.
+	 *
+	 * @since 0.1-dev
+	 */
+	public static function is_browser_support() {
```
georgestephanis: I'm personally uncomfortable with checking this server side, is there any way this could be an in-browser javascript check instead?  If it's not feasible, so be it, we can always revisit later.

providers/class.two-factor-fido-u2f.php
```diff
@@ -55,7 +147,190 @@ public function validate_authentication( $user ) {}
...
+	 * @return array|bool Array of keys on success, false on failure.
+	 */
+	public static function get_security_keys( $user_id ) {
+		if ( ! is_numeric( $user_id ) ) {
+			return false;
+		}
+
+		$keys = get_user_meta( $user_id, self::REGISTERED_KEY_USER_META_KEY );
+		if ( $keys ) {
+			foreach ( $keys as $index => $key ) {
```
georgestephanis: This may be easier done as

```
foreach ( $keys as &$key ) {
    $key = (object) $key;
}
unset( $key );
```

But I'm fine either way.

providers/class.two-factor-fido-u2f.php
```diff
@@ -55,7 +147,190 @@ public function validate_authentication( $user ) {}
...
+		}
+
+		$register = array(
+			'keyHandle'   => $register->keyHandle,
+			'publicKey'   => $register->publicKey,
+			'certificate' => $register->certificate,
+			'counter'     => $register->counter,
+		);
+
+		$register['name']      = 'New Security Key';
```
georgestephanis: Should this name be made available for translation?  Is it user editable anywhere?

providers/class.two-factor-fido-u2f.php
```diff
@@ -23,12 +41,38 @@ static function get_instance() {
 	}
 
 	/**
+	 * Class constructor.
+	 *
+	 * @since 0.1-dev
+	 */
+	protected function __construct() {
```
georgestephanis: It may be worth adding in an extra `if ( version_compare( PHP_VERSION, '5.3.0', '<' ) ) return;` or something in here just as a backup to avoid fatal errors.

georgestephanis:
Failing unit tests seem to be because the PHP 5.3+ library is being included for some reason on PHP 5.2 instances.

includes/Yubico/U2F.php
valendesigns: Could any of the salts in `wp-config.php` be used instead, or are they not in a compatible format?

includes/Yubico/U2F.php
georgestephanis: k, looks fine.

includes/Google/u2f-api.js
georgestephanis: k, looks fine.

includes/Yubico/U2F.php
shield-9: This is not salt. According to https://github.com/Yubico/php-u2flib-server/commit/75a8730037950f05f1aa371db6ac9f464ba0a192 :
>some devices shipping with a certificate where unused bits was not set to 0

providers/class.two-factor-fido-u2f.php
shield-9: We need to do *partial match* for performance, but `delete_user_meta()` cannot do it.

providers/class.two-factor-fido-u2f.php
shield-9: Yes, this will be user-editable. But renaming feature is not implemented yet.

providers/class.two-factor-fido-u2f.php
shield-9: Updated in 51bab9c

providers/class.two-factor-fido-u2f.php
shield-9: Even if we add it, parsing of this file will be failed at `\` of `self::$u2f = new u2flib_server\U2F( set_url_scheme( '//' . $_SERVER['HTTP_HOST'] ) );`.

includes/Google/u2f-api.js
shield-9: `u2f.EXTENSION_ID` is pointing *CryptoToken Component* on Google Chome. (not API key)
Dropbox (https://goo.gl/VgCpCB) and Yubico's demo (https://goo.gl/2o7bqN) are using the same js.

providers/class.two-factor-fido-u2f.php
shield-9: This function is used in `is_available_for_user()`. I think it is not easy to replace with js and would like to revisit later.

shield-9:
All tests passed now. And ready to merge.

georgestephanis:
Merging!  Any future tweaks we can do via future pull requests.

PR #23, opened 2015-08-20
Add a dummy input.

### LINKED ISSUE #20: PHP Notice when enabling

georgestephanis:
Dummy input so $_POST value is passed when no providers are enabled.

Fixes #20

Alternate resolution to #21


valendesigns:
Are we going with #24 instead?

georgestephanis:
I'm fine with either.  I think I like #23 a touch better, as it's just using the existing html block as opposed to echoing, but either way works.

PR #24, opened 2015-08-20
Add a dummy input. #23

mauteri:


valendesigns:
You need a period at the end of the comment to pass the tests.

PR #28, opened 2015-08-24
Combine "out of codes" strings into one string for better translatability

aaroncampbell:
You can't break a single string or sentence into two parts, translate separately, and concatenate. This will make the string fully translatable.


aaroncampbell:
Okay, I added a comment to make it pass the WPCS tests, but there seems to still be an issue with PHP 5.2 not liking the indentation of the multi-line function call. I'm not sure what the deal is there

PR #29, opened 2015-08-29
First functional pass at totp

aaroncampbell:
This is still kind of a mess, but it's functional TOTP code that works with Google Authenticator and Authy. It's some older code that has been shoehorned in to this class. It needs some cleaning up as well as some serious JavaScript love.

Things still to do:
- Allow regeneration of the key so that a new QR code is loaded
- Test the verification authcode and save the key via JS so it doesn't require you to save your profile
- Notices of success and failure really suck the way this cobbled together, so we need to handle those better too.

Hopefully this is a decent jumping off point for someone though, or at least maybe some code that someone can rob to use in a new writing of this.

### DIFF
--- a/providers/class.two-factor-totp.php
+++ b/providers/class.two-factor-totp.php
@@ -2,16 +2,39 @@
 /**
  * Class for creating a Time Based One-Time Password provider.
  *
- * @since 0.1-dev
- *
  * @package Two_Factor
  */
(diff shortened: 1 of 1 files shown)


georgestephanis:
I'm doing some tidying and tweaking on this in the `aaroncampbell-totp` branch on the main repo here.

https://github.com/georgestephanis/two-factor/compare/aaroncampbell-totp

PR #25, opened 2015-08-20
move #2 to it's on branch so other people can work on it

voldemortensen:
Still a bit to do. i.e. get it compatible with parent class.

PR #31, opened 2015-08-30
TOTP: Doing some tinkering with and bringing #29 in to a branch for other folks to collab.

georgestephanis:
If merged, fixes #29 (which this is based off of) and would be a replacement for #2, #25.

Feels much more fleshed out than #25, may be worth focusing on this branch and pulling any wins from #25 over.


georgestephanis:
I'm merging this now, we can evaluate swapping the QR Code Generator and other stuff after it's in.

PR #35, opened 2015-08-31
Use same provider after failed login

shield-9:
This way is more convenient for users.


georgestephanis:
:+1: That's the thing we were discussing with @stevenkword the other week.  Totally slipped my mind, many thanks!

stevenkword:
Thanks, now I don't have to create the PR. :)

PR #37, opened 2015-08-31
Catch exceptions when instantiating FIDO U2F

scotchfield:
I hit a bunch of fatal errors due to the exceptions in the Yubico code--added a wrapper around get_instance and in the test cases.

### DIFF
--- a/class.two-factor-core.php
+++ b/class.two-factor-core.php
@@ -86,7 +86,11 @@ class Two_Factor_Core {
 			 * Confirm that it's been successfully included before instantiating.
 			 */
 			if ( class_exists( $class ) ) {
-				$providers[ $class ] = call_user_func( array( $class, 'get_instance' ) );
+				try {
+					$providers[ $class ] = call_user_func( array( $class, 'get_instance' ) );
+				} catch ( Exception $e ) {
+					// Could not get instance of $class
+				}
 			}
 		}
 
--- a/class.two-factor-core.php
+++ b/class.two-factor-core.php
@@ -89,7 +89,7 @@ class Two_Factor_Core {
 				try {
 					$providers[ $class ] = call_user_func( array( $class, 'get_instance' ) );
 				} catch ( Exception $e ) {
-					// Could not get instance of $class
+					echo esc_html( sprintf( __( 'Could not get provider instance: %s', 'two-factor' ), $class ) );
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


class.two-factor-core.php
```diff
@@ -86,7 +86,11 @@ public static function get_providers() {
 			 * Confirm that it's been successfully included before instantiating.
 			 */
 			if ( class_exists( $class ) ) {
-				$providers[ $class ] = call_user_func( array( $class, 'get_instance' ) );
+				try {
+					$providers[ $class ] = call_user_func( array( $class, 'get_instance' ) );
+				} catch ( Exception $e ) {
+					// Could not get instance of $class
```
shield-9: How about doing `unset( $providers[ $class ] );` here?

class.two-factor-core.php
```diff
@@ -86,7 +86,11 @@ public static function get_providers() {
 			 * Confirm that it's been successfully included before instantiating.
 			 */
 			if ( class_exists( $class ) ) {
-				$providers[ $class ] = call_user_func( array( $class, 'get_instance' ) );
+				try {
+					$providers[ $class ] = call_user_func( array( $class, 'get_instance' ) );
+				} catch ( Exception $e ) {
+					echo esc_html( sprintf( __( 'Could not get provider instance: %s', 'two-factor' ), $class ) );
```
shield-9: How about doing unset( $providers[ $class ] ); here?
I think it is not best to `echo` here, because `get_providers()` is used by some functions.

scotchfield: That makes sense, I'll update. Do you think there's any reason to output a problem here or can we just unset silently?

shield-9: This is called during `init` hook. So, I'm concerned that error messages might be displayed before <!DOCTYPE>

georgestephanis:
@scotchfield: While I agree that printing something out doesn't make sense on init, I do think it would be worth noting it somehow -- perhaps via [`trigger_error()`](http://php.net/manual/en/function.trigger-error.php) ?

scotchfield:
@georgestephanis: I could go either way on this, but I'm kind of leaning toward silence (two-factor silently discards Two_Factor_FIDO_U2F on older PHP versions, for example) or error_log (instead of writing an error handler for trigger_error). What do you think?

		error_log( sprintf( __( 'Could not get provider instance: %s', 'two-factor' ), $class ) );

shield-9:
+1 for outputting log.

georgestephanis:
Merging this.  If we want to add a log later we can.

PR #38, opened 2015-09-01
Travis CI: Switch from WP 4.0 to /trunk testing

ntwb:
No need to run Travis-CI tests with WordPress 4.0, but tests should be running against `/trunk`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,8 @@ php:
 env:
   - WP_VERSION=latest WP_MULTISITE=0
   - WP_VERSION=latest WP_MULTISITE=1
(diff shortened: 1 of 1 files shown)


georgestephanis:
latest = current stable release, trunk = trunk?

ntwb:
Yes, `latest` = `4.3` https://github.com/georgestephanis/two-factor/blob/master/bin/install-wp-tests.sh#L33-L35

And `trunk` = `/trunk` https://github.com/georgestephanis/two-factor/blob/master/bin/install-wp-tests.sh#L36-L37

PR #39, opened 2015-09-02
FIDO U2F: Add error message for PHP version

shield-9:
Related #37

This is kinder for admin and developers.


georgestephanis:
Should `trigger_error()` be html-escaped?  I wouldn't expect so, as it could often just be going directly into log files, where the escaping would be useless, and in either case, it's not output.

In practice, as there is no actual html in the strings, it should be completely fine, I'm just worrying about conceptually.

georgestephanis:
Whatever we do, we should be consistent between #37 and #39 -- for `trigger_error()` vs `error_log()`

shield-9:
Yeah, you're right. We don't need to escape here.

PR #41, opened 2015-09-02
FIDO U2F: Add rename feature

shield-9:
*Based on inline edit code of post.*


georgestephanis:
Once this gets a review, it's fine to merge conceptually.

PR #45, opened 2015-09-04
Silence CodeSniffer

shield-9:
I think escaping is not required here. But I'd like to follow the instruction from code sniffer to silence it.


shield-9:
It seems #46 is better.

PR #48, opened 2015-09-04
Basic test coverage for Backup Codes

stevenkword:


georgestephanis:
:+1:

tests/phpunit/tests/providers/class.two-factor-backup-codes.php
```diff
@@ -0,0 +1,117 @@
...
+		$user = new WP_User( $this->factory->user->create() );
+		$codes = $this->provider->generate_codes( $user, array( 'number' => 1 ) );
+		$validate = $this->provider->validate_code( $user, $codes[0] );
+
+		$this->assertFalse( $this->provider->validate_code( $user, $codes[0] ) );
+	}
+
+	/**
+	 * Show that validate_code fails for a different user's code.
+	 * @covers Two_Factor_Email::generate_code
```
shield-9: "Email" is typo of "Backup_Codes"?

tests/phpunit/tests/providers/class.two-factor-backup-codes.php
```diff
@@ -0,0 +1,117 @@
...
+	 * @covers Two_Factor_Email::validate_code
+	 */
+	function test_generate_code_and_validate_code_false_different_users() {
+		$user = new WP_User( $this->factory->user->create() );
+		$codes = $this->provider->generate_codes( $user, array( 'number' => 1 ) );
+
+		$codes2 = $this->provider->generate_codes( $user2, array( 'number' => 1 ) );
```
shield-9: L111 and L112 should be reversed.

tests/phpunit/tests/providers/class.two-factor-backup-codes.php
stevenkword: Fixed

georgestephanis:
Looks good, but a heads up to #49, as this will break the mergeability of that.

stevenkword:
How should we handle that then.  I was assuming we'd merge this first then update #49.

scotchfield:
Feel free to break, that's no problem!

providers/class.two-factor-backup-codes.php
```diff
@@ -273,10 +273,12 @@ public function validate_authentication( $user ) {
 	public function validate_code( $user, $code ) {
 		$backup_codes = get_user_meta( $user->ID, self::BACKUP_CODES_META_KEY, true );
 
-		foreach ( $backup_codes as $code_index => $code_hashed ) {
-			if ( wp_check_password( $code, $code_hashed, $user->ID ) ) {
-				$this->delete_code( $user, $code_hashed );
-				return true;
+		if ( is_array( $backup_codes ) && ! empty( $backup_codes ) ) {
```
shield-9: It seems `if ( $backup_codes ) {` is enough here.

georgestephanis: No harm in being explicit for the moment.

stevenkword: It is possible for the metakey value to come back as 'false' and not an empty array if a specified user has not created any codes in the past.

PR #49, opened 2015-09-06
Move version check into test and out of phpunit xml

scotchfield:
It's a bit troublesome to manually add each test file, can we move the PHP version check into the FIDO test setup and simplify the phpunit XML?


phpunit.xml
```diff
@@ -11,10 +11,9 @@
 	</php>
 	<testsuites>
 		<testsuite>
-			<file>tests/phpunit/tests/two-factor.php</file>
-			<file>tests/phpunit/tests/providers/class.two-factor-dummy.php</file>
-			<!-- This test fails in PHP 5.2. FIDO U2F support is PHP 5.3+ only. -->
-			<file phpVersion="5.3.0" phpVersionOperator=">=">tests/phpunit/tests/providers/class.two-factor-fido-u2f.php</file>
+			<directory suffix=".php">tests/phpunit/tests</directory>
+			<exclude>tests/phpunit/tests/providers/class.two-factor-fido-u2f.php</exclude>
+			<file phpVersion="5.3.0">tests/phpunit/tests/providers/class.two-factor-fido-u2f.php</file>
```
georgestephanis: Would this line need the `phpVersionOperator=">="` bit added to it, so it runs for 5.4 and higher as well?

scotchfield: Technically not, `phpVersionOperator` defaults to `>=`. You're right though, probably better to be explicit. I'll add it.

I'm having a hard time getting set with an older phpunit (<= 3.6) and PHP 5.2 to figure out why this file is still being included in the older builds. Unless I'm missing something, I can't see why the FIDO tests are still being picked up. (Of course, they clearly are, as the big red X is showing!) :(

scotchfield: (re: excluding phpVersionOperator, I was borrowing from WordPress's phpunit.xml and stole that syntax)

scotchfield:
Sorry, I think I misunderstood--are these Travis checks failing because the sniffer is catching the syntax error, not phpunit?

shield-9:
How about adding class.two-factor-fido-u2f.php to https://github.com/georgestephanis/two-factor/blob/master/bin/travis.before_script.sh#L14?

scotchfield:
Sigh. I think I mucked up this pull request. Sorry about that. :/

@georgestephanis I'll close and start with just the XML and travis.before_build.sh changes, if that's okay.

@shield-9 You're right, that line in the configuration actually looks like it might be malformed. Most configs are set to ignore the code sniffer on the tests directory. I tried to fix, but then merged all over it.

PR #59, opened 2015-09-10
Handle interim logins

### LINKED ISSUE #58: Interim Logins
We need to handle interim logins. When a login expires while a user is still on a wp-admin page, the heartbeat opens wp-login in an iframe. When it does, it adds an interim_login input that is passed along. We need to grab that, add it to our 2fa form to pass along, then handle it when we process that. Basically, it means that instead of redirecting at the end of a successful interim login, we need to load  a page with a "success" message and let login_header add the proper classes that will trigger the modal (iframe) to be closed.

I'm working on this now, but wanted to log the issue.

aaroncampbell:
Fixes #58

### DIFF
--- a/class.two-factor-core.php
+++ b/class.two-factor-core.php
@@ -303,6 +303,7 @@ class Two_Factor_Core {
 
 		$available_providers = self::get_available_providers_for_user( $user );
 		$backup_providers = array_diff_key( $available_providers, array( $provider_class => null ) );
+		$interim_login = isset( $_REQUEST['interim-login'] );
 
 		$rememberme = 0;
 		if ( isset( $_REQUEST['rememberme'] ) && $_REQUEST['rememberme'] ) {
@@ -320,7 +321,11 @@ class Two_Factor_Core {
 				<input type="hidden" name="provider"      id="provider"      value="<?php echo esc_attr( $provider_class ); ?>" />
 				<input type="hidden" name="wp-auth-id"    id="wp-auth-id"    value="<?php echo esc_attr( $user->ID ); ?>" />
 				<input type="hidden" name="wp-auth-nonce" id="wp-auth-nonce" value="<?php echo esc_attr( $login_nonce ); ?>" />
-				<input type="hidden" name="redirect_to"   id="redirect_to"   value="<?php echo esc_attr( $redirect_to ); ?>" />
+				<?php   if ( $interim_login ) { ?>
+					<input type="hidden" name="interim-login" value="1" />
+				<?php   } else { ?>
+					<input type="hidden" name="redirect_to" value="<?php echo esc_attr( $redirect_to ); ?>" />
+				<?php   } ?>
 				<input type="hidden" name="rememberme"    id="rememberme"    value="<?php echo esc_attr( $rememberme ); ?>" />
 
 				<?php $provider->authentication_page( $user ); ?>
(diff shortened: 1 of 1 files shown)


class.two-factor-core.php
```diff
@@ -320,7 +321,11 @@ public static function login_html( $user, $login_nonce, $redirect_to, $error_msg
 				<input type="hidden" name="provider"      id="provider"      value="<?php echo esc_attr( $provider_class ); ?>" />
 				<input type="hidden" name="wp-auth-id"    id="wp-auth-id"    value="<?php echo esc_attr( $user->ID ); ?>" />
 				<input type="hidden" name="wp-auth-nonce" id="wp-auth-nonce" value="<?php echo esc_attr( $login_nonce ); ?>" />
-				<input type="hidden" name="redirect_to"   id="redirect_to"   value="<?php echo esc_attr( $redirect_to ); ?>" />
+				<?php   if ( $interim_login ) { ?>
```
shield-9: http://php.net/control-structures.alternative-syntax would improve readability of code

class.two-factor-core.php
```diff
@@ -492,6 +497,29 @@ public static function login_form_validate_2fa() {
 
 		wp_set_auth_cookie( $user->ID, $rememberme );
 
+		// Must be global because that's how login_header() uses it.
+		global $interim_login;
+		$interim_login = isset($_REQUEST['interim-login']);
```
shield-9: > /tmp/wordpress/src/wp-content/plugins/two-factor/class.two-factor-core.php:502:26: error - Expected 1 spaces after opening bracket; 0 found (PEAR.Functions.FunctionCallSignature.SpaceAfterOpenBracket)
> /tmp/wordpress/src/wp-content/plugins/two-factor/class.two-factor-core.php:502:26: error - Expected 1 spaces before closing bracket; 0 found (PEAR.Functions.FunctionCallSignature.SpaceBeforeCloseBracket)

class.two-factor-core.php
```diff
@@ -492,6 +497,29 @@ public static function login_form_validate_2fa() {
...
+
+		if ( $interim_login ) {
+			$customize_login = isset( $_REQUEST['customize-login'] );
+			if ( $customize_login ) {
+				wp_enqueue_script( 'customize-base' );
+			}
+			$message = '<p class="message">' . __('You have logged in successfully.') . '</p>';
```
shield-9: > /tmp/wordpress/src/wp-content/plugins/two-factor/class.two-factor-core.php:509:48: error - Expected 1 spaces after opening bracket; 0 found (PEAR.Functions.FunctionCallSignature.SpaceAfterOpenBracket)
> /tmp/wordpress/src/wp-content/plugins/two-factor/class.two-factor-core.php:509:48: error - Expected 1 spaces before closing bracket; 0 found (PEAR.Functions.FunctionCallSignature.SpaceBeforeCloseBracket)

class.two-factor-core.php
aaroncampbell: On a personal level...I disagree. I hate the alternative control structures. <3 { and }

On a more realistic note, I don't object to changing it. I just pulled this directly from core: https://core.trac.wordpress.org/browser/trunk/src/wp-login.php#L931

class.two-factor-core.php
aaroncampbell: I'll fix this. Again, it was pulled straight from core (and apparently was missed when added because it obviously doesn't meet their coding standards).

class.two-factor-core.php
aaroncampbell: Again...bitten by copy/paste from core :( https://core.trac.wordpress.org/browser/trunk/src/wp-login.php#L833

class.two-factor-core.php
georgestephanis: My personal preference is -- when braces are used, to put the php opening and closing on sequential lines.  Thus,

```
<?php
if ( thing ) {
?>
```

and

```
<?php if ( $thing ) : ?>
```

But that's merely my personal preference, and would hate for us to get bogged down bikeshedding.

PR #62, opened 2015-09-11
Include a fix for alternate login forms.

### LINKED ISSUE #61: Add some sort of fallback for alternate login forms.
Some login forms -- [WooCommerce, for example](https://github.com/woothemes/woocommerce/blob/8b601adb3ca9d3e20da7b94c2e8d7dd93e4bf4fe/includes/class-wc-form-handler.php#L754-L823) -- use alternate login forms and don't submit to `wp-login.php` -- and so we don't have the normal wp-login.php functions like `login_header()` available.

I'd suggest we add in a check and if we're trying to authenticate without being on wp-login, we stash the redirect param and redirect to wp-login to handle the 2fa part, then redirect back to continue.

georgestephanis:
In situations where login_header() is inaccessible (because we’re on a
page other than wp-login.php because of a plugin or something) manually
redefine `login_header()` so we can call it there as well.

Also make sure we’re always submitting to the correct page and links go
to the correct page — `wp_login_url()` — if a plugin changes that, they
should also be reordering things to call the actions we use so that it
still works.

Resolves issues with WooCommerce logins from front-end pages. :+1:

Fixes #61

### DIFF
--- a/class.two-factor-core.php
+++ b/class.two-factor-core.php
@@ -230,10 +230,6 @@ class Two_Factor_Core {
 	 * @param WP_User $user WP_User object of the logged-in user.
 	 */
 	public static function show_two_factor_login( $user ) {
-		if ( ! function_exists( 'login_header' ) ) {
-			require_once( ABSPATH . WPINC . '/functions.wp-login.php' );
-		}
-
 		if ( ! $user ) {
 			$user = wp_get_current_user();
 		}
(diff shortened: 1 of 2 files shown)


class.two-factor-core.php
```diff
@@ -230,10 +230,6 @@ public static function wp_login( $user_login, $user ) {
 	 * @param WP_User $user WP_User object of the logged-in user.
 	 */
 	public static function show_two_factor_login( $user ) {
-		if ( ! function_exists( 'login_header' ) ) {
-			require_once( ABSPATH . WPINC . '/functions.wp-login.php' );
```
georgestephanis: Incidentally, when I first wrote this, I don't know why I thought `wp-includes/functions.wp-login.php` existed -- it never has best as I can tell.

valendesigns:
There is no other way than to duplicate the `login_header` function?

georgestephanis:
The core `login_header()` function is defined in `wp-login.php`, right after it re-requires `wp-load.php` and would attempt to redefine `ABSPATH` and such.  There is no other way to include the core version of the file. :(

georgestephanis:
Related: https://core.trac.wordpress.org/ticket/20279

valendesigns:
Just speaking out load here, but how would we overcome this issue during merge if we can't get the functions moved out of `wp-login.php`? We certainly can't duplicate functions in Core and if we can't get them moved we have a very real issue here. I think we need to start the lobbying now and present a unified front on this. We could try and discuss this in Slack and see if we can get buy in on the issue. We need to reopen 20279 with a compelling reason.

georgestephanis:
The other option is to just `wp_safe_redirect()` the user to wp-login.php -- the difficulty is that some plugins set up a redirect from wp-login.php to their front-end login page with some plugins, which would be an eternal redirect loop -- hence why I did it this way.

I agree that we may need to implement something else for mergeability, but for the moment, this suffices.

cc: @markjaquith for input on core architecture and such.

valendesigns:
Did some testing with WooCommerce and everything looks good.

PR #65, opened 2015-09-17
TOTP: Fix issue with pack on PHP 5.6.3+ in 32bit mode.

### LINKED ISSUE #64: TOTP pack function fails on PHP 5.6.3+ in 32bit mode
Yep, apparently that's actually a thing. I spent hours tracking this down for an iThemes customer (we use the same pack function) only to find that they're on PHP 5.6.6, on 64bit architecture, but running 32bit PHP. PHP doesn't have a built in method for checking this, but we can use PHP_INT_SIZE. If it's 4 we're on 32bit, if it's 8 we're on 64bit. I used `PHP_INT_SIZE >= 8` in the test, just because I dealing with 32 and 64 bit versions of PHP isn't enough fun and I hope they add a 128bit version some day.

aaroncampbell:
Fixes #64

### DIFF
--- a/providers/class.two-factor-totp.php
+++ b/providers/class.two-factor-totp.php
@@ -231,7 +231,8 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 	 * @return string Binary packed string.
 	 */
 	private static function pack64( $value ) {
-		if ( version_compare( PHP_VERSION, '5.6.3', '>=' ) ) {
+		// We can use the new 64bit pack functionality if we're on PHP 5.6.3+ in 64bit mode (PHP_INT_SIZE == 8)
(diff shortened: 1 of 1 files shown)


valendesigns:
Before we merge is there anyone with objections and/or could test it's effectiveness on a different system?

aaroncampbell:
This fixed this issue for 3 of our iThemes Security Pro clients that I'm aware of (on at least two different hosts, https://www.infomaniak.ch/en and https://www.ovh.co.uk/, I don't know what host the third was on). It BOGGLES MY MIND that ANY hosts are running 64bit servers, PHP 5.6.3+, and 32bit PHP. I'm totally open to more people testing it though.

valendesigns:
I'm happy to move forward with the merge, I just wanted to make sure it was tested in more than one system. It's an odd edge case indeed, very puzzling to see a host doing this.

PR #68, opened 2015-09-23
Pass WP_User object in 'show_user_security_settings'

shield-9:
This fixes:
* Trying to get property of non-object in
class.application-passwords.php on line 94
* Trying to get property of non-object in
class.application-passwords.php on line 98
* Trying to get property of non-object in
providers/class.two-factor-fido-u2f-admin.php on line 98
* Trying to get property of non-object in
providers/class.two-factor-fido-u2f-admin.php on line 101


shield-9:
Oops, duplicate of #66

PR #69, opened 2015-09-23
FIDO U2F: Fix invalid HTML

### LINKED ISSUE #67: FIDO: Unable to save user profile
The FIDO provider is adding something that is killing the ability to submit the user profile. Skimming Chrome's inspector, it looks like it is parsing the FIDO list table as ending the profile page's form element.

shield-9:
`<form>` in `<form>` is not valid!

Fixes #67


kraftbj:
Confirmed that this resolves 67.

PR #71, opened 2015-09-23
TOTP: Pre-encode spaces to fix iOS issue with URLs containing spaces

### LINKED ISSUE #70: iOS chokes on spaces in site name when scanning QR Code
Apparently it's a known issue that iOS doesn't like spaces in URLs. Since we put the site name in the URL, it needs to be encoded beforehand to handle any potential spaces

aaroncampbell:
This fixes #70

### DIFF
--- a/providers/class.two-factor-totp.php
+++ b/providers/class.two-factor-totp.php
@@ -285,9 +285,11 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 	 * @return string A URL to use as an img src to display the QR code
 	 */
 	public function get_google_qr_code( $name, $key, $title = null ) {
+		// Pre-encode spaces because iOS chokes on them
+		$name = str_replace( ' ', '%20', $name );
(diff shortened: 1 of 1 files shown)


valendesigns:
@aaroncampbell Which devices have you tested this change on so far?

aaroncampbell:
@valendesigns  I have no iDevices, but got people to test. I think the only things it's been tested with is the iPhone 5 and 6.

valendesigns:
Do you have an Android device to test with?

valendesigns:
Tests are now passing and this PR is ready for merge, that is once it's been verified that no regression exists with non iOS devices. I highly doubt there is a regression here, but it's better to check than to assume. Cheers!

aaroncampbell:
Oh yeah! I have several android devices. I tested it on my my phone (Galaxy S5), and three tablets (varying Galaxy tabs).

valendesigns:
Perfect! Thanks Aaron.

Issue #54, opened 2015-09-09
Backup Codes: Add print option


kraftbj:
Most backup code providers will offer a print link that opens up a generic window suitable for printing codes.

stevenkword:
Absolutely.  I think this will need to wait on some UX decisions from the core team, but we should definitely do this.

georgestephanis:
It should be trivial to create a popup window with javascript that just contains a 'print' button and the ten new codes -- perhaps with some additional line like `Two-Factor Backup Codes for %username% at %sitename% <%siteurl%>` or the like.

valendesigns:
When with a save button instead in #82.

PR #74, opened 2015-09-25
base fix for #54, add print button for backup codes

voldemortensen:


providers/class.two-factor-backup-codes.php
```diff
@@ -140,7 +141,7 @@ public function user_options( $user ) {
 						dataType: 'JSON',
 						success: function( response ) {
 							$( '.two-factor-backup-codes-wrapper' ).show();
-							$( '.two-factor-backup-codes-unused-codes' ).html( '' );
+							$( '.two-factor-backup-codes-unused-codes' ).html( '<p>Backup codes for ' + document.domain + '</p>' );
```
valendesigns: Text should be translatable.

user-edit.css
```diff
@@ -34,4 +34,24 @@
...
 .two-factor-methods-table tbody tr:nth-child(odd) {
 	background-color: #f9f9f9;
-}
\ No newline at end of file
+}
+
+@media print {
+	body * {
+		visibility: hidden;
+		max-height: 100px;
```
valendesigns: What is the `max-height` for?

voldemortensen: This was an attempt to make sure it didn't print 2-3 extra blank pages. It didn't work however. In my print preview it still shows 2 extra pages.

valendesigns: Try using display instead of visibility.

voldemortensen: display doesn't work because it hides all children, and you can't use display to get children back if the parent isn't shown.

valendesigns: Possibly a hidden overflow could work with max-height, or simply use an explicit height and absolute positioning on all elements?

valendesigns:
Since we're relying on JS to print the backup codes, I feel we should go all-in and not relying on CSS to hide and show everything else. I couldn't find a reliable way to display a clean looking UI without a second page being added to the print queue. Instead I opened a new window to use as our print document and wrote it out manually. Let me know if any of you have objections or concerns with this approach.

voldemortensen:
I think we should do that as a no js fallback. I could be wrong, but I think opening a new window goes against accessibility. I'd have to have someone versed in that to confirm though.

valendesigns:
There could be accessibility concerns we need to address, but I'm not aware of any personally. We'll need a third party to weigh in on the subject. However, my concerns with a CSS approach are simple. We can't assume that 2fa backup codes is the only reason someone would have to print the user profile and the method for hiding other elements is buggy at best. If we can address those concerns with a CSS only approach then I'm all for it. Otherwise, we need to do what works. I'm not sure there is an ideal approach to this issue, just different ones with trade offs.

stevenkword:
Can we elaborate on the new window / accessibility argument?  That's how Google is doing it.

https://accounts.google.com/b/0/SmsAuthSettings

stevenkword:
We should also consider the way GitHub handles this process. They use a button, but when you click it, there is no print / modal. It is a "download & print”, so it never leaves the page and prompts you to save the TXT file locally.

https://github.com/settings/security

PR #75, opened 2015-09-25
Add/app ajax

voldemortensen:
this is just what I had from months ago. Just pushing what I had. It probably in no way works anymore and I'll be working on updating it.

### DIFF
--- a/class.application-passwords.php
+++ b/class.application-passwords.php
@@ -5,13 +5,19 @@ class Application_Passwords {
 	const USERMETA_KEY_APPLICATION_PASSWORDS = '_application_passwords';
 
 	public static function add_hooks() {
-		add_filter( 'authenticate',             array( __CLASS__, 'authenticate' ), 10, 3 );
-		add_action( 'show_user_profile',        array( __CLASS__, 'show_user_profile' ) );
-		add_action( 'edit_user_profile',        array( __CLASS__, 'show_user_profile' ) );
-		add_action( 'personal_options_update',  array( __CLASS__, 'catch_submission' ), 0 );
-		add_action( 'edit_user_profile_update', array( __CLASS__, 'catch_submission' ), 0 );
-		add_action( 'load-profile.php',         array( __CLASS__, 'catch_delete_application_password' ) );
-		add_action( 'load-user-edit.php',       array( __CLASS__, 'catch_delete_application_password' ) );
+		add_filter( 'authenticate',                        array( __CLASS__, 'authenticate' ), 10, 3 );
+		add_action( 'show_user_profile',                   array( __CLASS__, 'show_user_profile' ) );
+		add_action( 'edit_user_profile',                   array( __CLASS__, 'show_user_profile' ) );
+		add_action( 'personal_options_update',             array( __CLASS__, 'catch_submission' ), 0 );
+		add_action( 'edit_user_profile_update',            array( __CLASS__, 'catch_submission' ), 0 );
+		add_action( 'load-profile.php',                    array( __CLASS__, 'catch_delete_application_password' ) );
+		add_action( 'load-user-edit.php',                  array( __CLASS__, 'catch_delete_application_password' ) );
+		add_action( 'admin_enqueue_scripts',               array( __CLASS__, 'enqueue_scripts' ) );
+		add_action( 'wp_ajax_delete_application_password', array( __CLASS__, 'ajax_delete_application_password' ) );
+	}
+
+	public static function enqueue_scripts() {
+		wp_enqueue_script( 'two-factor', TWO_FACTOR_URL . 'js/two-factor.js', array( 'jquery' ) );
 	}
 
 	public static function authenticate( $input_user, $username, $password ) {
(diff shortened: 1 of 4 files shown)


voldemortensen:
As of 60dff38 there are still a number of things to do. i.e. insert the new app info into the table and display the new password.

js/two-factor.js
```diff
@@ -0,0 +1,35 @@
+/* global ajaxurl */
+jQuery(document).ready(function ($) {
+
+	$('.delete').on('click', function( e ) {
```
valendesigns: This is to generic. We need to give additional specificity because the likelihood of another `.delete` existing is very high.

class.application-passwords.php
```diff
@@ -228,6 +244,56 @@ public static function create_new_application_password( $user_id, $name ) {
...
+	 *
+	 * @since 0.1-dev
+	 *
+	 * @access public
+	 * @static
+	 */
+	public static function ajax_create_application_password() {
+		$user_id = get_current_user_id();
+		$name = $_POST['app_name'];
+		if ( ! isset( $_POST['create_application_password'] ) || wp_verify_nonce( $_POST['_nonce_create_application_password'], 'create_application_password') ) {
```
valendesigns: Why would we allow this to proceed if `$_POST['create_application_password']` is not set?

class.application-passwords.php
```diff
@@ -228,6 +244,56 @@ public static function create_new_application_password( $user_id, $name ) {
...
 	/**
+	 * Create application password via ajax
+	 *
+	 * @since 0.1-dev
+	 *
+	 * @access public
+	 * @static
+	 */
+	public static function ajax_create_application_password() {
+		$user_id = get_current_user_id();
```
valendesigns: Can an admin user create/delete app passwords for other users? If so, then we can't reliably use `get_current_user_id()`.

js/two-factor.js
```diff
@@ -0,0 +1,35 @@
+/* global ajaxurl */
```
valendesigns: JSCS should be throwing errors in Travis for a lot of the spacing between parentheses in this file. Not sure why it didn't.

js/two-factor.js
```diff
@@ -0,0 +1,35 @@
...
+		$.post(ajaxurl, data, function(r) {
+			if( true === r.success ) {
+				$('tr[data-slug="' + $(link).data('deleteApplicationPassword') + '"]').remove();
+			}
+		});
+	});
+
+	$('#do_new_application_password').on('click', function( e ) {
+		e.preventDefault();
+		var app_name = $('input[name="new_application_password_name"]').val();
```
valendesigns: We should only have one `var` at the very beginning and before any other code. `e.preventDefault();` can be moved below. A separate, and personal preference is to use `event` and not `e` for clarity. Though I can live without it.

js/two-factor.js
```diff
@@ -0,0 +1,35 @@
...
+		var nonce    = $('input[name="create_application_password"]').val();
+		var data = {
+			'app_name'                          : app_name,
```
valendesigns: No spaces after the object keys before `:`.

js/two-factor.js
```diff
@@ -0,0 +1,35 @@
...
+			'action'                            : 'create_application_password',
+			'create_application_password': nonce,
+		};
+
+		$.post(ajaxurl, data, function(r) {
+			alert(r);
```
valendesigns: Can we do this without alert? If not, it needs to be added to the global above.

class.application-passwords.php
```diff
@@ -228,6 +244,56 @@ public static function create_new_application_password( $user_id, $name ) {
...
+
+			$res = self::delete_application_password( $user_id, $slug );
+		}
+
+		if ( true === $res ) {
+			wp_send_json_success();
+		}
+
+		wp_send_json_error();
+		wp_die();
```
valendesigns: `wp_die` is redundant because `wp_send_json` the function that `wp_send_json_success` & `wp_send_json_error` uses does this automatically when `DOING_AJAX`.

js/two-factor.js
voldemortensen: oops, that was debugging code. I hurried and committed what I had as I was leaving the office.

class.application-passwords.php
voldemortensen: I switched this from some other checking and missed the ! and should have switched || to && as well.

js/two-factor.js
voldemortensen: this is another of those trying to switch between two coding standards on the fly.

js/two-factor.js
voldemortensen: I don't mind switching to `event`, although I think pretty much any dev who has done anything with jQuery understands `e`.

valendesigns:
I've added a few `@todo` items in the source, there are I'm sure more we can add, though this list will facilitate getting the obvious out of the way and moving the PR further along.

##### `ajax_create_application_password()`
- [ ] Verify user permissions if creating another users app password.
- [ ] Implement `self::set_user_application_passwords()` so the next page load doesn't display password info.
- [ ] Implement list table HTML for a new item.

##### `ajax_delete_application_password()`
- [ ] Verify user permissions if deleting another users app password.
- [ ] Handle immediately deleting a password after creating it.

##### Create App Password click event
- [ ] Handle adding the item HTML to the list table.

__note__: I will add new items to this list as they're discovered or suggested.

valendesigns:
I'll get back to working on this PR soon. :wink:

shield-9:
Closing in favor of d8a85bb removing Application Passwords from this repository and running it in its own instead.

PR #76, opened 2015-09-26
Travis is not properly checking JSCS

valendesigns:
I'll fix the ones on the master branch real quick.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,0 +1,9 @@
+{
+	"preset": "wordpress",
+	"excludeFiles": [
+		"**/vendor/**",
+		"**/*.min.js",
+		"**/node_modules/**",
+		"**/u2f-api.js"
+	]
+}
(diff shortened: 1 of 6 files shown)


shield-9:
Test is failing on PHP 5.2, but it looks like another problem of PHPCS.
Looks good!

valendesigns:
The issue is that namespacing requires PHP 5.3+ and is why it's failing not PHPCS. We need to do a PHP check and return if it's not met.

valendesigns:
@shield-9 You were right. It seems the failure is indeed some PHPCS issue, though it did bring to light the fact namespacing is not supported in PHP versions below 5.3.0 so it's not a complete waste of time looking into it further.

Anyone against the PHP check staying in the PR and should we move 5.2 into `allow_failures` so it doesn't muck up the build moving forward?

I'll likely merge tomorrow and update the other PR's so they meet JSCS, as well.

PR #80, opened 2015-09-29
Don't submit TOTP input values without user action

### LINKED ISSUE #77: QR code notice displays when saving FIDO as the primary provider
> Two Factor Authentication not activated, you must specify authcode to ensure it is properly set up. Please re-scan the QR code and enter the code provided by your application.

shield-9:
This fixes #77

By disabling input element, its value won't be submitted to the server.

### DIFF
--- a/providers/class.two-factor-totp.php
+++ b/providers/class.two-factor-totp.php
@@ -102,32 +102,34 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 	 * @param integer $user_id The user ID whose options are being updated.
 	 */
 	public function user_two_factor_options_update( $user_id ) {
-		if ( ! empty( $_POST['do_totp_authcode'] ) && isset( $_POST['_nonce_user_two_factor_totp_options'] ) ) {
-			check_admin_referer( 'user_two_factor_totp_options', '_nonce_user_two_factor_totp_options' );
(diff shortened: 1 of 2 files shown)


aaroncampbell:
I haven't tested or anything, but it seems like this would break no-js, which feels like a regression.

valendesigns:
This PR brought to my attention the fact we have inline JS in the plugin, which needs to be written differently. Inline `onclick` events like this are not allowed in Core.

valendesigns:
This PR needs to be updated to not use a JS approach. It should probably attempt to filter out the input value when the form is submitted instead.

providers/class.two-factor-totp.php
```diff
@@ -99,7 +100,7 @@ public function user_two_factor_options( $user ) {
 	 * @param integer $user_id The user ID whose options are being updated.
 	 */
 	public function user_two_factor_options_update( $user_id ) {
-		if ( isset( $_POST['_nonce_user_two_factor_totp_options'] ) ) {
+		if ( ! empty( $_POST['do_totp_authcode'] ) ) {
```
valendesigns: This should probably still check for the nonces existence. For example:
`if ( ! empty( $_POST['do_totp_authcode'] ) && isset( $_POST['_nonce_user_two_factor_totp_options'] ) ) {`

georgestephanis:
I believe this was fixed by #145 so we don't need this more complex fix?  Happy to reopen if I'm mistaken.

PR #82, opened 2015-10-08
Download button for backup codes

stevenkword:
Adds a download button to the backup codes provider in favor of the print button solution proposed in #74.  The download button solution is better aligned with our accessibility best practices and is also useful to people on mobile devices or without immediate access to a printer.


dbrgn:
I would also vote for .txt download over print. I usually store my recovery codes on an encrypted memory stick, so print is not of much use to me.

PR #83, opened 2015-10-09
Use proper tests checkout for WP_VERSION=latest

danielbachhuber:


valendesigns:
@danielbachhuber Thanks for the PR and sorry for it sitting for so long. However, could you submit changes to this file upstream to https://github.com/xwp/wp-dev-lib as that is what this plugin is using now. Cheers!

PR #92, opened 2015-11-02
Use a 128 bit secret key by default

### LINKED ISSUE #90: Google Authenticator requires a 128 bit secret key
[RFC 4226 §4 requires a 128 bit secret key](https://en.wikipedia.org/wiki/Google_Authenticator#Technical_description).

Using the Google Authenticator on iPhone 6 returns an error after scanning the secret barcode or entering the secret manually. Bumping the `DEFAULT_KEY_BIT_SIZE` in `providers/class.two-factor-totp.php` to 128 fixes the issue.

kasparsd:
[RFC 4226 §4](https://tools.ietf.org/html/rfc4226#section-4) requires a 128 bit secret key.


kasparsd:
To be fixed in #93.

PR #93, opened 2015-11-09
Use proper base32 encoding for OTP secret keys

kasparsd:
- Implements base32 encoding 35268d1a03e3119804bf1f42e7bbbc506d7ebfb3 (ported from iThemes Security under GPL)
- Bumps the default OTP secret key length to 160 bits (suggested by [RFC4226](https://tools.ietf.org/html/rfc4226#section-4)).
- Includes tests for `base32_encode`, `base32_decode` and `is_valid_authcode`.
- Make base32 related methods public and static.


valendesigns:
@kasparsd This looks great! It works well for me, those of use who can test it should. Side note... how are you testing it with an actual device? It seem pretty labor intensive the way I'm doing it.

kasparsd:
@valendesigns Glad to hear that! 

I simply comment out [L71](https://github.com/kasparsd/two-factor/blob/fix/base32/providers/class.two-factor-totp.php#L71) and this forces to generate new QR codes on every refresh.

valendesigns:
Yeah, me too. However, I couldn't get it to fail with either branch.

kasparsd:
@valendesigns Even `master` wouldn't fail?

valendesigns:
Yeah, not sure how many times I need to keep destroying the key for it to fail.. maybe I gave up to early. :wink:

valendesigns:
Could you resolve the conflicts when you get a chance? I moved the tests around... Sorry!

kasparsd:
@valendesigns Done!

valendesigns:
@kasparsd Thanks, and this branch also fixed PHP 5.2!

PR #96, opened 2015-12-02
Simplify the admin UI

kasparsd:
Work in progress!

two-step-preview

Todo:
- [x] The "Primary second factor provider" should be selected automatically from the list of enabled provides, because users shouldn't know which second factor is more secure.
- [ ] Better U2F UI/UX.


georgestephanis:
Closing as this is way out of date, but we'll try to recreate the concepts behind it.

PR #98, opened 2015-12-03
remove use of HTTP_HOST to 1) avoid warnings and 2) be more secure

### LINKED ISSUE #97: avoid warnings
PHP Notice:  Undefined index: HTTP_HOST in /vagrant/www/wordpress-develop/src/wp-content/plugins/two-factor/providers/class.two-factor-fido-u2f.php on line 54
Notice: Undefined index: HTTP_HOST in /vagrant/www/wordpress-develop/src/wp-content/plugins/two-factor/providers/class.two-factor-fido-u2f.php on line 54

I get these every time I use wp-cli while two-factor is enabled.

voldemortensen:
fixes #97

### DIFF
--- a/providers/class.two-factor-fido-u2f.php
+++ b/providers/class.two-factor-fido-u2f.php
@@ -50,11 +50,11 @@ class Two_Factor_FIDO_U2F extends Two_Factor_Provider {
 			return;
 		}
 
-		$url = get_option( 'home' );
-		$correct_url = preg_replace( '/https?:/', '', $url );
+		$app_url_parts = parse_url( home_url() );
+		$app_url = sprintf( '%s://%s', $app_url_parts['scheme'], $app_url_parts['host'] );
 
 		require_once( TWO_FACTOR_DIR . 'includes/Yubico/U2F.php' );
-		self::$u2f = new u2flib_server\U2F( set_url_scheme( $correct_url ) );
+		self::$u2f = new u2flib_server\U2F( $app_url );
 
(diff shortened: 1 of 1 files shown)


shield-9:
The argument of `u2flib_server\U2F` should be *the web origin, written as a URI without a path*. (https://developers.yubico.com/U2F/App_ID.html)

This won't work if WordPress is installed in sub-directory.

valendesigns:
We should be checking for WP-CLI and bailing if in that context.

voldemortensen:
It would still be more secure to take the home option and strip the path off. HTTP_HOST is considered user input and can't be trusted because it comes from the Host: HTTP header (which can be modified).

kasparsd:
:+1: for getting rid of `HTTP_HOST`.

How about something like this:

```
$app_url_parts = parse_url( home_url() );
$app_url = sprintf( '%s://%s', $app_url_parts['scheme'], $app_url_parts['host'] );

require_once( TWO_FACTOR_DIR . 'includes/Yubico/U2F.php' );
self::$u2f = new u2flib_server\U2F( $app_url );
```

since `home_url()` returns the URL with the correct protocol and there is no need for `set_url_scheme()`.

I'm only worried about the fact that U2F AppIDs must be HTTPS, see https://developers.yubico.com/U2F/App_ID.html

voldemortensen:
Updated the pull request to use home_url and parse_url. That makes more sense.

shield-9:
LGTM!

PR #100, opened 2015-12-06
Force enabling Primary method

### LINKED ISSUE #52: When someone selects a method as `Primary` it should always be `Enabled` too.
When it is set as primary, toggle the linked Enabled checkbox to on.

shield-9:
This fixes #52


kasparsd:
Probably fixed in #167.

PR #111, opened 2016-01-16
U2F: Cannot add U2F key for another user

### LINKED ISSUE #88: Cannot add U2F key for another user
I logged in as the admin user and tried to add a U2F YubiKey to another user (through the user management). When touching the YubiKey button, the message "Profile updated" appeared but the key was not added. I tried this 2 or 3 times, never worked.

I then logged in as that user directly and adding the key worked.

Can someone reproduce this? :)

shield-9:
$user_id in user-edit.php should be the ID of **edited** user, but it is
done after `admin_enqueue_scripts` hook.

This fixes #88.

### DIFF
--- a/providers/class.two-factor-fido-u2f-admin.php
+++ b/providers/class.two-factor-fido-u2f-admin.php
@@ -56,6 +56,18 @@ class Two_Factor_FIDO_U2F_Admin {
 		wp_enqueue_script( 'fido-u2f-admin', plugins_url( 'js/fido-u2f-admin.js', __FILE__ ), array( 'jquery', 'u2f-api' ), null, true );
 
 		$user_id = get_current_user_id();
+		if ( 'user-edit.php' == $hook ) {
+			if ( empty( $_POST['user_id'] ) ) {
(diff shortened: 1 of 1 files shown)


providers/class.two-factor-fido-u2f-admin.php
```diff
@@ -56,6 +56,18 @@ public static function enqueue_assets( $hook ) {
...
 		$user_id = get_current_user_id();
+		if ( 'user-edit.php' == $hook ) {
+			if ( empty( $_POST['user_id'] ) ) {
+				if ( empty( $_GET['user_id'] ) ) {
+					$user_id = '';
+				} else {
+					$user_id = $_GET['user_id'];
+				}
+			} else {
+				$user_id = $_POST['user_id'];
```
kasparsd: @shield-9 We can't rely on the user_id passed in the request. This could allow `contributors` to add security keys to other users.

kasparsd:
I'll work on a fresh pull request that doesn't rely on request information for getting the profile user ID.

PR #128, opened 2016-10-21
Bump the Google and Yubico U2F libraries

kasparsd:


providers/js/fido-u2f-admin.js
```diff
@@ -3,35 +3,38 @@
...
 	$button.click( function() {
+		var registerRequest;
+
 		if ( $button.hasClass( 'clicked' ) ) {
 			return false;
 		} else {
 			$button.addClass( 'clicked' );
 		}
 
-		setTimeout( function() {
```
kasparsd: I'm not sure why this was using `setTimeout()` here.

providers/js/fido-u2f-login.js
```diff
@@ -1,13 +1,15 @@
...
-	setTimeout( function() {
-		window.console.log( 'sign', u2fL10n.request );
+	window.console.log( 'sign', u2fL10n.request );
 
-		u2f.sign( u2fL10n.request, function( data ) {
-			window.console.log( 'Authenticate callback', data );
+	u2f.sign( u2fL10n.request[0].appId, u2fL10n.request[0].challenge, u2fL10n.request, function( data ) {
+		window.console.log( 'Authenticate callback', data );
 
+		if ( data.errorCode ) {
```
kasparsd: This also ensures that we don't end up with an infinite reload when U2F sign fails.

shield-9:
Looks good! Merging

PR #130, opened 2016-10-21
Allow administrators to add/remove U2F security keys for other users

### LINKED ISSUE #88: Cannot add U2F key for another user

kasparsd:
Work in progress.

Fixes #88.


kasparsd:
Replaced by #349.

PR #135, opened 2016-11-22
Use random_bytes()

paragonie-scott:
random_compat has shipped with WordPress since 4.4.0

### DIFF
--- a/class.two-factor-core.php
+++ b/class.two-factor-core.php
@@ -406,7 +406,7 @@ class Two_Factor_Core {
 	 */
 	public static function create_login_nonce( $user_id ) {
 		$login_nonce               = array();
-		$login_nonce['key']        = wp_hash( $user_id . mt_rand() . microtime(), 'nonce' );
+		$login_nonce['key']        = bin2hex(random_bytes(16));
 		$login_nonce['expiration'] = time() + HOUR_IN_SECONDS;
 
 		if ( ! update_user_meta( $user_id, self::USER_META_NONCE_KEY, $login_nonce ) ) {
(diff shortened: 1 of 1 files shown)


georgestephanis:
The check is failing because of whitespace tidyness --

```
class.two-factor-core.php:409:38: error - Expected 1 spaces after opening bracket; 0 found (PEAR.Functions.FunctionCallSignature.SpaceAfterOpenBracket)
class.two-factor-core.php:409:38: error - Expected 1 spaces before closing bracket; 0 found (PEAR.Functions.FunctionCallSignature.SpaceBeforeCloseBracket)
class.two-factor-core.php:409:46: error - Expected 1 spaces after opening bracket; 0 found (PEAR.Functions.FunctionCallSignature.SpaceAfterOpenBracket)
class.two-factor-core.php:409:46: error - Expected 1 spaces before closing bracket; 0 found (PEAR.Functions.FunctionCallSignature.SpaceBeforeCloseBracket)
```

If you change it to `bin2hex( random_bytes( 16 ) )` -- it should be fine.

Also, is 16 enough?  It should be trivial to adjust, so I'm fine increasing if it's better for entropy.

paragonie-scott:
Sure. I just did 16 bytes -> bin2hex -> 32 hexits, which is the same size as what was previously generated.

As a rule of thumb, I tend to target the 256-bit level (50% chance of a single collision after 2^128), which would mean 32.

paragonie-scott:
It would, arguably, be better to fail closed when the CSPRNG is insecure, but I can add a try-catch if failing open is preferable to WordPress.

PR #144, opened 2017-02-22
Disable if in dev mode

Ipstenu:
Disable the plugin if we're in Jetpack dev mode.

### DIFF
--- a/two-factor.php
+++ b/two-factor.php
@@ -4,24 +4,29 @@
  * Plugin URI: http://github.com/georgestephanis/two-factor/
  * Description: A prototype extensible core to enable Two-Factor Authentication.
  * Author: George Stephanis
- * Version: 0.1-dev
+ * Version: 0.1-dev-20161216
  * Author URI: http://stephanis.info
  * Network: True
  * Text Domain: two-factor
  */
 
-/**
- * Shortcut constant to the path of this file.
- */
-define( 'TWO_FACTOR_DIR', plugin_dir_path( __FILE__ ) );
+// Disable if in dev mode.
+if ( JETPACK_DEV !== true ) {
 
-/**
- * Include the base class here, so that other plugins can also extend it.
- */
-require_once( TWO_FACTOR_DIR . 'providers/class.two-factor-provider.php' );
-
-/**
- * Include the core that handles the common bits.
- */
-require_once( TWO_FACTOR_DIR . 'class.two-factor-core.php' );
-Two_Factor_Core::add_hooks();
+	/**
+	 * Shortcut constant to the path of this file.
+	 */
+	define( 'TWO_FACTOR_DIR', plugin_dir_path( __FILE__ ) );
+	
+	/**
+	 * Include the base class here, so that other plugins can also extend it.
+	 */
+	require_once( TWO_FACTOR_DIR . 'providers/class.two-factor-provider.php' );
+	
+	/**
+	 * Include the core that handles the common bits.
+	 */
+	require_once( TWO_FACTOR_DIR . 'class.two-factor-core.php' );
+	Two_Factor_Core::add_hooks();
+	
+}


two-factor.php
```diff
@@ -4,24 +4,29 @@
...
  * Network: True
  * Text Domain: two-factor
  */
 
-/**
- * Shortcut constant to the path of this file.
- */
-define( 'TWO_FACTOR_DIR', plugin_dir_path( __FILE__ ) );
+// Disable if in dev mode.
+if ( JETPACK_DEV !== true ) {
```
swissspidy: This would need a `defined()` check first. Also, I think the constant is called `JETPACK_DEV_DEBUG`.

swissspidy:
This would mean I couldn't use 2FA just because I'm running Jetpack in dev mode? Why?

kasparsd:
@Ipstenu Why is this necessary? I couldn't find a reference to an existing issue this solves.

Secondly, some sites might have `JETPACK_DEV_DEBUG` enabled on production and this change will disable their U2F implementation.

Ipstenu:
It was from a conversation a waaaay long time ago, and I thought I'd made a ticket but ...

Anywho. The point is that when developing locally, there isn't the same need for 2FA. Sucking a live site down for localhost improvements or creating a non-public staging environment is a case where the addition of 2FA is often (though I will agree not always) unnecessary.

I grabbed Jetpack's dev mode in lieu of WP having a 'No, not production' mode (which IMO it should but that's another bear to cross). In the majority of cases, one has Jetpack in dev mode when one is developing and doesn't need to send out data to A8C.

> Jetpack offers a Development Mode that is enabled when using Jetpack on a localhost.

That is never intended to be used in production. Can't stop people, I know but the point of it was meant for times when you DON'T need to run everything.

georgestephanis:
This isn't a good idea. If the user doesn't want it in local mode, they can just disable the plugin there. There's a lot of sites that use the JP constant on public sites in lieu of slimpack or the like.

Basically, if the site owner activates two factor, we shouldn't override that because we assume that we know better

If we were to do something like this, we would do one of a few things: disable if domain has no dots in it (like localhost) or if DNS for the current domain fails or maps to a private range.

As the concern is more for if the site is accessible via the web than if a constant may be accurately or inaccurately defined.

two-factor.php
georgestephanis: Also, rather than wrap the entire contents of the file in a conditional, it'd probably be easier to just test for the opposite and then `return;`

voldemortensen:
FWIW, I see _almost_ no problems with doing this. Just because people do use something that was clearly never meant to be used in production doesn't mean we should leave out smart functionality. If someone sets the DISABLE_FILE_MODS constant, they expect files not to be modified. They should similarly expect that JETPACK_DEV puts it into development mode. 2FA in _a lot_ of dev environments is overkill.

Personally, I think its a worse idea to disable if DNS is the private range. That would disable it for major intranet sites running on WordPress.

The only thing that has really mostly convinced me that this shouldn't go in is
>If the user doesn't want it in local mode, they can just disable the plugin there. 

The problem with that is the user experience. I foresee people being confused as their TOTP app has the domain for their live site, not a staging site. How do they log in? The secret key is the same, so the same tokens can be used, but that's not immediately apparent to someone who doesn't know how that works. Also, that's just one more thing to manage. If you accidentally move the entire database back to production (instead of just posts or whatever), you've just disabled 2FA on production. I'm sure there are a few more things I thought about but my brain is blanking on me.

By the time I got done writing this, I'm now on the fence about it. There are some great reasons to do it, and some great reasons not to.

kasparsd:
We shouldn't assume things on behalf of users. Secondly, `JETPACK_DEV_DEBUG` is a Jetpack thing and we never know how that value will change. I see this as equivalent of adding a check for `WP_DEBUG` and assuming that users don't want 2FA when `WP_DEBUG` is `true`.

Ipstenu:
Jetpack Dev mode is specially indented for use in local host environments. That's it's job. It's documented.

Debug mode is not the same.

I should have (and thought I had, so my bad here) made an issue first and then this, but the crux is that 2fa is overkill in devmode AND the only way to disable it is to log in. Which while the live code works on a different domain (and by the way, is that weird to anyone else?) it's not a logical jump that it should. Thus leading to people being confused and having to delete a plugin, causing Dev and Live to no longer match.

That said. This is clearly something that can go both ways.

I strongly feel we SHOULD make decisions for users. That's our job, to make the decisions after weighing convenience, security, and real world uses. Decisions. Not options :) and in the case of 2fa, where a lot of education is **required** for effective use, we have to consider the case of local hosts.

But. I did this wrong. Shoulda been an issue. I'll close this.

PR #147, opened 2017-02-28
Only show fully configured providers as enabled

### LINKED ISSUE #146: TOTP reports as enabled even if it hasn't been fully configured
Per #78 TOTP can be bypassed if it is not fully configured. Let's check that it is fully configured before reporting that it is enabled.

joshbetz:
Fixes #146


georgestephanis:
Looks good.

Issue #148, opened 2017-03-16
https not on standard port --> Registration Failed 2 at fido-u2f-admin.js


pwallner:
I have my https port not on 443 (I use 8443 instead), therefore I got a Registration Failed 2’ at fido-u2f-admin.js

I solved the issue to hack the code in class.two-factor-fido-u2f.php
```$app_id = sprintf( 'https://%s:8443', $home_url_parts['host'] );```

The 8443 is also part of the settings -> general -> wordpress url and website url

So maybe you can check the port and include it at $app_id?

Regards
Patrick

**Comment by kasparsd** [comment]
@pwallner Thanks for reporting this. FIDO U2F does allow [any port as long as it's over HTTPS](https://fidoalliance.org/specs/fido-u2f-v1.0-ps-20141009/fido-appid-and-facets-ps-20141009.html#processing-rules-for-appid-and-facetid-assertions).

**Comment by kasparsd** [comment]
@pwallner Can you please check if changing `$app_id` to the following works in your case:

```
$app_id = set_url_scheme( home_url(), 'https' );
```

**Comment by pwallner** [comment]
No it's not working because wordpress is not on root, means I have:

`https://www.xxx.com:8443/wordpress`

with your mod I get as appid:
```appId: "https://www.xx.com:8443/wordpress"```

with my mod:
```appId: "https://www.xx.com:8443"```
which is working

**Comment by kasparsd** [comment]
Thank you for testing @pwallner. You're correct -- we should be removing any [path from the AppID](https://developers.yubico.com/U2F/App_ID.html):

> For the Web, the Facet ID is the web origin, written as a URI without a path (and therefore without a trailing slash). For example https://login.paypal.com.

**Comment by pwallner** [comment]
Good way could be:
```
public function getDomain() {
    		$sURL    = site_url(); // WordPress function
    		$asParts = parse_url( $sURL ); // PHP function

    		if ( ! $asParts )
      			return $sURL;

    		$sScheme = $asParts['scheme'];
		if ( $sScheme != "https")
			echo "<script>console.log( 'For U2F \'https\' required!' );</script>";
    		$nPort   = $asParts['port'];		
    		$sHost   = $asParts['host'];
    		$nPort   = ('https' == $sScheme AND 443 == $nPort) ? '' : $nPort;
    		$nPort   = ! empty( $nPort ) ? ":$nPort" : '';
    		$sReturn = $sScheme . '://' . $sHost . $nPort;

    		return $sReturn;
	}
[...]

PR #149, opened 2017-03-16
Include URL protocol in the U2F AppID

kasparsd:
A possible fix for #148.

I'm not sure about the following:

- Should we be using `home_url()` or `site_url()` for the AppID? Keeping the `home_url()` makes sense because it has worked well so far.

- Some sites might include extra path name in `home_url()` which wasn't there before this change. This will break their U2F login.

- How do we write unit tests for this? I would like to check that the output of `get_u2f_app_id()` always includes `https://`, for example.

### DIFF
--- a/providers/class.two-factor-fido-u2f.php
+++ b/providers/class.two-factor-fido-u2f.php
@@ -53,12 +53,8 @@ class Two_Factor_FIDO_U2F extends Two_Factor_Provider {
 			return;
 		}
 
-		// U2F requires the AppID to use HTTPS and a top-level domain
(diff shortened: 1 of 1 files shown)


georgestephanis:
I don't know enough about this to review it adequately.  @shield-9 -- any thoughts?

kasparsd:
@georgestephanis This only introduces support for hostnames with a custom port in the site URL. The rest of functionality is exactly the same.

georgestephanis:
Okay, looks reasonable.

Issue #155, opened 2017-03-20
Invalidate email codes only when entered correctly


joshbetz:
Email code are currently only invalidated if you type it in wrong. Instead, they should probably be invalidated after some time limit (maybe 60 seconds).

kasparsd:
Indeed, it should also delete the token [if it was valid](https://github.com/georgestephanis/two-factor/blob/0c672510a9bdf608be93e7908779c3dc9ea59261/providers/class.two-factor-email.php#L77) to ensure that it can't be re-used.

kasparsd:
@joshbetz The tokens should be invalidated the first time they're entered correctly to avoid re-using the same token. Most tokens are [one-time use to prevent replay attacks](https://en.wikipedia.org/wiki/One-time_password):

> This means that a potential intruder who manages to record an OTP that was already used to log in to a service or to conduct a transaction will not be able to abuse it, since it will no longer be valid

joshbetz:
> The tokens should be invalidated the first time they're entered correctly to avoid re-using the same token.

Yep, #156 does that. This issue is about invalidating codes after some time limit though.

kasparsd:
Thanks for the confirmation @joshbetz. What is the purpose of having the timeout now that the tokens are invalidated only after entered correctly (both #156 and #160)?

How about we add the "Email new access code" link instead?

joshbetz:
#156 doesn't invalidate tokens after a timeout, only after they're entered correctly -- the same thing as #160. #155 is about the timeout.

kasparsd:
@joshbetz I'm sorry for the confusion. I'm just trying to understand the use-case for having the timeout?

PR #156, opened 2017-03-20
Don't resend email codes on unsuccessful logins

joshbetz:
If you type the code wrong, we currently resend a new token. This seems confusing and unnecessary.

### DIFF
--- a/providers/class.two-factor-email.php
+++ b/providers/class.two-factor-email.php
@@ -74,9 +74,9 @@ class Two_Factor_Email extends Two_Factor_Provider {
 	public function validate_token( $user_id, $token ) {
 		$hashed_token = get_user_meta( $user_id, self::TOKEN_META_KEY, true );
 		if ( wp_hash( $token ) !== $hashed_token ) {
-			$this->delete_token( $user_id );
 			return false;
 		}
+		$this->delete_token( $user_id );
 		return true;
 	}
 
@@ -120,7 +120,10 @@ class Two_Factor_Email extends Two_Factor_Provider {
 			return;
 		}
 
-		$this->generate_and_email_token( $user );
+		if ( ! isset( $_GET['action'] ) || 'validate_2fa' !== $_GET['action'] ) {
+			$this->generate_and_email_token( $user );
+		}
+		
 		require_once( ABSPATH .  '/wp-admin/includes/template.php' );
 		?>
 		<p><?php esc_html_e( 'A verification code has been sent to the email address associated with your account.', 'two-factor' ); ?></p>


kasparsd:
@joshbetz This also fixes #155, correct?

providers/class.two-factor-email.php
```diff
@@ -120,7 +120,10 @@ public function authentication_page( $user ) {
 			return;
 		}
 
-		$this->generate_and_email_token( $user );
+		if ( ! isset( $_GET['action'] ) || 'validate_2fa' !== $_GET['action'] ) {
```
kasparsd: @joshbetz We should probably ensure that a valid token has been generated and is available for `validate_token()` before bailing here.

joshbetz:
> This also fixes #155, correct?

No, codes are still valid forever until they're used.

providers/class.two-factor-email.php
```diff
@@ -74,9 +74,9 @@ public function generate_token( $user_id ) {
 	public function validate_token( $user_id, $token ) {
 		$hashed_token = get_user_meta( $user_id, self::TOKEN_META_KEY, true );
 		if ( wp_hash( $token ) !== $hashed_token ) {
-			$this->delete_token( $user_id );
 			return false;
 		}
+		$this->delete_token( $user_id );
```
kasparsd: @joshbetz This will also ensure that a token can't be re-used once entered correctly. Generating fresh tokens on every 2FA login page took care of this before your other changes.

kasparsd:
@joshbetz What if a user deletes the email or it doesn't get pass the spam filter? How would the request a new token?

providers/class.two-factor-email.php
joshbetz: Good point

joshbetz:
@kasparsd I guess we should add a link to generate and send a new code.

georgestephanis:
I'm not sure if we should regenerate and send a new code -- I'd think personally we should let the tokens live in the db until a successful login, which then wipes all unused tokens.

I've been in situations where email from a server is delayed by 10 minutes, and I've requested repeated emails -- if I finally got the first one, only to discover it had been voided by a subsequent prior request, it'd be frustrating especially as there's nothing wrong with the existing one.

joshbetz:
> I'm not sure if we should regenerate and send a new code -- I'd think personally we should let the tokens live in the db until a successful login, which then wipes all unused tokens.

Agree, but I don't think that can be solved independently from this issue?

voldemortensen:
> I'm not sure if we should regenerate and send a new code -- I'd think personally we should let the tokens live in the db until a successful login, which then wipes all unused tokens.

I think they should be stored in the db until a successful login and its within the validity period of the TOTP token, otherwise it defeats the purpose of TOTP (at least every TOTP based 2fa implementation I've seen or worked on). The tokens are only supposed to be valid for a set amount of time. Alternatively, to avoid the db altogether, whenever a token is submitted for validation, we could just generate the last 5 or 10 minutes (filterable to fit others needs) worth of tokens and check the submitted token against that list.

kasparsd:
This has been fixed in #160.

PR #160, opened 2017-03-28
Invalidate email tokens only after entered correctly

### LINKED ISSUE #155: Invalidate email codes only when entered correctly

kasparsd:
Fixes #155.

Previously it was generating new tokens every time a user entered incorrect token value. This sent out a new email for every login attempt.


providers/class.two-factor-email.php
```diff
@@ -73,10 +73,15 @@ public function generate_token( $user_id ) {
 	 */
 	public function validate_token( $user_id, $token ) {
 		$hashed_token = get_user_meta( $user_id, self::TOKEN_META_KEY, true );
-		if ( wp_hash( $token ) !== $hashed_token ) {
-			$this->delete_token( $user_id );
+
+		// Bail if token is empty or it doesn't match.
+		if ( empty( $hashed_token ) || ( wp_hash( $token ) !== $hashed_token ) ) {
```
georgestephanis: Shouldn't this return false if `$hashed_token` is empty?  If the meta somehow gets otherwise deleted, we probably shouldn't treat it as a successful match?

kasparsd: @georgestephanis I think it does return `false` if the `$hashed_token` is empty from what I understand. Did I miss something?

kasparsd:
@georgestephanis Can we merge this?

providers/class.two-factor-email.php
kasparsd: @georgestephanis Did you manage to check my comment. It already returns `false` if `$hashed_token` is empty.

providers/class.two-factor-email.php
```diff
@@ -120,7 +141,10 @@ public function authentication_page( $user ) {
 			return;
 		}
 
-		$this->generate_and_email_token( $user );
+		if ( ! $this->user_has_token( $user->ID ) ) {
+			$this->generate_and_email_token( $user );
```
kasparsd: Email a new code only if none found for the user. Previously we were emailing a new code every time user enters an incorrect code.

Issue #117, opened 2016-03-20
Use JS to detect Fido U2F support


kasparsd:
Quoting #116:

> After changing the http to https, and unconditionally returning true in providers/class.two-factor-fido-u2f.php, function is_browser_support, I can use the yubikey also with firefox and the u2f addon: https://addons.mozilla.org/en-US/firefox/addon/u2f-support-add-on/
> 
> So it would be nice if this would be supported from the plugin out of the box.

Tea23:
It looks like this is happening because instead of checking for U2F support, the plugin just checks your user agent to see if it's Chrome 41+. I set a user agent override to `Mozilla/5.0 (Windows NT 6.1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2228.0 Safari/537.36` and it's working fine.

Tea23:
Oh boy - really important caveat found... If I disable the override, the thing just *skips* U2F. This makes the U2F option in the plugin 100% redundant because an attacker can just set their user agent to Firefox, or actually use Firefox.

kraftbj:
We'd probably need to figure out the best way to test for support. Is there a common standard we can check against instead of UA?

norbusan:
Isn't there a way to check whether the actually used U2F function is available? Looking into the `Google/u2f-api.js` I see `u2f.getMessagePort` which tries to communicate. Isn't that the better test then checking for browser name?

shield-9:
With current structure of this plugin, checking U2F availability using JavaScript isn't easy. (And AFAIK, there is no way to check server-side without using browser name and version.)

norbusan:
Hmm, ok. Pity. What about simply doing *no* checks and metion in the docs that one needs to have either Chrome or Firefox with some plugin. Failures would be detected and indicated as such.

Tea23:
I can use Github's U2F in Firefox with the plugin without changing my user agent. It could be worth investigating how they do the check.

JacobHenner:
Any progress on this? Being able to use U2F with FF and WP would be great.

shield-9:
Your pull request is welcome! (Including Firefox support) At the moment, I have no plan to work on this issue, because we need better way to detect browser support.

kraftbj:
Thinking about how to do this, what do you think of:
1. Drop PHP browser check
2. Move check for `window.u2f` or something similar to detect for FIDO support via JS
[...]

PR #161, opened 2017-03-28
Load FIDO U2F JS only when needed

### LINKED ISSUE #126: Don't add Fido U2F JS to the login page
It should be added only to the U2F token login page and [not the username/password page](https://github.com/georgestephanis/two-factor/blob/b3e2eec7f4804ecc673e7ce9c644b9a6795c7158/providers/class.two-factor-fido-u2f.php#L84).

kasparsd:
Fixes #126 and part of #117.

### DIFF
--- a/providers/class.two-factor-fido-u2f.php
+++ b/providers/class.two-factor-fido-u2f.php
@@ -79,7 +79,6 @@ class Two_Factor_FIDO_U2F extends Two_Factor_Provider {
 			true
 		);
 
-		add_action( 'login_enqueue_scripts',                array( $this, 'login_enqueue_assets' ) );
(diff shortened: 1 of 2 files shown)


providers/class.two-factor-fido-u2f.php
```diff
@@ -95,19 +94,6 @@ public function get_label() {
 	}
 
 	/**
-	 * Enqueue assets for login form.
-	 *
-	 * @since 0.1-dev
-	 */
-	public function login_enqueue_assets() {
-		if ( ! self::is_browser_support() ) {
```
georgestephanis: Complement to #162 which handles the rest of the removal of browser support.

kasparsd:
@georgestephanis Can we merge this too?

georgestephanis:
yarp

PR #166, opened 2017-03-30
Better U2F Support Detection

### LINKED ISSUE #165: Report Firefox as a supported browser if the extension is installed
Currently the Google U2F JS library fails the U2F API version detection on Firefox with the [U2F Support extension](https://addons.mozilla.org/en-US/firefox/addon/u2f-support-add-on/) enabled.

See #117 and #163.

kasparsd:
Fixes #165.

Unfortunately there is no official way to detect U2F API support in web browsers. However, in reality only Chrome and Firefox can support U2F.

Chrome supports U2F natively via [`chrome.runtime.sendMessage()`](https://github.com/georgestephanis/two-factor/blob/8b03e926aeee703d8368b108040342f48538592b/includes/Google/u2f-api.js#L196-L206) so we can simply check for `chrome.runtime` and assume that it does support U2F.

The [Firefox U2F extension](https://addons.mozilla.org/en-US/firefox/addon/u2f-support-add-on/) works by [creating a global `window.u2f` object](https://github.com/prefiks/u2f4moz/blob/ab97706a178a3299a6568bab7dcdd320108c870f/ext/data/content-script.js#L120) that maps to its internal functions. So we check if `window.u2f` exists before adding the wrapper in `u2f-api.js`.

## Chrome

chrome-ok

## Firefox with Extension

firefox-ok

## Safari

safari-fail

### DIFF
--- a/includes/Google/u2f-api.js
+++ b/includes/Google/u2f-api.js
@@ -17,6 +17,12 @@
 var u2f = u2f || {};
 
 /**
+ * Check if browser supports U2F API before this wrapper was added.
+ * @type {int}
+ */
+var u2f.HasNativeApiSupport = ( u2f && u2f.register );
+
+/**
  * FIDO U2F Javascript API Version
  * @number
  */
--- a/includes/Google/u2f-api.js
+++ b/includes/Google/u2f-api.js
@@ -20,7 +20,7 @@ var u2f = u2f || {};
  * Check if browser supports U2F API before this wrapper was added.
  * @type {int}
  */
-var u2f.HasNativeApiSupport = ( u2f && u2f.register );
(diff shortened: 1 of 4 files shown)


includes/Google/u2f-api.js
```diff
@@ -17,6 +17,12 @@
 var u2f = u2f || {};
 
 /**
+ * Check if browser supports U2F API before this wrapper was added.
+ * @type {int}
+ */
+u2f.HasNativeApiSupport = ( ( u2f && u2f.register ) || ( chrome && chrome.runtime ) );
```
kasparsd: `u2f.register` should exist in Firefox with the extension enabled before we add our own `u2f.register` wrapper further down this file.

kasparsd:
@georgestephanis This can also be merged.

PR #167, opened 2017-03-30
Fix the primary provider UI

joshbetz:
In the 2fa setup screen, we show a radio button for the primary
provider, but we just use whatever is stored in the database. In other
places, we fallback to the first enabled provider if the one in the
database isn't active. This prevents inactive providers from being shown
as the primary provider.


joshbetz:
Maybe related to #52?

class.two-factor-core.php
```diff
@@ -590,9 +590,18 @@ public static function manage_users_custom_column( $output, $column_name, $user_
 	 */
 	public static function user_two_factor_options( $user ) {
 		wp_enqueue_style( 'user-edit-2fa', plugins_url( 'user-edit.css', __FILE__ ) );
+
 		$enabled_providers = array_keys( self::get_available_providers_for_user( $user->ID ) );
-		$primary_provider = get_user_meta( $user->ID, self::PROVIDER_USER_META_KEY, true );
+		$primary_provider = self::get_primary_provider_for_user( $user->ID );
+
+		if ( ! empty( $primary_provider ) && is_object( $primary_provider ) ) {
+			$primary_provider_key = get_class( $primary_provider );
```
kasparsd: @joshbetz `get_class()` will create a warning if a non-object is supplied so I added these additional checks.

class.two-factor-core.php
```diff
@@ -648,20 +657,23 @@ public static function user_two_factor_options( $user ) {
...
+
+			// Enable only the available providers.
 			$enabled_providers = array_intersect( $enabled_providers, array_keys( $providers ) );
 			update_user_meta( $user_id, self::ENABLED_PROVIDERS_USER_META_KEY, $enabled_providers );
 
-			// Whitelist the new values to only the available classes and empty.
+			// Primary provider must be enabled.
 			$new_provider = isset( $_POST[ self::PROVIDER_USER_META_KEY ] ) ? $_POST[ self::PROVIDER_USER_META_KEY ] : '';
-			if ( empty( $new_provider ) || array_key_exists( $new_provider, $providers ) ) {
+			if ( ! empty( $new_provider ) && in_array( $new_provider, $enabled_providers, true ) ) {
```
kasparsd: @joshbetz Here we ensure that the primary method is also enabled.

kasparsd:
@georgestephanis Could we merge this?

class.two-factor-core.php
joshbetz: 👍👍👍

PR #172, opened 2017-03-31
Add WP.org Assets

### LINKED ISSUE #47: We Need A Logo.
We need a logo and assets for the plugins banner.

Break out your Papyrus and Comic Sans, warm up the Rainbow Gradients and Drop Shadows, it's time.

kasparsd:
Fixes #47.

Keys, gates, locks and safes are cliché.

Most two factor implementations are based on:

- something you know
- something you have

So we use the trick of negative space to create a pair of elements that "fit".

## Cover Image

two-factor

--- 

## Plugin Icon

two-factor-search


georgestephanis:
Yessssss thankyou

PR #179, opened 2017-05-23
Let's use a slightly better source of randomness than array_rand()

georgestephanis:


providers/class.two-factor-provider.php
```diff
@@ -71,13 +71,21 @@ public function print_label() {
...
+		}
 		$code = '';
 		if ( ! is_array( $chars ) ) {
 			$chars = str_split( $chars );
 		}
+		// Make sure it isn't an associative array which could mess up the key lookup.
+		$chars = array_values( $chars );
 		for ( $i = 0; $i < $length; $i++ ) {
-			$key = array_rand( $chars );
+			$key = random_int( 0, sizeof( $chars ) - 1 );
```
ericmann: I like the idea of better randomness, but `random_int()` isn't available [in PHP < 7.0](http://php.net/manual/en/function.random-int.php).

We could bundle [random_compat](https://github.com/paragonie/random_compat) with the project to maintain support for older versions of PHP. Otherwise, this change will bump the minimum version requirement.

georgestephanis: I mean, Core does bundle a polyfill for `random_int()` -- https://github.com/WordPress/WordPress/blob/0a81bcf4fffef460f22fda3fe5ccdf0376d1e3d9/wp-includes/random_compat/random_int.php

ericmann: As, same lib already there :-) Carry on.

kasparsd:
Fixed in #218.

PR #181, opened 2017-06-03
Email: Resend Token

### LINKED ISSUE #177: Add a link to resend email tokens
Most users that had enabled email tokens had a token already generated and stored which is why it now fails to send it after #156 was merged in.

The temporary solution is to add a "Re-send token" link.

kasparsd:
Fixes #177 which was introduced in #160, #155.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ language:
     - node_js
 
 php:
-    - 5.2
-    - 5.3
     - 5.4
     - 5.5
     - 5.6
@@ -15,10 +13,6 @@ php:
 env:
     - WP_VERSION=trunk WP_MULTISITE=0
     - WP_VERSION=trunk WP_MULTISITE=1
-
-matrix:
-  allow_failures:
-    - php: 5.2
 
 before_script:
     - export DEV_LIB_PATH=dev-lib
--- a/class.two-factor-core.php
+++ b/class.two-factor-core.php
@@ -500,7 +500,7 @@ class Two_Factor_Core {
 			$provider = self::get_primary_provider_for_user( $user->ID );
 		}
 
-		// Allow providers to re-send codes, etc.
+		// Allow the provider to re-send codes, etc.
 		if ( true === $provider->pre_process_authentication( $user ) ) {
 			$login_nonce = self::create_login_nonce( $user->ID );
 			if ( ! $login_nonce ) {
--- a/class.two-factor-core.php
+++ b/class.two-factor-core.php
@@ -503,7 +503,7 @@ class Two_Factor_Core {
 				wp_die( esc_html__( 'Failed to create a login nonce.', 'two-factor' ) );
 			}
 
-			self::login_html( $user, $login_nonce['key'], $_REQUEST['redirect_to'], esc_html__( 'ERROR: Invalid verification code.' ), $provider );
+			self::login_html( $user, $login_nonce['key'], $_POST['redirect_to'], esc_html__( 'ERROR: Invalid verification code.', 'two-factor' ), $provider );
 			exit;
 		}
 
--- a/class.two-factor-core.php
+++ b/class.two-factor-core.php
@@ -501,7 +501,7 @@ class Two_Factor_Core {
 		}
 
 		// Allow providers to re-send codes, etc.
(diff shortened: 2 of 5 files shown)


class.two-factor-core.php
```diff
@@ -495,15 +495,27 @@ public static function login_form_validate_2fa() {
 			$provider = self::get_primary_provider_for_user( $user->ID );
 		}
 
+		// Allow providers to re-send codes, etc.
+		if ( $provider->process_request( $user ) ) {
```
kasparsd: We introduce `process_request()` for when we don't want to process the actual 2FA. However, we do this only after verifying the 2FA nonce.

providers/class.two-factor-email.php
```diff
@@ -152,6 +170,9 @@ public function authentication_page( $user ) {
 			<label for="authcode"><?php esc_html_e( 'Verification Code:', 'two-factor' ); ?></label>
 			<input type="tel" name="two-factor-email-code" id="authcode" class="input" value="" size="20" pattern="[0-9]*" />
 		</p>
+		<p class="two-factor-email-resend">
+			<input type="submit" class="button" name="two-factor-email-code-resend" value="<?php esc_attr_e( 'Resend Code', 'two-factor' ); ?>" />
```
kasparsd: I would prefer to have this as a regular link and a GET request instead but we don't have access to the current 2FA request token.

providers/class.two-factor-email.php
```diff
@@ -167,6 +188,22 @@ public function authentication_page( $user ) {
...
 
 	/**
+	 * Send the email code, if missing or requested.
+	 *
+	 * @param  WP_USer $user WP_User object of the logged-in user.
+	 *
+	 * @return boolean       If the two factor should be authenticated.
+	 */
+	public function process_request( $user ) {
+		if ( isset( $user->ID ) && isset( $_POST['two-factor-email-code-resend'] ) ) {
```
kasparsd: Note that the 2FA request nonce has been already verified before this is called.

providers/class.two-factor-email.php
```diff
@@ -79,6 +79,23 @@ public function user_has_token( $user_id ) {
...
+	 * Get the authentication token for the user.
+	 *
+	 * @param  int $user_id    User ID.
+	 *
+	 * @return string|boolean  User token or `false` if no token found.
+	 */
+	public function get_user_token( $user_id ) {
+		$hashed_token = get_user_meta( $user_id, self::TOKEN_META_KEY, true );
+
+		if ( ! empty( $hashed_token ) && is_string( $hashed_token ) ) {
```
georgestephanis: With `$single` being specified as `true` is there any way the returned data won't be a string?  Even if it's a number, wouldn't it be cast as a string?

providers/class.two-factor-provider.php
```diff
@@ -45,6 +45,18 @@ public function print_label() {
...
 
 	/**
+	 * Allow providers to process the two factor authentication request for
+	 * additional tasks such as re-sending tokens, etc.
+	 *
+	 * @param WP_User $user WP_User object of the logged-in user.
+	 *
+	 * @return boolean Return `true` to prevent further processing.
+	 */
+	public function process_request( $user ) {
```
georgestephanis: Should the function name here be `pre_process_request` or the like to acknowledge it runs first and can short out the other processing?

.travis.yml
```diff
@@ -5,8 +5,6 @@ language:
     - node_js
 
 php:
-    - 5.2
```
kasparsd: These are no longer available on the Trusty boxes.

providers/class.two-factor-email.php
kasparsd: @georgestephanis This is mainly to protect from returning anything but a string or `false`.

tests/providers/class.two-factor-email.php
```diff
@@ -169,4 +169,36 @@ function test_is_available_for_user() {
 		$this->assertTrue( $this->provider->is_available_for_user( false ) );
 	}
 
-}
\ No newline at end of file
+	/**
```
kasparsd: And here some tests for the new methods.

PR #186, opened 2017-07-20
Fix U2F browser support check

francislavoie:
On certain browsers that don't support U2F, it'll throw an error when trying to load the U2F js lib if `chrome` is not strictly undefined. I tried this on Epiphany, an open source browser, for example. I implemented a browser support check in my application based on this.

Additionally, but not covered by this PR, there is a Safari browser extension and helper app to implement U2F, but it won't work with this Wordpress plugin because it replaces the `u2f` object with its own. I suggest either moving the browser support check out of this file and into a different namespace, or waiting for a change to be made on the Safari extension to not replace the `u2f` object. I mentioned that here: https://github.com/Safari-FIDO-U2F/Safari-FIDO-U2F/issues/17

I don't actually use this plugin or Wordpress, so do with this what you will. 😄

### DIFF
--- a/includes/Google/u2f-api.js
+++ b/includes/Google/u2f-api.js
@@ -20,7 +20,7 @@ var u2f = u2f || {};
  * Check if browser supports U2F API before this wrapper was added.
  * @type {int}
  */
-u2f.HasNativeApiSupport = ( ( u2f && u2f.register ) || ( chrome && chrome.runtime ) );
+u2f.HasNativeApiSupport = ( ( u2f && u2f.register ) || ( typeof chrome !== 'undefined' && chrome.runtime ) );
 
 /**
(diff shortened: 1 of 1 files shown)


francislavoie:
Actually, I realized something - this check won't work with any browser's implementation where they load the u2f support after the `u2f-api.js` file is loaded. This is the case for the Safari implementation because it loads their extension's JS after the page loads its own. The browser support check happens at load time of the JS file, so `u2f.register` won't be set at that point, but does get set later by the extension's JS.

This PR still provides a valid fix for a specific use-case, but it won't work for the Safari implementation.

kasparsd:
@francislavoie Do you know if Safari loads the extension JS before or after `DOMContentLoaded`? Would love to make this work with every browser.

PR #207, opened 2017-12-22
Full RFC6238 Compatibility

ericmann:
The full TOTP specification supports not only SHA1, but also SHA256 and SHA512. Up til now, the TOTP provider in this plugin (and most other PHP implementations) claims to support a specified hash type, but doesn't actually work with anything other than SHA1.

This is due to key lengths and hash lengths being _different_ for the three hash variants.

This change introcudes support for both SHA256 and SHA512, porting the implementation directly from https://github.com/ericmann/totp.

See https://tools.ietf.org/html/rfc6238#section-1.2 for more information on the `MAY USE` notation for SHA256 and SHA512.

See https://tools.ietf.org/html/rfc6238#appendix-A for a fully compliant reference implementation in Java.

See https://tools.ietf.org/html/rfc6238#appendix-B for test vectors showing TOTPs generated for specific time values and the three hash variants.

See https://github.com/ericmann/totp/blob/master/test/phpunit/ReferenceTest.php for example unit tests verifying this particular implementation before it was ported to the plugin.


kasparsd:
This looks good to me @ericmann! Could we include some tests for this since you already have them in the upstream repo.

ericmann:
Good catch on the `try`/`catch`. I'll get that updated and pull in some tests soon, then we can move forward.

ericmann:
@kasparsd Feedback addressed, and I pulled in the reference tests from the RFC spec.

Kind of frustrating to try to work around `time()` without namespaces, but I think this'll do it :-)

karimqub:
Access

providers/class-two-factor-totp.php
```diff
@@ -303,10 +332,12 @@ public function validate_authentication( $user ) {
 	 *
 	 * @param string $key      The share secret key to use.
 	 * @param string $authcode The code to test.
+	 * @param string $hash      The hash used to calculate the code.
+	 * @param int    $time_step The size of the time step.
 	 *
 	 * @return bool Whether the code is valid within the time frame
 	 */
-	public static function is_valid_authcode( $key, $authcode ) {
+	public static function is_valid_authcode( $key, $authcode, $hash = 'sha1', $time_step = 30 ) {
```
kasparsd, line 340: Should `$hash` default to `self::DEFAULT_CRYPTO` to match what `calc_totp` does? Or maybe it could be a class property with a dedicated getter that passes the value through a filter to enable overrides?

Same with the `$time_step` -- should that default to `self::DEFAULT_TIME_STEP_SEC`?

providers/class-two-factor-totp.php
```diff
@@ -56,6 +56,35 @@ protected function __construct() {
...
+
+		return self::$now;
+	}
+
+	/**
+	 * Set up the internal state of time() invocations for deterministic generation.
+	 *
+	 * @param int $now Timestamp to use when overriding time().
+	 */
+	public static function __set_time( $now ) {
```
kasparsd, line 84: It this necessary strictly for unit tests?

ericmann: Correct. The tests need to explicitly pass in the time in order to deterministically generate a TOTP.

kasparsd (approved):
This looks great! Thanks so much @ericmann!

kasparsd:
Looks like Travis is no longer running the unit tests for this repo so I've created #405 to address it. Solving that first will allow us to ensure this change hasn't introduced a regression.

jeffpaul:
Per [yesterday's bug scrub](https://wordpress.slack.com/archives/C04L4KCUB/p1648040434425739), we're punting this to a future release to allow for narrow focus on the U2F deprecation work in 0.8.0.

jeffpaul:
@kasparsd seems like resolving merge conflicts here, checking that tests pass, could get this into the 0.11.0 release perhaps?

ericmann:
I have no idea why GH thinks I'm asking to merge so much - it's literally just the last 3 commits and they're rebased atop `master` already. A direct comparison of the same contribution shows the changes: https://github.com/WordPress/two-factor/compare/master...ericmann:two-factor:totp-parity?expand=1

jeffpaul:
@ericmann maybe an oddity with you originally pushing from a fork?  perhaps a new PR that cherrypicks those last 3 commits will show "better" on GitHub?

ericmann:
Closing this to fix the GH display oddity.

PR #209, opened 2018-01-22
Make 'Log In' button default on verification form

wturrell:
On the Verification Code screen, currently if you
press return after typing/pasting the code,
'Resend Code' is the submit button that is triggered,
rather than 'Log In', which is confusing for users
as nothing appears to happen when they've provided
the correct code, yet they see ever growing incoming
emails (in the case of email verification).  For
most users Log In would be the expected behaviour.

Similarly, when you tab, Resend Code is the first
button rather than login.

Fixed by changing the order of the buttons in the
HTML (this also makes buttons horizontally aligned).

### DIFF
--- a/providers/class.two-factor-email.php
+++ b/providers/class.two-factor-email.php
@@ -176,6 +176,7 @@ class Two_Factor_Email extends Two_Factor_Provider {
 		<p>
 			<label for="authcode"><?php esc_html_e( 'Verification Code:', 'two-factor' ); ?></label>
 			<input type="tel" name="two-factor-email-code" id="authcode" class="input" value="" size="20" pattern="[0-9]*" />
+			<?php submit_button( __( 'Log In', 'two-factor' ) ); ?>
 		</p>
 		<p class="two-factor-email-resend">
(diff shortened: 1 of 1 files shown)


kasparsd:
Thanks for reporting the issue and creating the pull request @wturrell!

"Resend Code" should probably be a regular link for a GET request but there was some issues with nonces, if I'm not mistaken.

providers/class.two-factor-email.php
```diff
@@ -176,6 +176,7 @@ public function authentication_page( $user ) {
 		<p>
 			<label for="authcode"><?php esc_html_e( 'Verification Code:', 'two-factor' ); ?></label>
 			<input type="tel" name="two-factor-email-code" id="authcode" class="input" value="" size="20" pattern="[0-9]*" />
+			<?php submit_button( __( 'Log In', 'two-factor' ) ); ?>
```
kasparsd: @wturrell I just noticed that the output of `submit_button()` includes a paragraph tag so this should be outside the wrapping paragraph.

PR #218, opened 2018-02-25
Remove the usage of array_rand() in Two_Factor_Provider::get_code()

dd32:
Remove the usage of array_rand() and use wp_rand() as per wp_generate_password().

This avoids the issues outlined in #179 by simply using the WordPress provided functions.

### DIFF
--- a/providers/class.two-factor-provider.php
+++ b/providers/class.two-factor-provider.php
@@ -85,12 +85,11 @@ abstract class Two_Factor_Provider {
 	 */
 	public function get_code( $length = 8, $chars = '1234567890' ) {
 		$code = '';
-		if ( ! is_array( $chars ) ) {
-			$chars = str_split( $chars );
+		if ( is_array( $chars ) ) {
+			$chars = implode( '', $chars );
 		}
 		for ( $i = 0; $i < $length; $i++ ) {
-			$key = array_rand( $chars );
-			$code .= $chars[ $key ];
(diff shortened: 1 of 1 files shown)


providers/class.two-factor-provider.php
```diff
@@ -85,12 +85,11 @@ public function pre_process_authentication( $user ) {
...
 		$code = '';
-		if ( ! is_array( $chars ) ) {
-			$chars = str_split( $chars );
+		if ( is_array( $chars ) ) {
+			$chars = implode( '', $chars );
 		}
 		for ( $i = 0; $i < $length; $i++ ) {
-			$key = array_rand( $chars );
-			$code .= $chars[ $key ];
+			$code .= substr( $chars, wp_rand( 0, strlen( $chars ) - 1 ), 1 );
```
kasparsd: @dd32 This runs `strlen( $chars )` on every iteration while `$chars` never changes -- maybe bring it outside the `for` loop?

```php
$seed_length = strlen( $chars );

for ( $i = 0; $i < $length; $i++ ) {
    $code .= substr( $chars, wp_rand( 0, $seed_length - 1 ), 1 );
}
```

dd32: A micro-optimization such as that didn't seem worth spending any time on. A further micro-optimization is to move the `- 1` outside of the loop too.

The code was copied directly from `wp_generate_password()`, but feel free to do that instead.

PR #219, opened 2018-02-25
U2F: Don't trigger a PHP warning about U2F if it's not in use.

dd32:
The warning about U2F can probably be removed entirely, but this at least won't think about triggering it if it's not enabled.


kasparsd:
Thanks @dd32!

PR #221, opened 2018-02-27
Add 2fa info column to network admins user list table

lakrisgubben:
PR adds a filter on `wpmu_users_columns` so the 2fa info shows when looking at the users in the network admin on a multisite install. I think it's even more important to get an easy overview there compared to on a single site so one can see if all super admins have 2fa activated. :)


kasparsd:
This is great, thank you @lakrisgubben!

PR #224, opened 2018-04-04
Allow plugin to be installed in wordpress through composer

chadwick37:
By adding the `"type": "wordpress-plugin"` we can install the plugin to the plugins folder with composer. https://getcomposer.org/doc/04-schema.md#type


kasparsd:
This is great, thanks @chadwick37!

PR #228, opened 2018-04-27
Added translation file and language directory.

hisanuco:


kasparsd:
Thanks for the pull request @hisanuco!

WP.org now supports community translations through https://translate.wordpress.org/projects/wp-plugins/two-factor so it would be better to use that service instead of bundled translations.

Support for that was added here:

https://github.com/georgestephanis/two-factor/commit/210f64a3d5fedd987b63131d914e20c4e619df18

PR #239, opened 2018-09-03
Optionally Force 2fa

### LINKED ISSUE #255: Enforce 2FA
Fixed in #239.

- https://wordpress.org/support/topic/feature-request-enforce-two-factor-login-for-all-users/

mikeselander:
Fixes #255.

Adds a network-level and site level (if non-multisite) setting to require two-factor authentication on a per-role or global level. If 2fa forcing is on and a user matches the criteria, we force the user into a 2fa settings screen and they cannot use the site until they add valid 2fa to their account.

**Super administrators can edit the settings with this interface via network (or site if single site) settings:**
screen shot 2018-09-04 at 1 40 41 pm

**Users who fall within the required buckets will see an interface like this:**
login-wout-2fa


mikeselander:
Hey folks, apologies - this is nowhere near review and I hit the wrong button 🤦‍♂️

kasparsd:
@mikeselander This looks great! Will you open a new pull request later or should I re-open this one?

mikeselander:
@kasparsd it still needs quite a bit of smoothing out which I'll be working on today. I'll re-open this once it's in a better state :)

assets/js/force-2fa.js
```diff
@@ -0,0 +1,88 @@
+/* global ajaxurl, jQuery */
+
+/**
+ * Checks that an element has a non-empty `name` and `value` property.
+ *
+ * @param  {Element} element The element to check
+ * @return {Boolean} true if the element is an input, false if not
+ */
+var isValidElement = function( element ) {
```
kasparsd: Should we wrap these methods in an anonymous function to isolate the namespace? Some of these function names are very generic and could cause conflicts.

assets/js/force-2fa.js
```diff
@@ -0,0 +1,88 @@
...
+};
+
+/**
+ * Retrieves input data from a form and returns it as a JSON object.
+ *
+ * @param  {HTMLFormControlsCollection} elements the form elements
+ * @return {Object} form data as an object literal
+ */
+var formToJSON = function( elements ) {
+	return [].reduce.call( elements, function( data, element ) {
```
kasparsd: jQuery has [`serializeArray()`](https://api.jquery.com/serializeArray/) -- could we re-use that instead?

assets/js/force-2fa.js
```diff
@@ -0,0 +1,88 @@
...
+	jQuery.post(
+		ajaxurl,
+		formData,
+		function () {
+			window.location.reload();
+		}
+	);
+};
+
+window.addEventListener( 'load', function() {
```
kasparsd: Maybe we could stick to jQuery helpers to support more browsers? `window.addEventListener( 'load', function() {} )` could be replaced with `$( document ).ready(function() {})` since we actually need to wait for `DOMContentLoaded`.

assets/js/force-2fa.js
```diff
@@ -0,0 +1,88 @@
...
+	var form = document.querySelector( '#force_2fa_form' );
+	form.addEventListener( 'submit', handleFormSubmit );
```
kasparsd: This could be replaced with `$( '#force_2fa_form' ).on( 'submit', handleFormSubmit );` for better browser compatibility.

class.two-factor-force.php
```diff
@@ -0,0 +1,488 @@
...
+
+	/**
+	 * Register scripts.
+	 */
+	public static function register_scripts() {
+		// Script for handling AJAX submission in force 2fa takeover screen.
+		wp_register_script(
+			'two-factor-form-script',
+			plugins_url( 'assets/js/force-2fa.js', __FILE__ ),
+			[],
```
kasparsd: Looks like this actually depends on jQuery so we should add it here.

class.two-factor-force.php
```diff
@@ -0,0 +1,488 @@
...
+			'0.1',
+			false
```
kasparsd: Could we add this to the footer instead?

class.two-factor-force.php
```diff
@@ -0,0 +1,488 @@
...
+	/**
+	 * Maybe force the 2fa login page on a user.
+	 *
+	 * If 2fa is required for a user (based on universal or role settings),
+	 * we display the 2-factor options page so that a user must validly enable
+	 * a 2-factor authentication of some kind to perform any action on their site.
+	 * This occurs both on the front and backend.
+	 */
+	public static function maybe_redirect_to_2fa_settings() {
+		if ( ! self::should_user_redirect( get_current_user_id() ) || isset( $_GET['force_2fa_screen'] ) ) {
```
kasparsd: Looks like `force_2fa_screen` is re-used in a few places so we could introduce a constant for the key.

class.two-factor-force.php
```diff
@@ -0,0 +1,488 @@
...
+	 *
+	 * @since 0.1-dev
+	 */
+	public static function force_2fa_login_html() {
+		wp_enqueue_script( 'jquery' );
+		wp_enqueue_script( 'two-factor-form-script' );
+
+		// If Fido is a valid 2fa Provider, enqueue its assets.
+		$providers = Two_Factor_Core::get_providers();
+		if ( in_array( 'Two_Factor_FIDO_U2F', array_keys( $providers ), true ) ) {
```
kasparsd: Looks like this is now coupled with `Two_Factor_FIDO_U2F` -- could we ensure that this works with all providers by default?

assets/js/force-2fa.js
```diff
@@ -0,0 +1,88 @@
...
+	event.preventDefault();
+
+	formData.action = 'two_factor_force_form_submit';
+
+	// Submit data to WordPress.
+	jQuery.post(
+		ajaxurl,
+		formData,
+		function () {
+			window.location.reload();
```
kasparsd: If we're not using the AJAX response, could we just skip all AJAX/serialization concept and point the form to the login URL?

class.two-factor-force.php
```diff
@@ -0,0 +1,488 @@
...
+
+		return array_keys( $safe_roles );
+	}
+
+	/**
+	 * Register settings for a single site.
+	 */
+	public static function register_single_site_force_2fa_options() {
+		// Add a new setting group for forcing 2fa in General Options.
+		add_settings_section(
```
kasparsd: Should we move the plugin settings into own file since we could then re-use them for other parameters?

kasparsd (changes requested):
Amazing work @mikeselander! I had a few suggestions and notes inline the pull request.

We could really simplify the implementation if we just redirected users to their profile page and didn't allow any other admin view until they configure the two factor. The current AJAX based implementation feels relatively fragile and coupled to existing providers.

What do you think?

assets/js/force-2fa.js
mikeselander: I don't understand how this would support more browsers - `addEventListener` has excellent support: https://caniuse.com/#feat=addeventlistener

assets/js/force-2fa.js
mikeselander: I'd really rather not - anonymous functions are a scourge to test and/or maintain and make the code far less readable. I would be happy to make a "namespace" via variable although the plugin would be better to add a build  pipeline from ES6 to address this kind of issue instead.

assets/js/force-2fa.js
mikeselander: See the above comment on compatibility, it really doesn't improve things much if at all.

class.two-factor-force.php
mikeselander: Yes, we probably could to that 👌

class.two-factor-force.php
mikeselander: Well, this code exists because the Fido provider is the only one that handles enqueueing like this iirc so there would be almost no benefit to other providers.

How assets are handled from one provider to the other should likely be standardized one way or the other and then this logic could be totally stripped or updated.  I'll add an `@todo` line here for now if that's OK?

class.two-factor-force.php
mikeselander: Good plan 👍

mikeselander:
@kasparsd thank you for the comments! I know this was a lot to read through so I appreciate it 🙏 

> We could really simplify the implementation if we just redirected users to their profile page and didn't allow any other admin view until they configure the two factor. The current AJAX based implementation feels relatively fragile and coupled to existing providers.

I considered that approach (certainly would have saved me a lot of time 😄) but I think the downfall is in the user experience. If I am a user and I am constantly forced into a page with 20-30 input boxes, I'm going to end up confused. 

Even if I am shown a banner I'm unlikely to read it thoroughly enough to understand and if I'm scrolled down to the section there will still be other controls in sight so it won't be clear to me what to do, exactly. I felt that the only way to be *very* clear what needed to happen was to display an isolated page with only the necessary information displayed.

I'd be willing to look at a non-AJAX approach if you have ideas for something that supported the same isolated view or clarity of action!

kasparsd (commented):
Thanks for the follow-up @mikeselander!

> I considered that approach (certainly would have saved me a lot of time 😄) but I think the downfall is in the user experience.

I completely agree -- the inline UI is really clear for users and also looks great in the UI video/gif you posted. My main concern is with providers that require more steps to configure the implementation such as FIDO U2F or others that might redirect you to an outside page and then do a callback redirect to the profile page with some key, for example.

> I'd be willing to look at a non-AJAX approach if you have ideas for something that supported the same isolated view or clarity of action!

I think the following combination could work:

- After a successful login, redirect users to their profile page and show a pop-up overlay stating they must enable one of the Two Factor authentication methods.

- Automatically scroll to the Two Factor settings in the footer of the page once they dismiss the message.

- Show a notice at the top of the page and below the Two Factor Settings header that says the same.

- Display a "success" message at the top of the page once they've enabled one of the Two Factor providers (probably not the Dummy one).

An AJAX-based implementation would also be potential harder to debug as users would report issues on the WP.org forums.

Looking forward to hearing your thoughts and ideas.

mikeselander:
@kasparsd sorry for the slow reply - this has been a busier week than I expected but I will swing back to this.

rmccue:
@mikeselander Do you have time to continue on this, or shall I pick it up? :)

mikeselander:
@rmccue I won't have time to spend on this for a quite a while yet :( I wouldn't argue if you had time to pick this up!

alternativesurfer:
> 
> 
> Fixes #255.
> 
> Adds a network-level and site level (if non-multisite) setting to require two-factor authentication on a per-role or global level. If 2fa forcing is on and a user matches the criteria, we force the user into a 2fa settings screen and they cannot use the site until they add valid 2fa to their account.
> 
> **Super administrators can edit the settings with this interface via network (or site if single site) settings:**
> screen shot 2018-09-04 at 1 40 41 pm
> 
> **Users who fall within the required buckets will see an interface like this:**
> login-wout-2fa

I don't see where to specify these requirements in Two Factor, or is this just a mock-up / in a dev-build right now?

mikeselander:
@alternativesurfer you've checked out this branch, correct? That UI will only work with this Pull Request checked out - it's not yet landed in the plugin.

adambirds:
When will this be finished? Kind of critical before we can use the addon. Thanks

mikeselander:
@adambirds I personally do not have much time for this and will not for several months. If someone else can pick this up it would get done much faster.

naomicbush:
What we've done for this is just use the front-end shortcode from #261 (fixed on our https://github.com/wpgwinnett/two-factor master branch). 

Here's our process:

* choose the roles that need to have forced 2FA
* choose the page that has the front-end shortcode
* when someone logs in, check their role and if it needs 2FA, redirect them to the page with the front-end shortcode

They can't go anywhere on our site until they setup 2FA. 

screely-1558726048742



This meets @kasparsd's requirement of 

> simplify the implementation...and didn't allow any other...view until they configure the two factor. The current AJAX based implementation feels relatively fragile and coupled to existing providers. 

and

@mikeselander's concern for

> isolated view or clarity of action

Happy to submit a PR if this works for you.

_@adambirds does this help?_

Onyx808:
What's the status on this feature?

BrookeDot:
On WordPress VIP we do this with filters instead of providing a UI. Just pointing this out in case it's useful to others: https://github.com/Automattic/vip-go-mu-plugins/blob/da7fcd24b773e1957687c6eb56f0b56ddd402d9e/two-factor.php#L56

cvbrian:
What's the current status of this feature?

flowdee:
Would love to have this merged +1

jeffpaul:

jeffpaul:
Per [yesterday's bug scrub](https://wordpress.slack.com/archives/C04L4KCUB/p1648040434425739), given our narrow focus in 0.8.0 on the U2F deprecation we're going to punt this to a future release but note that it is a worthwhile concept.

alexclst:
I look forward to when this gets included. Any updates on the state of this?

jeffpaul:
@alexclst this is still in need of someone picking up the existing work in the PR and updating based on feedback in https://github.com/WordPress/two-factor/pull/239#pullrequestreview-153752223.

iandunn:
I agree with Kaspars that this would be much simpler if we just redirected to the profile page instead of setting up an interstitial UI. I also agree with Mike, though, about the poor UX of that page.

I think the best way forward would be to:

1. Remove all the extra code in this PR, and only do the redirect. That would make this simple and lean.
2. Open a new issue to improve the UI/UX, including maybe moving it to its own page. That would improve it for all flows, instead of just the one where 2FA is being enforced.

rmccue:
There's a fairly significant disadvantage to using the profile page, which is that it requires rendering the entire admin. This potentially increases the attack surface and could lead to undesirable behaviour for plugins that aren't aware of the 2FA plugin. Keeping it siloed either on the frontend or login ensures nothing from the admin is accidentally leaked that way.

iandunn:
🤔 Wouldn't removing the user's capabilities solve that (assuming the plugin is written correctly)? Or are you thinking of something different?

joehoyle:
I've fixed the merge conflicts here!

joshbetz:
Maybe this is obvious, but removing caps only serves to motivate people to enable Two Factor. It doesn't enhance security because anyone with the username and password can just enable 2FA to get full access. If you have a privileged user that never enables 2FA (maybe they don't login for a long time), their account is still vulnerable via a pwned password.

I would suggest an alternative way to think about this might be to enable Email by default for anyone that doesn't have other factors and then require everyone to have at least 1 factor enabled. (So you could disable email after enable TOTP, for example.)

I know this gets complicated if you want to enable Email as a factor, so maybe the default factor could be configurable. For example, maybe you want an admin to be able to generate a backup code and deliver it out of band.

joshbetz:
Could be as simple as:

```
	add_filter( 'two_factor_enabled_providers_for_user', 'force_2fa', 10, 2 );
	function force_2fa( $enabled, $user_id ) {
		if ( ! $force_2fa ) {
			return $enabled;
		}
		
		if ( count( $enabled ) ) {
			return $enabled;
		}

		return [ 'Two_Factor_Email' ];
	}
```

NewJenk:
> Could be as simple as:
> 
> ```
> 	add_filter( 'two_factor_enabled_providers_for_user', 'force_2fa', 10, 2 );
> 	function force_2fa( $enabled, $user_id ) {
> 		if ( ! $force_2fa ) {
> 			return $enabled;
> 		}
> 		
> 		if ( count( $enabled ) ) {
> 			return $enabled;
> 		}
> 
> 		return [ 'Two_Factor_Email' ];
> 	}
> ```

I think there's a minor typo here, guessing $force_2fa should be $user_id, like this:

```
add_filter( 'two_factor_enabled_providers_for_user', 'force_2fa', 10, 2 );
function force_2fa( $enabled, $user_id ) {
	if ( ! $user_id ) {
		return $enabled;
	}

	if ( count( $enabled ) ) {
		return $enabled;
	}

	return [ 'Two_Factor_Email' ];
}
```

But yea, I've quickly tested this locally and it seems to work great - a really nice way to quickly and conveniently enforce email 2fa. It also automatically pre-selects email on the user settings screen, so if a user wants to change 2fa settings they'll see that email is pre-selected.

alexclst:
There should be a way to hard-code the force settings from `wp-config.php`, so that we can enforce these in a way that normal admin users cannot change. While filters are more capable, I think that also having a define we can add would further harden security. Otherwise one admin could get 2FA set up, and then disable the requirement right away.

jeffpaul:
@iandunn @joehoyle @georgestephanis is there a preferred approach to either getting this updated and towards merge (or perhaps closed with a fresh PR and new approach and then towards merge)?  If so, I can try and help pull some engineering time to get this to `done`.

eric-michel:
We've been using the "use email if nothing selected" method of forcing 2FA:
```
function force_2fa_for_users($providers, $user_id) {
    // Force Email provider if no providers are enabled
    if (empty($providers) && class_exists('Two_Factor_Email')) {
      $providers[] = 'Two_Factor_Email';
    }
    
    return $providers;
  }
```

This works quite well (although I recently found that if the user selected FIDO U2F but enters no keys, 2FA gets disabled) Ultimately, I think an official "require 2FA feature" could be barely more complicated than the function I've shown above. Adding a filter to turn it on + the ability for a developer to set a `wp-config` setting and forcing email as a fallback method when no other is available is fully sufficient.

I'd go so far as to say requiring 2FA should be on by default with the option to disable that feature, but I understand not wanting to change the behavior for existing installations.

BrookeDot:
@masteradhoc Thanks for pointing to that fork. I took a quick look so may have missed it, but does the HM fork include a filter to enable/disable this functionality?

My quick read shows that it's enabled by default. While in the case of this feature plugin it should likely be a simple opt-in. My general preference would be with a filters for minimum capability and enabled/disabled, but also I can see the value of having it in the UI. 

In either case, I think a revisit of #255 and a fresh PR is the way to go. I would like to see agreement on:

- Filters and WP-config option or just filters (I vote filters only)
- Enabled or disabled by default?  (I vote disabled)
- Include UI (I vote yes)

@jeffpaul as you were the most recent comment in regards to a fresh PR. and @mikeselander as you opened the original PR even if by your own admission it was supposed to be as a draft PR. 

If we can get some consensus I can take a look at likely forking the Human Made approach back into core but will likely need some help writing the tests.

masteradhoc:
@BrookeDot your help would be very appreciated! I think as well that we want to:
- give admins in the settings page an option to force it for user roles of their choice
- filter to activate/disable this behaviour is already there as far as i saw but good to doublecheck
- constant over wp config is for sure a great idea

So far my opinion :)

BrookeDot:
> filter to activate/disable this behaviour is already there as far as i saw but good to doublecheck

🤦‍♀️  The file was callapsed because it was new, but does contain the main logic with filters:
https://github.com/humanmade/two-factor/blob/252e3e0891ecf08adaf910d1ea5debd20ad39170/class.two-factor-force.php

Looking at:

``` 
/**
	 * Get which user roles have two-factor forced.
	 *
	 * @since 0.1-dev
	 *
	 * @return array
	 */
	public static function get_forced_user_roles() {
		/**
		 * User roles which have two-factor forced.
		 *
		 * @param array $roles Roles which are required to use 2fa.
		 */
		return (array) apply_filters( 'two_factor_forced_user_roles', get_site_option( self::FORCED_ROLES_META_KEY, false ) );
	}
```

Brings up the question of if this should be role based (easier to understand for the average user) or if it should be cap based (simplified implementation, as one could just do `edit_posts` for example instead of selecting multiple roles)

masteradhoc:
@BrookeDot i think user roles would be quite easier to understand from a UX point.

eric-michel:
> What may be a middle ground, is on install enable it for every role that has edit_posts capability or above.

That would absolutely work for our purposes. I admit to being pretty myopic on this topic because we build mostly static sites that don't have subscriber accounts. Almost all user accounts are admins, so having a single account with 2FA turned off is as dangerous as having them all turned off (if that one gets compromised). I totally get wanting to give users without admin privileges the option rather than forcing it (since the danger a given compromised account could pose is low).

adriantoll:
I would argue against requiring 2FA by default, even on new installations. Our use case is making 2FA available to those editors who want it on, but requiring it for administrators. The extra friction of an email with a code is going to frustrating for many of our editors, whereas all our administrator accounts have TOTP enabled, which is a piece of cake with a password manager. The kinds of people we work with - not particularly tech-literate - would have no idea about how to set up TOTP codes so they'd be stuck with email codes. I know we could turn it off on new installations, but from my point of view it would be an extra bit of hassle that I'd rather not have to think about. 

Less selfishly and more generally I think changing the behaviour of a plugin that's been around for this long would cause issues - I can't imagine a good way of notifying people of that change. If you had a persistent notification in the dashboard after a new installation that would be annoying because it's only relevant the first time someone sees it.

BrookeDot:
> I don't believe @BrookeDot is suggesting defaulting it to required for users without an admin selecting to.

I was suggesting a default (that could be filtered) of `edit_post` capability. Once the plugin is enabled MFA is required for anyone who could update a post.

However, after the discussion here about how caps are confusing to most non-devs I would revise the default to be role based, this isn't as good because it doesn't account for custom roles, so perhaps there is a hybrid approach. 

> But to be clear: if this option goes in, the default behavior should be no-change and not enforcing 2fa for every account, it's just an option for site admins to require it.

I do think that is the most friction less way, especially if this is added to core. Despite my previous commons (and preference) that it's enforced, for a project this size that may not make sense.

dknauss:
Reviewed this diff closely as prior art for #846. Three defects worth flagging so they aren't carried forward:

**Super Admin can't actually be selected.** `global_force_2fa_by_role_field()` renders a checkbox for `super-admin` by merging it into `get_editable_roles()`, but `validate_forced_roles()` whitelists against `array_keys( get_editable_roles() )`, which never contains `super-admin`. The selection is silently discarded on save — while `is_two_factor_forced()` does check for that role. Broken end to end.

**`readonly` on a checkbox is a no-op.** The "universally forced disables the per-role checkboxes" behavior is cosmetic; the inputs remain interactive and submittable.

**Front-end requests are redirected.** `maybe_redirect_to_2fa_settings()` is hooked to `parse_request` as well as `admin_init`, so a logged-in subscriber reading a post is bounced into wp-admin.

Separately, the AJAX layer (`assets/js/force-2fa.js` plus the `two_factor_ajax_options_update` action) predates REST-based TOTP setup. Serializing the profile form and POSTing to `admin-ajax.php` would no longer provision TOTP, since that now goes through `two-factor/1.0`.

PR #240, opened 2018-09-04
Incremental Docblock Updates

mikeselander:
There is quite a bit of missing docblock information that makes it difficult to efficiently run tests and for IDEs to parse what each function/method should be doing and when. 

This is a small, incremental improvement in documentation that gets the documentation slightly closer to being useful and effective.


mikeselander:
Also worth noting that the failure looks like a one-offComposer failure on one version of PHP. The rest pass.

PR #248, opened 2018-09-17
ID instead of WP_User was passed to get providers for current user.

weactivist:
When editing a user other then the current user wrong values are fetched. This problem occurs because of wrong value being passed to method "get_available_providers_for_user".


kasparsd:
@weactivist This is great, thanks for your contribution.

Could you please add tests that verify the updated functionality?

PR #252, opened 2018-09-26
Add support for deleting TOTP keys

### LINKED ISSUE #138: TOTP: Cannot regenerate secret key
There doesn't seem to be away of regenerating a secret key for the TOTP.

If a device is lost you should be able to regenerate a new key so that device cannot be used to obtain access to the site.

kasparsd:
Fixes #138.

- [x] Introduce helpers for fetching, updating and deleting user TOTP secret keys.
- [x] Add a "Reset Key" button to remove a TOTP key from the user account.

reset-key-button

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -6,7 +6,7 @@
 
 	<rule ref="WordPress-Core">
 		<exclude name="Generic.Formatting.MultipleStatementAlignment" />
-	<rule>
+	</rule>
 	<rule ref="WordPress-Docs">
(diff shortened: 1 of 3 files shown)


phpcs.xml
```diff
@@ -4,7 +4,9 @@
 
 	<config name="installed_paths" value="vendor/wp-coding-standards/wpcs" />
 
-	<rule ref="WordPress-Core"/>
+	<rule ref="WordPress-Core">
+		<exclude name="Generic.Formatting.MultipleStatementAlignment" />
```
kasparsd: Will make for smaller diffs.

Issue #256, opened 2018-10-31
Can't login when using login_url filter


cameronjonesweb:
I've got a site that uses the `login_url` filter to set the login URL to a page with a login form shortcode, in this instance provided by EDD. When attempting to log in, the 2FA screen shows, but when you enter the code it just goes back to the login form. If I edit the source of the 2FA screen form and change the action to go to wp-login.php instead of the custom login form the login works correctly

kasparsd:
Thanks for reporting the issue @cameronjonesweb!

This seems to be a duplicate of #222. Let me know if that's not the case.

cameronjonesweb:
I don't really think so, but the other ticket seems a bit more ambiguous so it may be.

For my specific issue, I believe this could be fixed by adding a new filter for the form action here: https://github.com/georgestephanis/two-factor/blob/master/class.two-factor-core.php#L338. That way, I can filter the action of the form so that logins will work correctly

kasparsd:
@cameronjonesweb That form is already using the filtered login URL:

https://github.com/georgestephanis/two-factor/blob/6d14edcaa11c0745efa1dc35f3b8b134f0508827/class.two-factor-core.php#L319

For re-using this form on other pages we would need to remove all the extra [login page header](https://github.com/georgestephanis/two-factor/blob/6d14edcaa11c0745efa1dc35f3b8b134f0508827/class.two-factor-core.php#L328) and [footer HTML](https://github.com/georgestephanis/two-factor/blob/6d14edcaa11c0745efa1dc35f3b8b134f0508827/class.two-factor-core.php#L409-L415). We would also have to make sure that all processing attached to [`login_form_{$action}`](https://developer.wordpress.org/reference/hooks/login_form_action/) is firing as expected on those custom pages:

https://github.com/georgestephanis/two-factor/blob/99d654404d2bd908c710649bfc04a37274c23b6b/class.two-factor-core.php#L41-L42

Currently only the `wp-login.php` file has that action firing.

cameronjonesweb:
That's not what I meant at all. The fact that you're using the filtered value is what's causing the issue, when that URL for the form action needs to be wp-login.php to work. I'm suggesting adding another filter so that the form action URL can be fixed without changing the value that `wp_login_url` should be

kasparsd:
Thanks for pointing me in the right direction. Looks like core has it hard-coded to `wp-login.php`:

[...]

PR #257, opened 2018-11-02
Fix a login issue on WP-Engine hosting

### LINKED ISSUE #201: This is not working on a WP Engine site
When I try to authenticate with the code form email I get  502 Bad Gateway
nginx

Ideas?

ccoley:
WP-Engine requires additional parameters to exist on the URL for POST requests, and it applies those using a filter on the site_url hook when using the login_post scheme. So I've made sure to apply that filter when running in a WP-Engine environment.

Fixes #201

### DIFF
--- a/class.two-factor-core.php
+++ b/class.two-factor-core.php
@@ -318,6 +318,12 @@ class Two_Factor_Core {
 		$interim_login = isset( $_REQUEST['interim-login'] ); // WPCS: override ok.
 		$wp_login_url = wp_login_url();
 
+		// If we're running in a WP-Engine environment then we need to apply the
+		// site_url filter with the login_post scheme for the form action.
+		if ( function_exists( 'is_wpe' ) && is_wpe() ) {
+			$wp_login_url = apply_filters( 'site_url', $wp_login_url, null, 'login_post', null );
+		}
+
 		$rememberme = 0;
 		if ( isset( $_REQUEST['rememberme'] ) && $_REQUEST['rememberme'] ) {
 			$rememberme = 1;


class.two-factor-core.php
```diff
@@ -318,6 +318,12 @@ public static function login_html( $user, $login_nonce, $redirect_to, $error_msg
 		$interim_login = isset( $_REQUEST['interim-login'] ); // WPCS: override ok.
 		$wp_login_url = wp_login_url();
 
+		// If we're running in a WP-Engine environment then we need to apply the
+		// site_url filter with the login_post scheme for the form action.
+		if ( function_exists( 'is_wpe' ) && is_wpe() ) {
```
kasparsd: @ccoley This is a host specific integration and the function name is pretty generic and could conflict with existing code bases.

In #256 @cameronjonesweb suggested to switch to `site_url()` for the form action URLs to match how the [WP core does it](https://github.com/WordPress/WordPress/blob/da7a80d67fea29c2badfc538bfc01c8a585f0cbe/wp-login.php#L1082), which should also resolve this issue.

ccoley: @kasparsd I was originally going to use `site_url()` but I looked back in the code and saw that you were originally using that, but then changed it to use `wp_login_url()` in [this pull request](https://github.com/georgestephanis/two-factor/pull/62/files/bce8bf33e8177839d1027099f5fe33abd8e9191b#diff-fa3c61990ea53ab19ab294efedcfeffa) to fix WooCommerce login forms and issue #61.

I thought that my way of doing it was less likely to cause negative side effects, but I'm happy to use `site_url()` instead if that is what you want.

ccoley: Also, if the `is_wpe()` function is too generic, then we could also check for existence of the `IS_WPE` environment variable.

kasparsd: @ccoley That is great find regarding the #62!

All of the two-factor processing currently relies on the `wp-login.php` logic [and hooks](https://github.com/georgestephanis/two-factor/blob/99d654404d2bd908c710649bfc04a37274c23b6b/class.two-factor-core.php#L41-L42) so we must `POST` the form to `wp-login.php`. I'm not sure it has ever worked with a filtered `wp-login.php` destination. Will test it with a local install of WooCommerce.

kasparsd: [Confirmed](https://github.com/georgestephanis/two-factor/issues/256#issuecomment-435565107) that WooCommerce on-page login form works as expected with the `wp-login.php` form action.

kasparsd:
Fixed in #258.

PR #258, opened 2018-11-03
Login action URL

### LINKED ISSUE #256: Can't login when using login_url filter

kasparsd:
Fixes #256 and #257.

- Hardcodes the two-factor form action URL to `wp-login.php` (similar to [WP core](https://github.com/WordPress/WordPress/blob/da7a80d67fea29c2badfc538bfc01c8a585f0cbe/wp-login.php#L1082)) since all of the two-factor processing relies on [`login_form_{$action}` hooks](https://github.com/georgestephanis/two-factor/blob/99d654404d2bd908c710649bfc04a37274c23b6b/class.two-factor-core.php#L41-L42).

- Lots of WPCS fixes related to this simple change and the surrounding form HTML.

- Tested that this doesn't introduce regression with custom login forms outside `wp-login.php`, such as WooCommerce.

### DIFF
--- a/class.two-factor-core.php
+++ b/class.two-factor-core.php
@@ -378,9 +378,7 @@ class Two_Factor_Core {
 					</a>
 				</p>
 			</div>
-			<?php
-			elseif ( 1 < count( $backup_providers ) ) :
-			?>
+			<?php elseif ( 1 < count( $backup_providers ) ) : ?>
 			<div class="backup-methods-wrap">
 				<p class="backup-methods">
 					<a href="javascript:;" onclick="document.querySelector('ul.backup-methods').style.display = 'block';">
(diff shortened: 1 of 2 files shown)


kasparsd:
@cameronjonesweb, @ccoley Could you please do a code review and test if this resolves both #256 and #257?

ccoley:
@kasparsd I've tested this and it resolves #257.

PR #267, opened 2018-12-22
Add multisite support for FIDO U2F

### LINKED ISSUE #102: FIDO U2F doesn't support WordPress Multisite
FIDO U2F maps website URI (App ID) with the Security key which makes it fail on WordPress Multisite where each site would need to re-register the key to make it work.

I'm not sure if there is any solution since WordPress doesn't support any kind of SSO solution by default.

kasparsd:
Fixes #102.


kasparsd:
Closing since U2F FacedID doesn't support sites on different domains, see [this comment](https://github.com/georgestephanis/two-factor/issues/102#issuecomment-449560179).

PR #271, opened 2019-03-19
Disable XML-RPC and REST login by default

### LINKED ISSUE #87: XMLRPC bypasses 2fa
It looks like XMLRPC uses `authenticate()` directly. The app password system works great with this, as it hooks into the `authenticate` filter. However, the rest of the plugin that ​_enforces_​ 2fa hooks into `wp_login` which fires in `wp_signon()` but NOT in `authenicate()`. Basically, you can sign in via XMLRPC with a user/pass OR a user/appPass. Both work and neither needs 2fa

kasparsd:
Fixes #87.

- [x] Disable XML-RPC and REST endpoint login for users for at least one two factor method enabled. Use the `authenticate` filter which is used by both [XML-RPC](https://github.com/WordPress/WordPress/blob/4f861b95489387d0fd3dd1d97f7c5ddbce8e0db6/wp-includes/class-wp-xmlrpc-server.php#L275) and REST login endpoints.

- [x] Introduce a filter `two_factor_user_api_login_enable` to override this behaviour.

- [x] Use [wp-dev-lib](https://github.com/xwp/wp-dev-lib) at a specific version now that it uses semver.

- [x] Add tests for `is_user_api_login_enabled()` and `is_api_request()` helpers.


phpcs.xml
```diff
@@ -7,19 +7,15 @@
...
+
 	<rule ref="WordPress-Docs">
 		<exclude name="Squiz.Commenting.FileComment.MissingPackageTag"/>
 		<exclude name="Squiz.Commenting.FileComment.Missing"/>
 		<exclude name="Squiz.Commenting.FileComment.WrongStyle"/>
 		<exclude name="Squiz.Commenting.FileComment.SpacingAfterComment"/>
 		<exclude name="Squiz.Commenting.ClassComment.Missing"/>
 	</rule>
-	<rule ref="WordPress-Extra">
-		<exclude name="WordPress.CSRF.NonceVerification" />
```
kasparsd: These were removed in 2.0 of WPCS.

jikamens:
So, just to be clear, this means I can't use the Android WordPress app if I have two-factor authentication enabled on my account, is that correct?

georgestephanis:
Unless you go by way of Jetpack, or install the application passwords plugin and use an app pass for it.

jikamens:
Is that plugin being maintained? Application password revocation doesn't work. It looks like someone reported this issue in a wordpress.org forum topic almost three years ago and it was never resolved.

jikamens:
Also: I installed the application password plugin, confirmed that it was working with xmlrpc.php, then upgraded to the current version of the two-factor plugin, and no the application password that was working before is no longer working.

So your assertion that I can use an application password to continue using the Android Wordpress even though my account has two-factor enabled on it is apparently incorrect.

kasparsd:
@jikamens With the Application Passwords plugin installed, you can choose to disable the two-factor hook that prevents logins without an actual second factor like so:

```php
add_filter( 'two_factor_user_api_login_enable', '__return_true' ); 
```

Note that the Application Passwords plugin won't disable the API login with the regular user passwords -- [it just checks the application passwords in addition to the regular user password here](https://github.com/georgestephanis/application-passwords/blob/b519188c28a197660b6843eaccb5fbeadd29ff8e/class.application-passwords.php#L345-L385).

kasparsd:
@georgestephanis Can you please add me to the Application Passwords as a contributor? Thanks!

jikamens:
>With the Application Passwords plugin installed, you can choose to disable the two-factor hook that prevents logins without an actual second factor like so:
>
>`add_filter( 'two_factor_user_api_login_enable', '__return_true' );`

My understanding is that I don't need Application Passwords installed to do this. I can just do this without Application Passwords, and then I will be able to continue using the API just as I was before the newist release of the two-factor plugin.

I use a password manager and my login password for the blog is an extremely long, random string. I also use Wordfence and fail2ban on top of that to prevent brute-force attacks. So I'm not particularly worried about someone brute-forcing my password, even if I weren't using the two-factor plugin. Therefore, given that the newest release of the two-factor plugin makes it harder for me to use the Wordpress Android app, I guess I'll just turn it off.

rmccue:
Just to check, does this affect third-party authentication methods? IMO, it should really only affect the Basic Auth method for the REST API, as that's the only one which uses passwords.

kasparsd:
@rmccue It stops the `wp_authenticate_username_password()` process [via the `authenticate` filter](https://github.com/georgestephanis/two-factor/blob/ebbdaf3692deff5794d181e21944679f2dd3fd2c/class.two-factor-core.php#L254-L263) if the user logging in has 2FA enabled _and_ the current request is [an "API request"](https://github.com/georgestephanis/two-factor/blob/ebbdaf3692deff5794d181e21944679f2dd3fd2c/class.two-factor-core.php#L281-L291).

So I think [Application Passwords](https://github.com/georgestephanis/application-passwords) won't work for users with 2FA enabled because it also uses the `authenticate` filter unless  [`two_factor_user_api_login_enable`](https://github.com/georgestephanis/two-factor/blob/ebbdaf3692deff5794d181e21944679f2dd3fd2c/class.two-factor-core.php#L273) is enabled.

I'm not sure if this is the best approach but core idea was to disable XML-RPC for users with 2FA enabled since it uses Basic Auth with the user password.

PR #281, opened 2019-04-16
Add l10n support and complete missing translations.

Sonic853:
I created a two-factor.pot file using wp-cli to make it easy to generate mo files.


includes/function.login-header.php
```diff
@@ -40,7 +40,7 @@ function login_header( $title = 'Log In', $message = '', $wp_error = null ) {
 	$login_title = get_bloginfo( 'name', 'display' );
 
 	/* translators: Login screen title. 1: Login screen name, 2: Network or site name */
-	$login_title = sprintf( __( '%1$s &lsaquo; %2$s &#8212; WordPress' ), $title, $login_title );
+	$login_title = sprintf( __( '%1$s &lsaquo; %2$s &#8212; WordPress', 'two-factor' ), $title, $login_title );
```
kasparsd: This file is currently being synced with the [same file in WordPress core](https://github.com/WordPress/WordPress/blob/60670888486617c6eeba7f6d038dd821941bbda3/wp-login.php#L25-L253) which is why we're not adding the custom textdomain for these strings. WordPress should pull those translations from the core translations.

class.two-factor-core.php
```diff
@@ -58,7 +58,11 @@ public static function add_hooks() {
 	 * Sites on WordPress 4.6+ benefit from just-in-time loading of translations.
 	 */
 	public static function load_textdomain() {
-		load_plugin_textdomain( 'two-factor' );
+		load_plugin_textdomain(
+			'two-factor',
+			false,
+			dirname(plugin_basename(__FILE__)).'/languages/'
```
kasparsd: The plugin is using the translations provided by WordPress.org so we don't bundle them with the plugin.

class.two-factor-core.php
```diff
@@ -337,7 +341,7 @@ public static function backup_2fa() {
 		if ( isset( $providers[ $_GET['provider'] ] ) ) {
 			$provider = $providers[ $_GET['provider'] ];
 		} else {
-			wp_die( esc_html__( 'Cheatin&#8217; uh?' ), 403 );
+			wp_die( esc_html__( 'Cheatin&#8217; uh?', 'two-factor' ), 403 );
```
kasparsd: Thanks for adding these!

class.two-factor-core.php
kasparsd: @Sonic853 This also needs to be removes since WP core now handles the translation files.

Sonic853: Updated.

PR #284, opened 2019-04-22
Fix U2F feature detection in Firefox

kasparsd:
- [x] Fix U2F feature detection in Firefox where `window.u2f` is ready-only now. Move the check to the registration logic instead.


kasparsd:
This had just too much in there. Created #285 to fix the core issue only.

PR #288, opened 2019-05-01
Added A new Screenshot & Updated Tested Version

alihusnainarshad:
Added A new Screenshot (login screen view) & Updated Tested Version 5.5 to 5.1.1


kasparsd:
Thanks for the pull request @alihusnainarshad! The login screenshot looks great!

The markdown version of the readme is generated from `readme.txt` -- could you please move the screenshot update to that file:

https://github.com/georgestephanis/two-factor/blob/63de5fc4056308930fad0752e432606a0fbbff6a/readme.txt#L23-L26

Could you please make the file extension lowercase to match the rest of sceenshots and run it through [a PNG optimizer](https://tinypng.com/)?

The tested up to 5.1 [already covers]() version 5.1.1 in the semantic sense so we can postpone the update until 5.2 is release:

> Make note that the Tested up to field ignores minor versions, as plugins shouldn’t break with a minor update. This means you only need to define the major version it is tested against and the WordPress.org plugin directory will automatically add the minor version for you.

from https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/

alihusnainarshad:
**thanks a lot for your guidance.**  I'm closing this PR Now and will Create new One.

Have a great time.

PR #294, opened 2019-05-07
Add two_factor_totp_title filter to allow TOTP title to be changed by site-owner

### LINKED ISSUE #291: TOTP: Add filter to allow title to be changed
Currently, the title used in an Authentication app is set to `$site_name . ':' . $user->user_login` (https://github.com/georgestephanis/two-factor/blob/master/providers/class.two-factor-totp.php#L88) There are cases where it would be handy to allow a site owner to filter this.

For example, within a Multisite one may wish to show a network name instead of an individual site name. I'm thinking something like `two_factor_totp_title()` may work as a filter name.

BrookeDot:
This PR adds the `two_factor_totp_title` filter which allows the TOTP code to be filtered by a site owner. 

Example usage:

```
function my_totp_title(){
        return 'My Website';
}
add_filter( 'two_factor_totp_title', 'my_totp_title' );
```

Testing:
Add the filter to a feature plugin or your theme's `functions.php` file then re-add the QR code. 

Fixes #291

### DIFF
--- a/providers/class.two-factor-totp.php
+++ b/providers/class.two-factor-totp.php
@@ -80,12 +80,13 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 		<?php if ( empty( $key ) ) :
 			$key = $this->generate_key();
 			$site_name = get_bloginfo( 'name', 'display' );
+			$totp_title = apply_filters( 'two_factor_totp_title', $site_name . ':' . $user->user_login, $key );
 			?>
 			<p>
 				<?php esc_html_e( 'Please scan the QR code or manually enter the key, then enter an authentication code from your app in order to complete setup.', 'two-factor' ); ?>
 			</p>
 			<p>
-				<img src="<?php echo esc_url( $this->get_google_qr_code( $site_name . ':' . $user->user_login, $key, $site_name ) ); ?>" id="two-factor-totp-qrcode" />
+				<img src="<?php echo esc_url( $this->get_google_qr_code( totp_title, $key, $site_name ) ); ?>" id="two-factor-totp-qrcode" />
 			</p>
 			<p>
 				<code><?php echo esc_html( $key ); ?></code>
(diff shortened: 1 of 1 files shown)


BrookeDot:
Looks like tests for PHP 5.4 and 5.5 failed here, I can't tell if it's something in the PR or with Travis itself :(

providers/class.two-factor-totp.php
```diff
@@ -80,12 +80,13 @@ public function user_two_factor_options( $user ) {
 		<?php if ( empty( $key ) ) :
 			$key = $this->generate_key();
 			$site_name = get_bloginfo( 'name', 'display' );
+			$totp_title = apply_filters( 'two_factor_totp_title', $site_name . ':' . $user->user_login );
```
kasparsd: Could we pass `$user` as the second parameter to the `two_factor_totp_title` filter? That would simplify user related lookups during the filter.

kasparsd:
> Your server is running PHP version 5.4.45 but WordPress 5.3-alpha-45282-src requires at least 5.6.20.

Looks like we can now update `.travis.yml` to remove checks for PHP<5.6. Created #295 for that.

providers/class.two-factor-totp.php
BrookeDot: Yeah, I was thinking about that too. Just so I'm clear are you thinking something like:

`apply_filters( 'two_factor_totp_title', $site_name . ':' . $user->user_login¸ $user );`

kasparsd: Yes, exactly.

BrookeDot:
Updated the PR and waiting for Travis CI to finish now.

kasparsd:
Thanks @BrookeDot! [Version 0.4.7 is out](https://github.com/georgestephanis/two-factor/releases/tag/0.4.7) 🎉.

PR #296, opened 2019-06-24
Remove the submit action from the "Resend" button

prtksxna:
When someone enters the OTP from email and hit <kbd>Return</kbd> a new OTP is generated and emailed instead of the form being submitted. This is happening because there are two `input`s in the form that are `submit`s. The _Resend_ should be a `button`.

### DIFF
--- a/providers/class.two-factor-email.php
+++ b/providers/class.two-factor-email.php
@@ -180,7 +180,7 @@ class Two_Factor_Email extends Two_Factor_Provider {
 			<?php submit_button( __( 'Log In', 'two-factor' ) ); ?>
 		</p>
 		<p class="two-factor-email-resend">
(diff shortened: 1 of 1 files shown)


kasparsd:
Thanks for the pull request @prtksxna!

I'm unable to replicate the original issue -- pasting in the code from the email and hitting `Return` does the login instead of re-sending the code:

login-ok

This happens because the Login button is before the Resend button in the source code, as far as I understand:

https://github.com/georgestephanis/two-factor/blob/dd8853caed1ff304c1d9137c9712e8cce5f09798/providers/class.two-factor-email.php#L180-L184

Are you running the latest version 0.4.7 of the plugin? Which browser and operating system are you using?

kasparsd:
Closing because unable to replicate. Please re-open with instructions on how to replicate the issue.

PR #308, opened 2019-12-11
Improve Jetpack SSO and Two Factor compatibility

pyronaur:
When a user clicks “Login with WordPress.com”, they’re [logged in with Jetpack](https://github.com/Automattic/jetpack/blob/792b26b5539d07cc35fdd93567942f2ad481eef2/modules/sso/class.jetpack-sso-helpers.php#L246-L257), and then [immediately logged out](https://github.com/WordPress/two-factor/blob/1354d5a5f77ee3d9ebad2266920f01920bde7cbe/class.two-factor-core.php#L238) for the 2FA authentication.

The issue is that Jetpack logs in using [`wp_set_auth_cookie`](https://github.com/Automattic/jetpack/blob/e783446cabd8214aaea261fa5176018b32d9476f/modules/sso.php#L769-L771) and sets the [cookie timeout to a year](https://github.com/Automattic/jetpack/blob/792b26b5539d07cc35fdd93567942f2ad481eef2/modules/sso/class.jetpack-sso-helpers.php#L246-L257), but Two-Factor does this all over again, the second time around - based on the `rememberme` value, which is 0 in the case of a WordPress.com authentication, so when an user provides the 2FA key, they’re logged in only for a day.

Users are logged out when the session cookie expires (differs based on browser, but usually when the browser is closed) because `rememberme` is set to 0 .

This patch makes sure that Jetpack is installed and the SSO module is active, if `jetpack-sso` is set as the login action, it's going to set the `rememberme` value to `1`.

In addition, I've added a filter `two_factor_rememberme` so that the rememberme value can be modified externally, potentially for other SSO Plugins.

### DIFF
--- a/class.two-factor-core.php
+++ b/class.two-factor-core.php
@@ -806,7 +806,11 @@ class Two_Factor_Core {
 			}
 		}
 	}
-
+	/**
+	 * Determine the rememberme value
+	 *
+	 * @return int - The rememberme value
+	 */
 	public static function rememberme() {
 
 		$rememberme = 0;
--- a/class.two-factor-core.php
+++ b/class.two-factor-core.php
@@ -369,10 +369,7 @@ class Two_Factor_Core {
 		$backup_providers = array_diff_key( $available_providers, array( $provider_class => null ) );
 		$interim_login = isset( $_REQUEST['interim-login'] ); // WPCS: CSRF ok.
 
-		$rememberme = 0;
-		if ( isset( $_REQUEST['rememberme'] ) && $_REQUEST['rememberme'] ) {
-			$rememberme = 1;
-		}
+		$rememberme = self::rememberme();
 
 		if ( ! function_exists( 'login_header' ) ) {
 			// We really should migrate login_header() out of `wp-login.php` so it can be called from an includes file.
@@ -809,4 +806,25 @@ class Two_Factor_Core {
 			}
 		}
 	}
+
+	public static function rememberme() {
+
+		$rememberme = 0;
+
+		if (
+			( isset( $_REQUEST['rememberme'] ) && $_REQUEST['rememberme'] )
+			|| ( 
+				isset( $_GET['action'] )
+				&& 'jetpack-sso' === $_GET['action']
+				&& method_exists( 'Jetpack', 'is_module_active' ) 
+				&& Jetpack::is_module_active( 'sso' )
+			)
+		) {
+			$rememberme = 1;
+		}
+
+		return (int) apply_filters( 'two_factor_rememberme', $rememberme );
+
+	}
 }
+
--- a/class.two-factor-core.php
+++ b/class.two-factor-core.php
@@ -817,10 +817,10 @@ class Two_Factor_Core {
 
 		if (
 			( isset( $_REQUEST['rememberme'] ) && $_REQUEST['rememberme'] )
-			|| ( 
+			|| (
 				isset( $_GET['action'] )
 				&& 'jetpack-sso' === $_GET['action']
-				&& method_exists( 'Jetpack', 'is_module_active' ) 
+				&& method_exists( 'Jetpack', 'is_module_active' )
 				&& Jetpack::is_module_active( 'sso' )
 			)
 		) {


kasparsd:
Thanks for opening the pull request @justnorris!

> they’re logged in with Jetpack, and then immediately logged out for the 2FA authentication.

This appears to be how the two-factor plugin enforces the actual second factor input because otherwise users would be considered logged-in right after entering the password and skipping the second-factor validation.

> Two-Factor does this all over again, the second time around - based on the rememberme value, which is 0 in the case of a WordPress.com authentication, so when an user provides the 2FA key, they’re logged in only for a day.

It appears that `wp_set_auth_cookie()` [applies the `auth_cookie_expiration` filter](https://github.com/WordPress/WordPress/blob/aa919cd160edb239f8b0665361cee1c7aa0761ab/wp-includes/pluggable.php#L837-L846) for both the standard and the `remember` logins. So why isn't the `wp_set_auth_cookie()` call in two-factor:

https://github.com/WordPress/two-factor/blob/83bf79584025420714d3323da6266b7cdb64a799/class.two-factor-core.php#L645

using that Jetpack override?

justnorris:
> This appears to be how the two-factor plugin enforces the actual second factor input because otherwise users would be considered logged-in right after entering the password and skipping the second-factor validation.

Yes. And because of it, the cookies that were just set by Jetpack are now gone - because when Jetpack authenticates a user, it has hardcoded `true` for the `$remember` argument.

> So why isn't the wp_set_auth_cookie() call in two-factor [..] using that Jetpack override?

1) Jetpack [adds and removes the filter](https://github.com/Automattic/jetpack/blob/e783446cabd8214aaea261fa5176018b32d9476f/modules/sso.php#L769-L771) on the fly

--

2) In `wp_set_auth_cookie`, if `$remember` argument isn't truthy, WordPress will set `$expires` variable to 0:

https://github.com/WordPress/WordPress/blob/db18e7d50b02972431185ff842b9993770893a61/wp-includes/pluggable.php#L827-L848

It's going to use the `auth_cookie_expiration` filter to set the `$expiration`, that's used to generate the contents of the cookie:

https://github.com/WordPress/WordPress/blob/db18e7d50b02972431185ff842b9993770893a61/wp-includes/pluggable.php#L886-L892

But when the cookie is set, `$expire` is used instead:

https://github.com/WordPress/WordPress/blob/db18e7d50b02972431185ff842b9993770893a61/wp-includes/pluggable.php#L939-L944

kasparsd:
OK, so the `setcookie()` calls [are actually using](https://github.com/WordPress/WordPress/blob/aa919cd160edb239f8b0665361cee1c7aa0761ab/wp-includes/pluggable.php#L939-L944) the `$expire` values which are set to dynamic `auth_cookie_expiration` + 12h [in case of `remember` enabled](https://github.com/WordPress/WordPress/blob/aa919cd160edb239f8b0665361cee1c7aa0761ab/wp-includes/pluggable.php#L843) or to `0` when [`remember` is not selected](https://github.com/WordPress/WordPress/blob/aa919cd160edb239f8b0665361cee1c7aa0761ab/wp-includes/pluggable.php#L847).

However, the login cookie [sent by Jetpack](https://github.com/Automattic/jetpack/blob/e783446cabd8214aaea261fa5176018b32d9476f/modules/sso.php#L770) explicitly sets the `remember` to `1`.

A simple fix would be to add `rememberme=1` to the return URL from the WP.com login since that would [allow two-factor](https://github.com/WordPress/two-factor/blob/ab8d6510a17606b200d1e44a603c13b5014bac52/class.two-factor-core.php#L643-L646) and other plugins to respect that setting.

Another approach is to make the two-factor plugin check the expiration of the "regular" auth cookie and set the `remember` value based on that.

justnorris:
We just wrote the same thing using different wording at the same time :)

> A simple fix would be to add rememberme=1 to the return URL from the WP.com login since that would allow two-factor and other plugins to respect that setting.

I thought of that, but I'm not sure how simple this would be.

> Another approach is to make the two-factor plugin check the expiration of the "regular" auth cookie and set the remember value based on that.

That may work, but would probably need to parse the cookie. The method I've added in the pull request seems like a simpler solution to me.

kasparsd:
Having something generic and not coupled to specific plugins will help us in the long term.

How about something like (not tested):

```php

/**
 * Get the "remember me" value from the preliminary 
 * authentication cookie.
 *
 * @return boolean
 */
public static function auth_cookie_has_remember_me() {
    $auth_cookie = wp_parse_auth_cookie();

    if ( ! empty( $auth_cookie['expiration'] ) && (int) $auth_cookie['expiration'] > 0 ) {
        return true;
    }

    return false;
}
```

justnorris:
🤔 That might work. I'll test it out and update the patch if it works.

kasparsd:
We might need to store it in a static class property since the de-auth happens in another method:

https://github.com/WordPress/two-factor/blob/83bf79584025420714d3323da6266b7cdb64a799/class.two-factor-core.php#L233-L241

Should be fine, though.

justnorris:
The problem with this would be that if the window is reloaded for any reason - it'll lose the `rememberme` value again. Perhaps store rememberme and/or expires in a custom cookie?

justnorris:
Alternatively, we can just show the remember me checkbox: https://github.com/WordPress/two-factor/blob/master/class.two-factor-core.php#L398

kasparsd:
> The problem with this would be that if the window is reloaded for any reason - it'll lose the rememberme value again.

They would also loose the first-factor authentication state (right?) so I think we're OK.

justnorris:
Just gave this a spin. It's not going to work - cookies have to be set before they're read.

When 2FA plugin clears cookies, it just overwrites the auth cookies so that they're not persisted by the browser, but they're not yet readable by `$_COOKIE` because the user is not sending them, so at `wp_login()`,  `wp_parse_auth_cookie()` doesn't yet know that there's an auth cookie.

justnorris:
So we're back to either expose `rememberme` value with a filter and/or add Jetpack compatibility. I can probably place the Jetpack compatibility part elsewhere, for example, in VIP Go MU Plugins for VIP, but then the community versions of 2FA and Jetpack won't be compatible.

Alternatively, the `rememberme` checkbox can to be rendered so that the user is in control instead.

Or do both - add a filter and let the user control the checkbox.

kasparsd:
> I can probably place the Jetpack compatibility part elsewhere, for example, in VIP Go MU Plugins for VIP, but then the community versions of 2FA and Jetpack won't be compatible.

Do I understand correctly that the issue with Jetpack SSO is that the session TTL is ignored if used together with the Two Factor plugin? Do the rest of the features work as expected?

Ideally we shouldn't need any compatibility layer for other plugins but there is no real harm in adding it if it helps users. How about we introduce a compat-type of file similar to [how Jetpack does it](https://github.com/Automattic/jetpack/tree/725ff9d864abec8bb539ed54df3ee4b4140e61eb/modules/theme-tools/compat)?

PR #316, opened 2020-01-08
Introduce a compatibility layer for integration specific code

kasparsd:
Builds on #308.


kasparsd:
@pyronaur What are your thoughts on this implementation?

class-two-factor-compat.php
```diff
@@ -0,0 +1,48 @@
...
+	 *
+	 * @return void
+	 */
+	public function init() {
+		/**
+		 * Jetpack
+		 *
+		 * @see https://wordpress.org/plugins/jetpack/
+		 */
+		add_filter( 'two_factor_rememberme', array( $this, 'jetpack_rememberme' ) );
```
pyronaur: Maybe add a condition to only attach the hook if Jetpack is installed?

kasparsd:
Thanks for the review and feedback @pyronaur! Were you also able to test it in your dev environment?

class-two-factor-compat.php
kasparsd: Were you thinking of checking for the presence of the `Jetpack` class or one of the constants `JETPACK__VERSION` or `JETPACK__PLUGIN_FILE`? The `is_plugin_active()` check could fail if Jetpack is installed in a different directory, for example (during dev or some custom plugin tree).

pyronaur: I was thinking testing `Jetpack` class presence

pyronaur:
Give me a few moments - I'll give it a few tests.

pyronaur:
@kasparsd Tested this just now - it works!

PR #321, opened 2020-01-11
Coding Standards: Use strict comparison

akkspros:
Coding Standards: Use strict comparison


kasparsd:
Thanks for opening this @akkspros!

Looks like there are lots more fixes that can be applied automatically by running `vendor/bin/phpcbf .` -- could you please do that as part of this changeset?

akkspros:
yes i will do that

PR #324, opened 2020-01-21
Call an action when user is authenticated

Kubitomakita:
We cannot determine when the user has been actually authenticated after the second factor.

We cannot hook into the `wp_login` nor `set_logged_in_cookie` as this plugin is hooking into this action to remove authentication cookie and present the second factor form.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -645,7 +645,7 @@ class Two_Factor_Core {
 		}
 
 		wp_set_auth_cookie( $user->ID, $rememberme );
-		
+
 		do_action( 'two_factor_user_authenticated', $user );
 
 		// Must be global because that's how login_header() uses it.
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -645,6 +645,8 @@ class Two_Factor_Core {
 		}
 
 		wp_set_auth_cookie( $user->ID, $rememberme );
+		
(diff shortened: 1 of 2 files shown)


kasparsd:
This looks great, thank you for the contribution @Kubitomakita!

Could you please document the hook in the readme file? Maybe extend the [Description section](https://github.com/WordPress/two-factor/blob/e380d6a92846d60647e6195184d9948e4235fd95/readme.txt#L11-L21) of the readme with a sub-section "Hooks & Filters":

```
= Actions & Filters =

Here is a list of action and filter hooks provided by the plugin:

- `two_factor_user_authenticated` action which receives the logged in `WP_User` object as the first argument for determining the logged in user right after the authentication workflow. 
```

Kubitomakita:
Sure thing @kasparsd! Just added

kasparsd:
@Kubitomakita Looks like phpcs [found](https://travis-ci.com/WordPress/two-factor/jobs/281354742#L570) the following issue:

> /class-two-factor-core.php:648:1: error - Whitespace found at end of line (Squiz.WhiteSpace.SuperfluousWhitespace.EndLine)

Could you please fix that?

Kubitomakita:
I'm sorry, saw the Travis failed but couldn't see the error in the log... now everything's ok.

PR #340, opened 2020-03-24
PHPCS code style fixes

paulschreiber:
* whitespace cleanup
* explicitly declare function visiblity
* remove deprecated WPCS comments
* remove leading underscore from variables name
* use `rawurlencode()` instead of `urlencode()`
* use strict `in_array()` checks


phpcs.xml
```diff
@@ -6,6 +6,16 @@
...
 		<exclude name="Generic.Formatting.MultipleStatementAlignment" />
+		<exclude name="WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase" />
+		<exclude name="WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase" />
+		<exclude name="WordPress.DateTime.CurrentTimeTimestamp.Requested" />
+		<exclude name="WordPress.DateTime.RestrictedFunctions.date_date" />
+		<exclude name="WordPress.Files.FileName.InvalidClassFileName" />
+		<exclude name="WordPress.NamingConventions.ValidHookName.UseUnderscores" />
+		<exclude name="WordPress.DB.PreparedSQLPlaceholders.QuotedSimplePlaceholder" />
+		<exclude name="WordPress.DB.PreparedSQL.InterpolatedNotPrepared" />
+		<exclude name="WordPress.DB.PreparedSQL.NotPrepared" />
```
kasparsd: We probably wouldn't want to disable this for all the new code so could we disable these inline where applicable right now?

providers/class.two-factor-totp.php
```diff
@@ -377,10 +380,10 @@ public static function calc_totp( $key, $step_count = false, $digits = self::DEF
 	 */
 	public static function get_google_qr_code( $name, $key, $title = null ) {
 		// Encode to support spaces, question marks and other characters.
-		$name = rawurlencode( $name );
-		$google_url = urlencode( 'otpauth://totp/' . $name . '?secret=' . $key );
+		$name       = rawurlencode( $name );
+		$google_url = rawurlencode( 'otpauth://totp/' . $name . '?secret=' . $key );
 		if ( isset( $title ) ) {
-			$google_url .= urlencode( '&issuer=' . rawurlencode( $title ) );
+			$google_url .= rawurlencode( '&issuer=' . rawurlencode( $title ) );
```
kasparsd: `$google_url` is going to be passed as a value to the `chl` query parameter so I think `urlencode()` might be needed here to ensure this doesn't break the current functionality per https://github.com/WordPress/WordPress-Coding-Standards/issues/1234#issuecomment-561435233

composer.json
```diff
@@ -13,8 +13,9 @@
     "php": ">=5.6"
   },
   "require-dev": {
+    "dealerdirect/phpcodesniffer-composer-installer": "^0.5.0",
```
kasparsd: This isn't currently needed because we define the included standards here:

https://github.com/WordPress/two-factor/blob/1c828f90ee9461a3577c35d461d6160e93d336a6/phpcs.xml#L5

phpcs.xml
paulschreiber: I'm fixing the other errors shortly. I wanted to get the linter passing so the unit tests will run.

providers/class.two-factor-totp.php
paulschreiber: From the following comment:
"At least for modern PHP, it seems that urldecode() can handle both, so there doesn't seem to be an obvious reason to use urlencode(), when rawlurlencode() follows an RFC (unless the generated URL is going to be sent to a system that has a different version of PHP, or has a method of decoding querystrings that doesn't support %20 for spaces)."

providers/class-two-factor-backup-codes.php
```diff
@@ -294,7 +298,7 @@ public function authentication_page( $user ) {
 	 * @return boolean
 	 */
 	public function validate_authentication( $user ) {
-		return $this->validate_code( $user, $_POST['two-factor-backup-code'] );
+		return $this->validate_code( $user, filter_input( INPUT_POST, 'two-factor-backup-code', FILTER_SANITIZE_NUMBER_INT ) );
```
kasparsd: The backup codes could include alpha-numeric characters if the defaults for this helper are changed at some point:

https://github.com/WordPress/two-factor/blob/e380d6a92846d60647e6195184d9948e4235fd95/providers/class.two-factor-provider.php#L86-L96

providers/class-two-factor-totp.php
```diff
@@ -274,7 +280,7 @@ public static function is_valid_authcode( $key, $authcode ) {
 		 *
 		 * @param int $max_ticks Max ticks of time correction to allow. Default 4.
 		 */
-		$max_ticks = apply_filters( 'two-factor-totp-time-step-allowance', self::DEFAULT_TIME_STEP_ALLOWANCE );
+		$max_ticks = apply_filters( 'two_factor_totp_time_step_allowance', self::DEFAULT_TIME_STEP_ALLOWANCE );
```
kasparsd: This should also remain unchanged if we want to avoid breaking existing setups.

two-factor.php
```diff
@@ -15,10 +15,15 @@
  */
 define( 'TWO_FACTOR_DIR', plugin_dir_path( __FILE__ ) );
 
+/**
+ * Version of the plugin.
+ */
+define( 'TWO_FACTOR_VERSION', '0.5.1' );
```
kasparsd: Introducing this will add another item to the pre-release checklist to be updated but that's fine since it's in the same file as the main plugin version.

providers/class.two-factor-totp.php
kasparsd: > unless the generated URL is going to be sent to a system that has a different version of PHP, or has a method of decoding querystrings that doesn't support %20 for spaces.

We are sending this as a query string value to Google servers for generating the QR code and I remember there being issues exactly with how the query strings are parsed on their end.

In the worst case scenario the generated QR code will have an invalid authentication secret which will cause the generated codes to fail the authentication.

Could you maybe verify this locally just to confirm?

providers/class-two-factor-totp.php
paulschreiber: Given we're still at 0.x.x, let's fix this now (while people expect it to break) and add a note to the readme.

kasparsd: This plugin is used on WP VIP and 10k+ other instances as reported by WP.org. I would strongly suggest that we don't break existing installations because of a naming preference for core hooks.

Changing this for a merge into core would be totally OK, though.

Another alternative is to introduce a "deprecated" file which takes care of these legacy setups. I'm not sure that's worth the effort for this cleanup.

providers/class-two-factor-backup-codes.php
paulschreiber: Fixed.

phpcs.xml
```diff
@@ -4,30 +4,25 @@
...
-		<exclude name="Squiz.Commenting.VariableComment.Missing"/>
 	</rule>
 
+	<arg name="colors" />
+
+	<arg value="ps" />
+	<arg name="basepath" value="."/>
+	<arg name="parallel" value="50"/>
+	<arg name="extensions" value="php"/>
+	<arg name="severity" value="1"/>
```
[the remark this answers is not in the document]
paulschreiber: Yes. This will help us catch warnings, not errors.

includes/Yubico/U2F.php
```diff
@@ -169,11 +169,11 @@ public function doRegister($request, $response, $includeCert = true)
         if($tmpKey === null) {
             throw new Error('Decoding of public key failed', ERR_PUBKEY_DECODE );
         }
-        $registration->publicKey = base64_encode($pubKey);
+        $registration->public_key = base64_encode($pubKey);
```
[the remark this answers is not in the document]
paulschreiber: `includes/` is already excluded. This was a search/replace error. I've reverted the changes. We'll need to manually review the other places where `publicKey` was changed to `public_key`. Can you check those?

providers/class-two-factor-totp.php
paulschreiber: How many of those folks are using the filter? Probably a small fraction. (We don't have telemetry.)

paulschreiber:
@kasparsd This changeset is getting large. Why don't we break it up into a few smaller PRs? We can do things whitespace, docblock, visibility ffixes first, which are noisy but low-risk. Let me know if we can discuss this on the WP Slack.

providers/class-two-factor-totp.php
kasparsd: Yeah, we'll never know the exact numbers. I still think we shouldn't include this breaking change in a cleanup pull request.

kasparsd:
@paulschreiber Yes, that sounds great! Maybe move the hook updates and the QR encoding changes to a different pull request? The rest has very little potential of regression.

I pinged you on the `#core-passwords` channel on WP.org slack. Please note that asynchronous communication on pull requests and issues would still work better for me due to other commitments and timezone differences.

providers/class-two-factor-fido-u2f-admin-list-table.php
```diff
@@ -96,7 +102,7 @@ protected function display_tablenav( $which ) {
 	 */
 	public function single_row( $item ) {
 		?>
-		<tr id="key-<?php echo esc_attr( $item->keyHandle ); ?>">
+		<tr id="key-<?php echo esc_attr( $item->key_handle ); ?>">
```
kasparsd: This would need to be changed back if we're keeping WPCS out of the vendor libraries.

providers/class-two-factor-fido-u2f-admin.php
```diff
@@ -309,10 +320,10 @@ public static function delete_link( $item ) {
 	public static function wp_ajax_inline_save() {
 		check_ajax_referer( 'keyinlineeditnonce', '_inline_edit' );
 
-		require( TWO_FACTOR_DIR . 'providers/class.two-factor-fido-u2f-admin-list-table.php' );
+		require TWO_FACTOR_DIR . 'providers/class-two-factor-fido-u2f-admin-list-table.php';
 		$wp_list_table = new Two_Factor_FIDO_U2F_Admin_List_Table();
 
-		if ( ! isset( $_POST['keyHandle'] ) ) {
+		if ( ! isset( $_POST['key_handle'] ) ) {
```
kasparsd: This too would need to be reverted if we're keeping the vendor libraries away from WPCS.

providers/class-two-factor-fido-u2f-admin.php
```diff
@@ -324,7 +335,7 @@ public static function wp_ajax_inline_save() {
 		}
 
 		foreach ( $security_keys as &$key ) {
-			if ( $key->keyHandle === $_POST['keyHandle'] ) {
+			if ( $key->key_handle === $_POST['key_handle'] ) {
```
kasparsd: This would need to be reverted if we're keeping the vendor libraries away from WPCS.

providers/class-two-factor-fido-u2f.php
```diff
@@ -251,23 +257,23 @@ public static function add_security_key( $user_id, $register ) {
 
 		if (
 			! is_object( $register )
-				|| ! property_exists( $register, 'keyHandle' ) || empty( $register->keyHandle )
-				|| ! property_exists( $register, 'publicKey' ) || empty( $register->publicKey )
+				|| ! property_exists( $register, 'key_handle' ) || empty( $register->key_handle )
```
kasparsd: This would need to be reverted if we're keeping the vendor libraries away from WPCS.

tests/class-two-factor-core.php
```diff
@@ -37,41 +42,62 @@ public static function tearDownAfterClass() {
...
 
 		return $user;
 	}
 
+	/**
+	 * Clean up the dummy user object data.
+	 */
 	public function clean_dummy_user() {
-		unset( $_POST[Two_Factor_Core::ENABLED_PROVIDERS_USER_META_KEY] );
+		unset( $_POST[ Two_Factor_Core::ENABLED_PROVIDERS_USER_META_KEY ] ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
```
kasparsd: Could we disable `WordPress.Security.NonceVerification` for all tests by default?

tests/providers/class.two-factor-fido-u2f.php
```diff
@@ -1,198 +0,0 @@
-<?php
```
kasparsd: Is this being renamed? Couldn't find this in a quick glance.

kasparsd:
This is such a great work and it would be amazing to merge this in. Could we please move the filter naming changes and the QR logic changes to separate pull requests just to be sure we don't introduce a regression?

providers/class-two-factor-totp.php
paulschreiber: Done.

tests/providers/class.two-factor-fido-u2f.php
paulschreiber: class.two-factor-fido-u2f.php --> class-two-factor-fido-u2f.php

providers/class-two-factor-fido-u2f-admin.php
paulschreiber: Fixed

tests/class-two-factor-core.php
paulschreiber: Done.

class-two-factor-core.php
```diff
@@ -818,7 +827,7 @@ public static function user_two_factor_options( $user ) {
 								<th scope="row"><input type="radio" name="<?php echo esc_attr( self::PROVIDER_USER_META_KEY ); ?>" value="<?php echo esc_attr( $class ); ?>" <?php checked( $class, $primary_provider_key ); ?> /></th>
 								<td>
 									<?php $object->print_label(); ?>
-									<?php do_action( 'two-factor-user-options-' . $class, $user ); ?>
+									<?php do_action( 'two_factor_user_options_' . $class, $user ); ?>
```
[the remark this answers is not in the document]
kasparsd: @paulschreiber Looks like this action naming change is still there. Could we move it to a dedicated pull request, please?

providers/class-two-factor-email.php
```diff
@@ -42,7 +48,7 @@ static function get_instance() {
 	 * @since 0.1-dev
 	 */
 	protected function __construct() {
-		add_action( 'two-factor-user-options-' . __CLASS__, array( $this, 'user_options' ) );
+		add_action( 'two_factor_user_options_' . __CLASS__, array( $this, 'user_options' ) );
```
kasparsd, line 51: And this action name change too.

providers/class-two-factor-fido-u2f.php
```diff
@@ -83,7 +89,7 @@ protected function __construct() {
 			true
 		);
 
-		add_action( 'two-factor-user-options-' . __CLASS__, array( $this, 'user_options' ) );
+		add_action( 'two_factor_user_options_' . __CLASS__, array( $this, 'user_options' ) );
```
kasparsd, line 92: And this action name change.

providers/class-two-factor-email.php
paulschreiber: Done.

paulschreiber:
I merged `fix/phpcs` into `feature/local-dev-concept`. One test fails:
```
1) Tests_Two_Factor_Backup_Codes::test_validate_authentication
Failed asserting that false is true.
```

class-two-factor-core.php
```diff
@@ -526,29 +536,33 @@ public static function show_two_factor_login( $user ) {
 	 * @since 0.1-dev
 	 */
 	public static function backup_2fa() {
-		if ( ! isset( $_GET['wp-auth-id'], $_GET['wp-auth-nonce'], $_GET['provider'] ) ) {
+		$wp_auth_id = filter_input( INPUT_GET, 'wp-auth-id', FILTER_SANITIZE_NUMBER_INT );
+		$nonce      = filter_input( INPUT_GET, 'wp-auth-nonce', FILTER_SANITIZE_STRING );
+		$provider   = filter_input( INPUT_GET, 'provider', FILTER_SANITIZE_STRING );
+
+		if ( ! $wp_auth_id || ! $wp_auth_nonce || ! $provider ) {
```
joshbetz, line 543: Looks like `$wp_auth_nonce` should be `$nonce` here

kasparsd: Thank you so much for finding this. Looks like we didn’t catch it during the initial code review. And I couldn’t figure out why the current code was failing on my test site...

Issue #247, opened 2018-09-12
More like a request


daniel181:
I dont know where to place this request, but is it possible for a buddypress support?
and or edit the settings so you can only choice about one or more selections for the two factor like giving users only the option for either google or email and nothing more, or just one selection.

kasparsd:
Thanks for sending in your suggestions @daniel181!

So you're suggesting two features:

- Support for buddypress. Could you please elaborate on that? What needs to change in the plugin to make this happen?

- Plugin settings to disable some of the default two factor providers. This is a good one and I'll open a new issue for that.

daniel181:
sorry for late reply i was sick..

however the thing is i like your two way factor, however if you use buddypress users can never access it since all profiles will be re-linked to buddypress, so maybe there could be a way to make it also buddypress unable like something if you dont use buddypress the plugin will only be active in the user account but if using buddypress then the plugin would add a extra field in the buddypress profile setup for users to select which one the admin had setup. so that it would appear in there as a selection if the user wants to.. right now i'm using wordpress + buddypress but i have to ask users if they want a two factor setup for their account if the user can do this on thier own that would be great in buddypress.

and maybe just a thought could it also be possible for lossing phone? or is that inplanted?

Sorry for my pour english.

kasparsd:
This is almost a duplicate of #189.

daniel181:
almost, he was using the woocommence, i use buddypress.

kasparsd:
@daniel181 Looks like there is a general need for a two-factor profile shortcode component. What are the best practices in BuddyPress for adding such features? Is that through a shortcode or a certain hook?

daniel181:
i believe a component function..

olasinola:
Eagerly awaiting for the frontend shortcode feature. From the Pull Request #261 , I think apart from issues with FIDO U2F and multisites, its working for single sites.

jeffpaul:
@daniel181 is there a block-based handling for this that BuddyPress might better support/integrate with these days?

Issue #189, opened 2017-08-04
Add the possibility to activate/deactive features in front page dashboard for customers


pwallner:
I have a website with woocommerce and I block the access to the dashboard to my customers. Would it be possible to add 2FA to the front page dashboard, so my customers can activate/deactivate/delete at least google authenticator in the front page. Nice would be if they have all options (email, google, U2F, one time passwords incl. regenerate)

monecchi:
Good point @pwallner! The plugin should provide a few action/filter hooks in order to do that.

pgreivel:
That sounds like a great idea. I ran into the same issue. So would love to have this possibility too! :)

phoolto:
Would be a good idea to do this through a shortcode so you can add it to you desired location. For example add it to the WC menu.

ecgroupcom:
Add the possibility of activating / deactivating features in the main page panel for clients either in Woocommerce, BBpress, BuddyPress, WC Marketplace ... that the client activates this security in the same way, personalize each message sent if This has activated verification by mail.

DestructiveBurn:
O, dear, this was mentioned on Aug 4, 2017, and it still hasn't been implemented into the Two Factor.
This really should get done for those who block the backend dashboard and use the WooCommerce login system.   There are other two-step plugins doing it.

It's time to add it!

NicolasKulka:
+1 pour l'ajout front Woocommerce

AMorgaut:
I also want to allow Two factor connection for my woocomerce clients

It would mean either that:
- when the user goes to its "my information" page accessible from "my account", from where he can renew his password, he would find the Two factor settings option
- or better, an addidionnatal "two factor authentication' menu item would be proposed after the "my informations" one, and serve a dedication setting menu

PR #344, opened 2020-04-29
Filter the enabled providers for a user

christianc1:
This PR adds a filter `two_factor_enabled_providers_for_user` to allow developers to filter the the providers that are enabled for a specific user.

An example of where they may be useful is enabling the email provider for a subset of users.
````php
add_filter( 'two_factor_enabled_providers_for_user', function( $providers, $user_id ) {
    $user = get_userdata( $user_id );
    if ( in_array( 'editor', $user->roles, true ) {
        array_push( $providers, 'Two_Factor_Email' );
    }
    return array_unique( $providers );
}, 10, 2 );
````

As of right now, the only way to do this that I could find is to filter the user meta, which is fine, but less elegant.

Note: I kept the second parameter of the hook signature as `$user_id : int` as opposed to `$user : WP_User` for uniformity with a few other similar filters.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -193,7 +193,7 @@ class Two_Factor_Core {
 		/**
 		 * Filter the available two-factor authentication providers for this user.
 		 *
-		 * @param string $providers The available providers.
+		 * @param array  $providers The available providers.
 		 * @param int    $user_id   The user ID.
 		 */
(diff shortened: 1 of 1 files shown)


christianc1:
Also worth noting that while this PR doesn't necessarily solve the front end aspects of #189 and #247,  together with the `two_factor_primary_provider_for_user` filter, this would allow you to filter the necessary values and fully decouple the enabled/primary providers from the user profile UI.

PR #351, opened 2020-05-04
Escape the U2F key handle during lookup

kasparsd:


kasparsd:
The related method had good test coverage and they're [passing](https://travis-ci.com/github/WordPress/two-factor/jobs/327242315) so I'm merging this in.

PR #352, opened 2020-05-04
Add email token expiry

### LINKED ISSUE #220: Email backup codes should have an expiry
Currently Email Backup Codes are valid until they're used  - even if the user logs in using another backup method.

Ideally even the Email Backup Codes should have an expiry. At the minimum an expiry of a Day would be good here, although more practically an Hour would make sense to me.

kasparsd:
Fixes #220.

- Keep track of the timestamp when the token was generated.
- Allow the token TTL to be customized via a filter.

### DIFF
--- a/providers/class.two-factor-email.php
+++ b/providers/class.two-factor-email.php
@@ -249,7 +249,7 @@ class Two_Factor_Email extends Two_Factor_Provider {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


kasparsd:
Ready for code review.

kasparsd:
Thank you @mjangda!

PR #356, opened 2020-05-13
Register scripts before being enqueued

cojennin:
Registering the `fido-u2f-api` and `fido-u2f-login` scripts in the `__constructor` is generating the following notice:
`
Notice: wp_register_script was called incorrectly. Scripts and styles should not be registered or enqueued until the wp_enqueue_scripts, admin_enqueue_scripts, or login_enqueue_scripts hooks. 
`

It seems like these `wp_register_script` calls should be moved to `login_enqueue_assets` as that's where they're utilized. 

(Props @scottnelle for tracking this down)

### DIFF
--- a/providers/class.two-factor-fido-u2f.php
+++ b/providers/class.two-factor-fido-u2f.php
@@ -67,22 +67,9 @@ class Two_Factor_FIDO_U2F extends Two_Factor_Provider {
 		require_once( TWO_FACTOR_DIR . 'providers/class.two-factor-fido-u2f-admin.php' );
 		Two_Factor_FIDO_U2F_Admin::add_hooks();
 
-		wp_register_script(
-			'fido-u2f-api',
-			plugins_url( 'includes/Google/u2f-api.js', dirname( __FILE__ ) ),
-			null,
-			self::asset_version(),
-			true
-		);
-
-		wp_register_script(
-			'fido-u2f-login',
-			plugins_url( 'js/fido-u2f-login.js', __FILE__ ),
-			array( 'jquery', 'fido-u2f-api' ),
-			self::asset_version(),
-			true
-		);
-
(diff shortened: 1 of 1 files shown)


providers/class.two-factor-fido-u2f.php
```diff
@@ -130,6 +114,23 @@ public function get_label() {
 	 * @since 0.1-dev
 	 */
 	public function login_enqueue_assets() {
+		
+		wp_register_script(
+			'fido-u2f-api',
```
kasparsd, line 119: Looks like this is also referenced here:

https://github.com/WordPress/two-factor/blob/2b4ce5cd41a8859f5524f651493151baa61b7683/providers/class.two-factor-fido-u2f-admin.php#L75-L81

so we need to make sure this script is registered before it is enqueued there.

cojennin: Ah yea, made a change to the PR to register on `admin_enqueue_scripts`. But when does [login_enqueue_assets](https://github.com/WordPress/two-factor/blob/f8d2fda9c4f381c4242e66abf62486e47116d2ef/providers/class.two-factor-fido-u2f.php#L132) get called?

rinatkhaziev:
Hi @kasparsd ,

Any plans to get this merged in the next release?

kasparsd:
Fixes #361.

kasparsd:
There appear to be a few code style issues reported here https://travis-ci.com/github/WordPress/two-factor/jobs/333298974 and a merge conflict with the latest changes introduced by the latest worked merged into `master`.

Is it OK if I open a new pull request to fix these and also make sure the scripts are available on the login screen?

kasparsd:
Replacing this with #368 if that's OK.

@rinatkhaziev, @cojennin Could you please code review #368?

PR #365, opened 2020-06-17
restore space removal in authcode

paulschreiber:
previously #339

### DIFF
--- a/providers/class-two-factor-totp.php
+++ b/providers/class-two-factor-totp.php
@@ -172,7 +172,7 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 			// Validate and store a new secret key.
 			if ( ! empty( $_POST['two-factor-totp-authcode'] ) && ! empty( $_POST['two-factor-totp-key'] ) ) {
 				// Don't use filter_input() because we can't mock it during tests for now.
-				$authcode = sanitize_text_field( $_POST['two-factor-totp-authcode'] );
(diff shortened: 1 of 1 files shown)


providers/class-two-factor-totp.php
```diff
@@ -172,7 +172,7 @@ public function user_two_factor_options_update( $user_id ) {
 			// Validate and store a new secret key.
 			if ( ! empty( $_POST['two-factor-totp-authcode'] ) && ! empty( $_POST['two-factor-totp-key'] ) ) {
 				// Don't use filter_input() because we can't mock it during tests for now.
-				$authcode = sanitize_text_field( $_POST['two-factor-totp-authcode'] );
+				$authcode = filter_var( sanitize_text_field( $_POST['two-factor-totp-authcode'] ), FILTER_SANITIZE_NUMBER_INT );
```
kasparsd, line 175: Will this remove any leading zeros from codes such as `012345`? Or is the TOTP algorithm designed to not ever generate a number that starts with a zero?

paulschreiber:
```
wp> filter_var("012 345", FILTER_SANITIZE_NUMBER_INT)
=> string(6) "012345"
```

kasparsd: OK, good to now that it's not the same as `(int) '012 345'`.

PR #368, opened 2020-06-26
Register scripts during *_enqueue_scripts hook instead

### LINKED ISSUE #361: wp_register_script causes a PHP Notice
I have no idea why I just started getting this, but I just updated this [VIP Go repo](https://github.com/Automattic/vip-go-mu-plugins) in my local environment and I suddenly started getting two PHP notices. Turns out the notice came from these functions:

https://github.com/WordPress/two-factor/blob/529c973deb4677e45cc243b1336dd718aa472611/providers/class.two-factor-fido-u2f.php#L70

`Notice: wp_register_script was called incorrectly. Scripts and styles should not be registered or enqueued until the wp_enqueue_scripts, admin_enqueue_scripts, or login_enqueue_scripts hooks. Please see Debugging in WordPress for more information. (This message was added in version 3.3.0.) in /srv/www/ona-journalists/public_html/wp-includes/functions.php on line 5167
`

kasparsd:
Fixes #361.

### DIFF
--- a/providers/class-two-factor-fido-u2f.php
+++ b/providers/class-two-factor-fido-u2f.php
@@ -73,21 +73,10 @@ class Two_Factor_FIDO_U2F extends Two_Factor_Provider {
 		require_once TWO_FACTOR_DIR . 'providers/class-two-factor-fido-u2f-admin.php';
 		Two_Factor_FIDO_U2F_Admin::add_hooks();
 
-		wp_register_script(
-			'fido-u2f-api',
-			plugins_url( 'includes/Google/u2f-api.js', dirname( __FILE__ ) ),
-			null,
-			self::asset_version(),
-			true
-		);
-
-		wp_register_script(
-			'fido-u2f-login',
-			plugins_url( 'js/fido-u2f-login.js', __FILE__ ),
-			array( 'jquery', 'fido-u2f-api' ),
-			self::asset_version(),
-			true
-		);
+		// Ensure the script dependencies have been registered before they're enqueued at a later priority.
+		add_action( 'admin_enqueue_scripts', array( __CLASS__, 'enqueue_scripts' ), 5 );
+		add_action( 'wp_enqueue_scripts', array( __CLASS__, 'enqueue_scripts' ), 5 );
+		add_action( 'login_enqueue_scripts', array( __CLASS__, 'enqueue_scripts' ), 5 );
 
 		add_action( 'two_factor_user_options_' . __CLASS__, array( $this, 'user_options' ) );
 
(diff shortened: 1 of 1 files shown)


providers/class-two-factor-fido-u2f.php
```diff
@@ -131,12 +120,27 @@ public function get_label() {
...
 
 	/**
-	 * Enqueue assets for login form.
+	 * Register script dependencies used during login and when
+	 * registering keys in the WP admin.
 	 *
-	 * @since 0.1-dev
+	 * @return void
 	 */
-	public function login_enqueue_assets() {
```
kasparsd, line 138: This wasn't used anywhere.

providers/class-two-factor-fido-u2f.php
```diff
@@ -131,12 +120,27 @@ public function get_label() {
...
+		wp_register_script(
+			'fido-u2f-api',
+			plugins_url( 'includes/Google/u2f-api.js', dirname( __FILE__ ) ),
+			null,
+			self::asset_version(),
+			true
+		);
+
+		wp_register_script(
+			'fido-u2f-login',
```
kasparsd, line 138: This is used during our custom "login" page https://github.com/wordpress/two-factor/blob/a97c32b6639a3557fd149725b973615246549ffe/class-two-factor-core.php#L581-L726

providers/class-two-factor-fido-u2f.php
```diff
@@ -131,12 +120,27 @@ public function get_label() {
...
+	 * registering keys in the WP admin.
 	 *
-	 * @since 0.1-dev
+	 * @return void
 	 */
-	public function login_enqueue_assets() {
-		wp_enqueue_script( 'fido-u2f-login' );
+	public static function enqueue_scripts() {
+		wp_register_script(
+			'fido-u2f-api',
```
kasparsd, line 130: This is a dependency for `fido-u2f-admin` and `fido-u2f-login`.

providers/class-two-factor-fido-u2f.php
```diff
@@ -73,21 +73,10 @@ protected function __construct() {
...
-
-		wp_register_script(
-			'fido-u2f-login',
-			plugins_url( 'js/fido-u2f-login.js', __FILE__ ),
-			array( 'jquery', 'fido-u2f-api' ),
-			self::asset_version(),
-			true
-		);
+		// Ensure the script dependencies have been registered before they're enqueued at a later priority.
+		add_action( 'admin_enqueue_scripts', array( __CLASS__, 'enqueue_scripts' ), 5 );
```
kasparsd, line 77: Use action priority 5 to ensure the scripts have been registered before they're enqueued later at priority 10.

rinatkhaziev:
This looks good, however,  is there a particular reason why we do script registration and enqueueing separately? I may be missing some context, but it seems to me that the complexity might be reduced a bit if combined into a single `wp_enqueue_script` call.

kasparsd:
@rinatkhaziev That's a good question. I think the reason for that is that both admin and login screens rely on the same JS dependencies so we're able to reference those dependencies without having to enqueue them two multiple times and potentially mess up the file paths, versions, etc.

rinatkhaziev:
Ah I see, thanks!

PR #381, opened 2020-08-26
Fix backup authentication method variable resolution

kasparsd:
Fixes a regression introduced in #340.

Props @joshbetz!


providers/class-two-factor-totp.php
```diff
@@ -71,7 +71,7 @@ public static function get_instance() {
 	 * Returns the name of the provider.
 	 */
 	public function get_label() {
-		return _x( 'Time Based One-Time Password (Google Authenticator)', 'Provider Label', 'two-factor' );
+		return _x( 'Time Based One-Time Password (TOTP)', 'Provider Label', 'two-factor' );
```
kasparsd, line 74: There are more ways to do TOTP these days so make it generic.

providers/class-two-factor-totp.php
```diff
@@ -142,7 +142,7 @@ public function user_two_factor_options( $user ) {
 			</p>
 		<?php else : ?>
 			<p class="success">
-				<?php esc_html_e( 'Secret key configured and registered.', 'two-factor' ); ?>
+				<?php esc_html_e( 'Secret key is configured and registered. It is not possible to view it again for security reasons.', 'two-factor' ); ?>
```
kasparsd, line 145: This has been a frequent question in the support forums so we explain why the secret key is not shown again.

providers/class-two-factor-totp.php
```diff
@@ -454,6 +454,9 @@ public function is_available_for_user( $user ) {
 	public function authentication_page( $user ) {
 		require_once ABSPATH . '/wp-admin/includes/template.php';
 		?>
+		<p>
+			<?php esc_html_e( 'Please enter the code generated by your authenticator app.', 'two-factor' ); ?>
```
kasparsd, line 458: Explain where to get the code from.

PR #382, opened 2020-08-26
Ensure there are no blank spaces or line breaks around the token

### LINKED ISSUE #379: Email auth codes copy paste issue
When double-clicking an auth code to copy it, it also selects a blank space that the auth submission form detects as an extra digit.
https://prnt.sc/u40blj
https://prnt.sc/u40cqk
It's very minor issue, but also a very annoying one.
I suggest you modify it to ignore spaces, or modify the way the auth code is provided. e.g: https://prnt.sc/u40how
Thank you for the awesome plugin!

kasparsd:
Fixes #379.

### DIFF
--- a/providers/class-two-factor-email.php
+++ b/providers/class-two-factor-email.php
@@ -317,7 +317,10 @@ class Two_Factor_Email extends Two_Factor_Provider {
 			return false;
 		}
 
-		return $this->validate_token( $user->ID, $_REQUEST['two-factor-email-code'] );
+		// Ensure there are no spaces or line breaks around the code.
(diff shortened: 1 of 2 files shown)


providers/class-two-factor-email.php
```diff
@@ -269,7 +269,7 @@ public function authentication_page( $user ) {
 		<p><?php esc_html_e( 'A verification code has been sent to the email address associated with your account.', 'two-factor' ); ?></p>
 		<p>
 			<label for="authcode"><?php esc_html_e( 'Verification Code:', 'two-factor' ); ?></label>
-			<input type="tel" name="two-factor-email-code" id="authcode" class="input" value="" size="20" pattern="[0-9]*" />
```
kasparsd, line 272: Allow all kinds of inputs to avoid blocking users as they copy codes with wrapping empty spaces or line breaks from their email clients.

PR #162, opened 2017-03-28
Remove server-side U2F support detection

kasparsd:
Fixes #117 and #78 . Ready for code review.

- Removes all PHP based U2F support detection.
- Introduce error handling when registering new keys.

Here is a notification if U2F browser support can't be detected:

utf-not-supported-notice

Here are the HTTPS related notifications:

https-notification

PR #386, opened 2020-10-03
Only show "Security Keys" section if "FIDO U2F Security Keys" is enabled

r-a-y:
Hi,

This PR only shows the "Security Keys" section if the "FIDO U2F Security Keys" provider is enabled.

Here's a GIF of the PR:

GIF

If JS is disabled, the "Security Keys" section gets displayed as expected.

----

I created this PR for better UX. If the "Security Keys" section is shown, but the "FIDO U2F Security Keys" provider is not enabled, this might confuse people thinking that this might be something that needs to be filled in.

I would also recommend changing the `Requires an HTTPS connection. Configure your security keys in the "Security Keys" section below.` string. I don't think it is necessary to mention `Requires an HTTPS connection`. Either the server supports HTTPS or it doesn't. The end user doesn't care about this.

Let me know what you think.

### DIFF
--- a/providers/class-two-factor-fido-u2f-admin.php
+++ b/providers/class-two-factor-fido-u2f-admin.php
@@ -182,7 +182,7 @@ class Two_Factor_FIDO_U2F_Admin {
 		}
 
 		?>
-		<div class="security-keys" id="security-keys-section">
+		<div class="security-keys no-js" id="security-keys-section">
 			<h3><?php esc_html_e( 'Security Keys', 'two-factor' ); ?></h3>
 
 			<?php if ( ! is_ssl() ) : ?>
--- a/providers/class-two-factor-fido-u2f-admin.php
+++ b/providers/class-two-factor-fido-u2f-admin.php
@@ -182,7 +182,7 @@ class Two_Factor_FIDO_U2F_Admin {
 		}
 
 		?>
-		<div class="security-keys no-js" id="security-keys-section">
+		<div class="security-keys" id="security-keys-section">
 			<h3><?php esc_html_e( 'Security Keys', 'two-factor' ); ?></h3>
 
 			<?php if ( ! is_ssl() ) : ?>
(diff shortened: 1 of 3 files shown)


kasparsd:
Thanks for the suggestion @r-a-y!

I agree that the separate location for the security keys is confusing. Ideally, this would be displayed inline the option similar to the other providers.

One issue with the suggested approach is that users can't add or remove the keys unless the feature is enabled. I feel like the security keys section should always be available even when the feature itself is disabled.

> I don't think it is necessary to mention Requires an HTTPS connection. Either the server supports HTTPS or it doesn't. The end user doesn't care about this.

This wording was added in #162 after a lot of support requests asking why the feature doesn't work. Users didn't know that HTTPS is required on the login page for this to work. It is added only if the admin area is accessed without the HTTPS so it doesn't display for users with the correct HTTPS setup.

r-a-y:
> One issue with the suggested approach is that users can't add or remove the keys unless the feature is enabled. I feel like the security keys section should always be available even when the feature itself is disabled.

Maybe we could add a button that says "Configure" and that would toggle and scroll down to the Security Keys section instead?

> This wording was added in #162 after a lot of support requests asking why the feature doesn't work. Users didn't know that HTTPS is required on the login page for this to work. It is added only if the admin area is accessed without the HTTPS so it doesn't display for users with the correct HTTPS setup.

Ahh right, I'm testing this locally on a non-HTTPS set up, which is why I'm seeing this wording.

Perhaps the FIDO provider shouldn't even be enabled if the server doesn't support HTTPS. Shouldn't a simple `is_ssl()` suffice to make this determination? A lot has changed since #162. HTTPS is more widespread than it was three years ago.

jeffpaul:

jeffpaul:
Given that the approach for the 0.8.0 release is to fully deprecate U2F, the changes in this PR will no longer be relevant.  Apologies for the long delay on review here and thank you @r-a-y for your contribution and time.

PR #387, opened 2020-10-03
Add <label> to provider name

r-a-y:
Hi,

This PR wraps the provider name with the `<label>` element for better UX.

GIF:

GIF

I've also renamed the `Name` column to `Type` since this column also has buttons depending on the provider (TOTP, Backup Verification Codes). I can revert this if this isn't desired.


jeffpaul:
@georgestephanis / @kasparsd any thoughts on a review of this PR and whether it's ready for consideration in v0.8.0?

r-a-y:
> Given that we're looking at a larger UX review of this part of the plugin

Yeah, the Enabled / Primary UX needs to be overhauled. I'm injecting the labels with [some javascript](https://github.com/r-a-y/bp-two-factor/blob/96b62b015564af3fdfa681371ac47f187acf9da1/assets/settings.js#L52) at the moment as a workaround until the interface is reworked at some point.

kasparsd:
This is a great UX and accessibility improvement. Let's merge this in.

PR #390, opened 2020-10-20
TOTP: Allow 'Enter' key to submit the auth code form

### LINKED ISSUE #91: Pressing "Enter" instead of clicking the "Update Profile" button fails to enable the TOTP
Looks like the `personal_options_update` filter isn't triggered when using the Enter key after entering the value in `two-factor-totp-authcode` input field.

r-a-y:
Currently for the TOTP method, if you have to add an authentication code for the QR code, you have to manually click on the `Submit` button in order for the form to submit properly.

This PR allows the `Enter` key to submit the auth code form, which is better for UX.

If you'd rather have the inline JS put into an external file, let me know. Inline JS was already used in the `authentication_page()` method so I followed suit.

Fixes #91.


jeffpaul:

jeffpaul:
Per [yesterday's bug scrub](https://wordpress.slack.com/archives/C04L4KCUB/p1648040434425739), we're going to punt this to future release so that we can focus on the U2F deprecation in 0.8.0.

kasparsd:
#518 introduced automatic submit of the input if the input character count matches the expected code length, so it is now doing autosubmit.

PR #398, opened 2021-02-18
Fix: Backup code generate notice not visible.

maheshwaghmare:
Two-Factor Backup Codes generated like below:

E.g.

1. 26056637
2. 34800832
3. 68473048
4. 28263149
5. 68946117
6. 73912621
7. 70914813
8. 59360534
9. 82345347
10. 59028553

When using all of them then the notice appears. But due to the missing `echo`, it was not displayed.

Now, This displayed as below:

Error Message


kasparsd (approved):

jeffpaul:
@kasparsd I see you've approved this, anything else blocking it from getting merged in?

PR #400, opened 2021-05-05
login_site_html_link hook fix

pepijnnichting:
Use the login_site_html_link hook just like in wp-login.php

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -681,17 +681,18 @@ class Two_Factor_Core {
 		<?php endif; ?>
 
 		<p id="backtoblog">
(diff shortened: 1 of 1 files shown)


kasparsd:
@pepijnnichting I wonder if we could replace the whole footer section with a call to `login_footer()` just like WP core does it here:

https://github.com/WordPress/WordPress/blob/c6028577a462f235da67e5d3dcf1dc42f9a96669/wp-login.php#L1446

and leave just our custom styles above that call.

pepijnnichting:
I did a quick check and this seems to work! This looks like a much better solution.

kasparsd:
Thanks @pepijnnichting! I wonder why it wasn't using this before? Were you able to test the whole login flow with this update?

pepijnnichting:
@kasparsd Tested the flow for email, dummy and totp as provider and all work. I dit not check fido and backup-codes.

PR #410, opened 2021-07-21
Minor correction

sm0rux:
According https://wordpress.org/plugins/two-factor/ the plugin is tested up to WP 5.7.2 :)


kasparsd:
Thanks for the fix! This has been added as part of #415.

Issue #175, opened 2017-04-29
Breaks bbPress profile editing


kasparsd:
See https://wordpress.org/support/topic/plugin-breaks-bbpress-profile-editing/

> When you click on someone’s username in a bbPress forum, it brings up information from your WordPress profile and integrates it into your site’s theme. If you’re viewing your own profile, you can also edit it from there, without ever going into wp-admin:
> 
> When enabling Two Factor, however, it completely breaks the page, and produces this error at the bottom of a stream of unformatted text:

```
Fatal error: Uncaught Error: Call to undefined function convert_to_screen() in /home/public/wp-admin/includes/class-wp-list-table.php:143 Stack trace: 
#0 /home/public/wp-content/plugins/two-factor/providers/class.two-factor-fido-u2f-admin.php(189): WP_List_Table->__construct() 
#1 /home/public/wp-includes/class-wp-hook.php(298): Two_Factor_FIDO_U2F_Admin::show_user_profile(Object(WP_User)) 
#2 /home/public/wp-includes/class-wp-hook.php(323): WP_Hook->apply_filters('', Array) 
#3 /home/public/wp-includes/plugin.php(453): WP_Hook->do_action(Array) 
#4 /home/public/wp-content/plugins/two-factor/class.two-factor-core.php(636): do_action('show_user_secur...', Object(WP_User)) 
#5 /home/public/wp-includes/class-wp-hook.php(298): Two_Factor_Core::user_two_factor_options(Object(WP_User)) 
#6 /home/public/wp-includes/class-wp-hook.php(323): WP_Hook->apply_filters('', Array) 
#7 /home/public/wp-includes/plugin.php(453): WP_Hook->do_action(Array) 
#8 /home/public/wp-content/plugins/bbpress/includes/users/functions.php(1662): do_action('show in /home/public/wp-admin/includes/class-wp-list-table.php on line 143
```

jeffpaul:
@CryoutCreations has a fix for this in #414.

iandunn:
U2F is deprecated and no longer works in Chrome, so the provider is being removed in #439 . Given that, there's probably no reason to keep this open anymore.

iandunn:
Er, actually, looking at the PR, it isn't specific to U2F, so I'll reopen.

Issue #335, opened 2020-02-27
Make the Two Factor settings portable to other settings screens


kasparsd:
Currently the two-factor setup for each user is tied to the standard user profile page. This makes it impossible to configure the two-factor authentication for sites that don't allow users to access the standard profile page.

georgestephanis:
My main concern here would be portability of styles -- it may not render correctly.

What other plugins create user screens on the front-end, and do they have any consistent hook usage that can be used to add extra fields?  Or would it be better if we provided a function that they could use to render our stuff?

Or sort out a common action?

gytisrepecka:
Ideal situation would be to have a shortcode which would display two-factor settings.

I recently attempted to include two-factor settings into WooCommerce's `my-account/edit-account` page. I could enable/disable two-factor methods, but AJAX didn't work (recovery codes, showing QR code for TOTP, etc.) - I probably missed some JavaScripts that are available in `wp-admin`.

Here are modifications I've done:

Edited `class-two-factor-core.php`: following WordPress hooks (`edit_user_profile` and `edit_user_profile_update`) I have used corresponding functions in WooCommerce hooks:

```
add_action( 'woocommerce_edit_account_form', array( __CLASS__, 'user_two_factor_options' ) );
add_action( 'woocommerce_update_customer', array( __CLASS__, 'user_two_factor_options_update' ) );
```
Then, to make sure after showing two-factor settings page dies rendering, also added following to `providers/class.two-factor-fido-u2f-admin.php`:

1. in function `add_hooks` changed:
```
add_action( 'show_user_security_settings', array( __CLASS__, 'show_user_profile' ) );
```
To:
```
add_action( 'show_user_security_settings', array( __CLASS__, 'show_user_profile', 'woocommerce_edit_account_form' ) );
```

Though in frontend settings does not display e-mail address, nor QR code, so I believe there is something still missing...

I am also unsure if `enqueue_assets` function should be hooked into frontend - when I tried nothing changed anyway (didn't get AJAX working):
```
add_action( 'wp_enqueue_scripts',       array( __CLASS__, 'enqueue_assets' ) );
```

Anyone care to join into solving this?

luisjavierbautista:
Hi there,

There is any update about this?
I'm trying to include two-factor settings into UltimateMember profile pages, but does not display e-mail address, nor QR code

gytisrepecka:
[...]

PR #414, opened 2021-09-03
Workaround to avoid crashing bbPress' frontend profile page

CryoutCreations:
This workarounds skips the addition of the TwoFactor fields to the user profile when it's part of bbPress' frontend profile edit page. These are settings that shouldn't probably by editable on the forum profile anyway. 

Fixes the `Fatal error: Uncaught Error: Call to undefined function convert_to_screen()` reported in #175


class-two-factor-core.php
```diff
@@ -925,6 +925,11 @@ public static function manage_users_custom_column( $output, $column_name, $user_
 	 * @param WP_User $user WP_User object of the logged-in user.
 	 */
 	public static function user_two_factor_options( $user ) {
+		
+		/* workaround for two factor options crashing bbpress' profile page */
+		if ( function_exists('bbp_is_user_home_edit') && bbp_is_user_home_edit() ) return;
```
kasparsd, line 930: This is currently relying on a specific integration and historically this plugin doesn't have any conditional logic to account for specific integrations. Could we somehow convert this into something that can be placed in the `class-two-factor-compat.php` file which currently deals with Jetpack specific things?

https://github.com/WordPress/two-factor/blob/master/class-two-factor-compat.php

Which of the functionality in this method is actually not compatible with the front-end output?

r-a-y: You might want to check out my comment where I outline frontend integration with this problem (and a few more) here: https://github.com/WordPress/two-factor/issues/335#issuecomment-719087800

For this specific issue, it would probably be better to redeclare the admin functions on the frontend when needed.

CryoutCreations: > 
> 
> This is currently relying on a specific integration and historically this plugin doesn't have any conditional logic to account for specific integrations. Could we somehow convert this into something that can be placed in the `class-two-factor-compat.php` file which currently deals with Jetpack specific things?
> 
> https://github.com/WordPress/two-factor/blob/master/class-two-factor-compat.php
> 
> Which of the functionality in this method is actually not compatible with the front-end output?

I updated the PR to include an example error screenshot.
Until the required admin functions are exposed on the frontend (should they?), it might be possible to rewrite the workaround in the compat file to conditionally remove the two action hooks instead. I'll look into this.

CryoutCreations:
Per @kasparsd's request I moved the bbPress conditionality check to the `two-factor-compat` class and changed the function call to use a more general function. 

Hopefully the code now meets the logic separation requirements.

jeffpaul:
Per [yesterday's bug scrub](https://wordpress.slack.com/archives/C04L4KCUB/p1648040434425739), while we want to be compatible with official plugins we're going to punt this to a future release so that the 0.8.0 release can focus on the U2F deprecation.

jeffpaul:
@georgestephanis perhaps we close this PR and https://github.com/WordPress/two-factor/issues/175 and allow things to get resolved within bbp?

masteradhoc:
@georgestephanis @jeffpaul isnt the reason here that `WP_List_Table` is called? We dont call `convert_to_screen` directly but over `WP_List_Table` its called.

If i see `WP_List_Table` is called only on our fido implementation, [see here:](https://github.com/search?q=repo%3AWordPress%2Ftwo-factor%20WP_List_Table&type=code)

Given that this is removed soon (#439) i think its good to close this and focus on #335 to build a proper integration option.

Issue #412, opened 2021-08-12
Email notification turning & into HTML entity (&amp;)


nbwpuk:
We have come across an issue where the email subject line is including the `get_bloginfo( 'name' ) ` which for us includes an ampersand. However when the user receives the email notification is is converted into the HTML entity '`&amp;`' rather than the plain text '&'.

I have tested by replacing the `get_bloginfo( 'name' ) ` with a plain text string and the email is received correctly.

jeffpaul:
Sounds like an escaping issue, if you've got time and interest we'd gladly review a PR on this @nbwpuk?

kasparsd:
There doesn't appear to be any escaping happening during the email send here (note the stripping of HTML tags, though):

https://github.com/WordPress/two-factor/blob/f7bffdd5914775853324015e8a0fb32d9bcb57cc/providers/class-two-factor-email.php#L227-L248

I'm wondering if this is core `wp_mail()` doing this (or the underlying email library per https://github.com/PHPMailer/PHPMailer/issues/2115). Or maybe the value of the `blogname` option has it as `&amp;`? @nbwpuk Could you please check the raw option value on this page `/wp-admin/options.php`?

r-a-y:
@kasparsd - You need to use `wp_specialchars_decode( get_option( 'blogname' ), ENT_QUOTES )`. WP uses it internally when sending emails such as in `wp_notify_postauthor()`.

I can send a PR if you'll accept it.

kasparsd:
That's a great suggestion @r-a-y! A pull request would be awesome -- thank you!

Here are [the references to `wp_specialchars_decode()`](https://github.com/WordPress/WordPress/search?q=wp_specialchars_decode) in WP core.

nbwpuk:
Hey, sorry I went quiet after I raised this. Life took over.

@kasparsd Yes, I can confirm that options.php is showing `&` rather then `&amp;`

@r-a-y Thanks that has worked perfect and I have successfully tested on the site that was having issues.

I have submitted the above PR, but this is my first ever so do let me know if anything is out of place 👍

PR #418, opened 2021-09-07
Update class-two-factor-email.php to escape the HTML entity

nbwpuk:
Added wp_specialchars_decode( ) to escape the HTML entity on the Email Subject line #412


r-a-y:
@nbwpuk - PR looks good, but you should submit the PR against your `nbwpuk-email-patch` branch and not your `master` branch.

kasparsd:
@r-a-y I think this looks good as is. The origin branch doesn't have a practical difference, right?

kasparsd (approved):
Looks good!

r-a-y:
Sure, but it's a Git history thing. Keeps things tidy. If you don't care about that, then feel free to merge.

PR #419, opened 2021-09-11
Add filter two_factor_email_token_length

jcdjacobs:
The length of the email token (code) can be set with filter two_factor_email_token_length
The default is 8, like before


jeffpaul:
@kasparsd any code review feedback on the approach here?

kasparsd:
The formatting fixes are addressed in #653 which also adds similar filter to backup codes.

jcdjacobs:
Thank you very much for implementing this feature in #653 !

PR #420, opened 2021-09-24
Improve TOTP autocomplete behaviour

squaredpx:
Hi! This is a little improvement over #373 that included the autocomplete attribute. I've set it attribute to 'one-time-code' so browsers and password managers can autocomplete the code if necessary/possible. I've tested it on the latest versions of 1Password and Safari, Chrome, Firefox & Edge and it autocompletes the TOTP code (previously it only worked on Safari)

Great job with the plugin by the way, thanks!

### DIFF
--- a/providers/class-two-factor-totp.php
+++ b/providers/class-two-factor-totp.php
@@ -459,7 +459,7 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 		</p>
 		<p>
 			<label for="authcode"><?php esc_html_e( 'Authentication Code:', 'two-factor' ); ?></label>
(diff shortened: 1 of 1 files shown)


jeffpaul:
Per [yesterday's bug scrub](https://wordpress.slack.com/archives/C04L4KCUB/p1648040434425739), we're punting this to a future release so we can focus on the U2F deprecation in 0.8.0.

kasparsd:
This is great UX improvement and a minimal change so we can merge this in and adjust if necessary based on user feedback.

kasparsd (approved):

PR #421, opened 2021-10-15
Added a filter to override the email headers

talhaimam:
Hello,
The email message was always sent in a text/plain format by default. The email messages containing the security code were absolutely minimal. I've added a filter to allow editing the email headers, so that the email content can also be rendered in other formats such as text/html. Furthermore, the to, cc and bcc addresses can be also be altered by adding these values to the header array using this filter.

I've tested this filter across different settings and environments, working nicely. Please review!

The plugin also had other filters which were not listed in the readme file, I've gone ahead and listed them there.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -2,7 +2,7 @@
 Contributors: georgestephanis, valendesigns, stevenkword, extendwings, sgrant, aaroncampbell, johnbillion, stevegrunwell, netweb, kasparsd, alihusnainarshad, passoniate
 Tags: two factor, two step, authentication, login, totp, fido u2f, u2f, email, backup codes, 2fa, yubikey
 Requires at least: 4.3
-Tested up to: 5.9
+Tested up to: 6.0
 Requires PHP: 5.6
 Stable tag: trunk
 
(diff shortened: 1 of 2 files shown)


jeffpaul:
Per [yesterday's bug scrub](https://wordpress.slack.com/archives/C04L4KCUB/p1648040434425739), @georgestephanis would rather see a change to the email code to also send multipart text/html (which WP Core likely does) but is otherwise fine with this change in theory.

talhaimam:
@jeffpaul Makes more sense, let me look into it.

talhaimam:
@jeffpaul Getting back to this, the 2fa codes are sent using the wp_mail function, which sends the email in text/plain format by default. The wp_mail function does have in-built filters to change the email headers. So yes the core does have the capability to send multipart text/html, but it needs to be specified. Do you suggest altering the wp_mail function to send email in multipart text/html format by default?

jeffpaul:
@talhaimam I think having this plugin default to multipart text/html by default makes sense, yes

talhaimam:
@jeffpaul one more thing, multipart/mixed or multipart/alternative mime-types require MIME encoding in the email content. If we set the default type to multipart, then the plugin's default message would need to be changed as well. Plugin Documentation would also need to mention that anyone editing the default email content, would need to make sure that the mime encoding is included. Setting it to send text/html though, wouldnt require any of this.

PR #261, opened 2018-11-06
[WIP] Two Factor User Profile Shortcode

kasparsd:
Fixes #247.

## Approach

- Abstract the profile settings away from WP core user profile.

- Introduce a shortcode for rendering the two-factor user settings.

## Known Issues

- `WP_List_Table` used for FIDO U2F settings doesn't work outside WP admin. 

        Fatal error: Uncaught Error: Call to undefined function convert_to_screen() in /srv/www/wordpress-default/wp-admin/includes/class-wp-list-table.php:132 Stack trace: #0 /srv/www/projects/plugins/two-factor/providers/class.two-factor-fido-u2f-admin.php(182): WP_List_Table->__construct() #1 /srv/www/wordpress-default/wp-includes/class-wp-hook.php(286): Two_Factor_FIDO_U2F_Admin::show_user_profile(Object(WP_User)) #2 /srv/www/wordpress-default/wp-includes/class-wp-hook.php(310): WP_Hook->apply_filters('', Array) #3 /srv/www/wordpress-default/wp-includes/plugin.php(453): WP_Hook->do_action(Array) #4 /srv/www/projects/plugins/two-factor/class.two-factor-core.php(748): do_action('show_user_secur...', Object(WP_User)) #5 /srv/www/projects/plugins/two-factor/class.two-factor-core.php(235): Two_Factor_Core::user_two_factor_options(Object(WP_User)) #6 /srv/www/wordpress-default/wp-includes/shortcodes.php(319): Two_Factor_Core::user_profile_shortcode('', '', 'two-factor-user...') #7 [internal function]: do_shortcode_tag(Array) #8 /sr in /srv/www/wordpress-default/wp-admin/includes/class-wp-list-table.php on line 132

PR #424, opened 2021-12-23
Feature/ Frontend TOTP activation shortcode

JurriaanK:
This implements a [two-factor-set-totp] shortcode which can be used for front-end TOTP 2fa activation.

I looked into making all provider settings available frontend, but that would require extensive modifications to the whole plugin. So I took a rather straightforward approach. I created specific functions for generating the shortcode output and saving the input because outputting to frontend and saving options from a frontend request demands a different approach than backend, but I tried to reuse as much of the backend functions as possible.  I think this brings a much wanted feature in a relatively clean approach. 

Notes:
- Currently only first time activation is possible. Reset of the key has to be done via the user profile. Allowing users to reset via frontend requires a global setting on admin level in my opinion, but two-factor currently doesn't have a global settings page yet.
- The shortcode doesn't use wp-admin urls, which is handy for sites where wp-admin is blocked for non-admin users.
- After using the frontend settings the TOTP provider is enabled and set as primary.

### DIFF
--- a/providers/class-two-factor-totp.php
+++ b/providers/class-two-factor-totp.php
@@ -53,7 +53,72 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 		add_action( 'edit_user_profile_update', array( $this, 'user_two_factor_options_update' ) );
 		add_action( 'two_factor_user_settings_action', array( $this, 'user_settings_action' ), 10, 2 );
 
+		add_shortcode( 'two-factor-set-topt', array( $this, 'topt_shortcode' ) );
+
 		return parent::__construct();
+	}
+
+	/**
+	 * Output for the two-factor-set-topt shortcode
+	 *
+	 * @param array $atts Array of attributes.
+	 *
+	 * @return string
+	 */
+	public function topt_shortcode( $atts ) {
+		$user = wp_get_current_user();
+
+		if ( ! $user ) {
+			return esc_html__( 'Login to set up your Time-based One-Time Password.', 'two-factor' );
+		}
+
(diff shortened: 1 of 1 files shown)


jeffpaul:
Per [yesterday's bug scrub](https://wordpress.slack.com/archives/C04L4KCUB/p1648040434425739), general sentiment was in favor of this but going to punt to future release so we can focus on the U2F deprecation in 0.8.0.

iandunn:
Related #261

jeffpaul:
@kasparsd as we did with #261, sadly, perhaps his might be best to close and look to craft a block for this instead?

JurriaanK:
@jeffpaul @kasparsd To some extend I understand the reasoning, but why not also support a shortcode. Many themes don't use blocks yet.
And how many years will it take for this plugin to have a proper front end solution for activating 2fa? We are still using this addition, waiting for it to be merged or till a proper alternative is available.
Is it possible to provide the requirements this feature should meet so it will be merged?

jeffpaul:
My preference would be to not have a shortcode within the plugin as that's not something that's a "first party" experience in the editor these days.  Yes there's a shortcode block but that's more a route to ensure there's no breaking change for folks with shortcodes already in use.  Introducing a new shortcode feels against the preference towards blocks for more modern WP integrations.  I get the argument that many folks are still using "classic" themes, but again I angle more towards nudging folks towards more modern WP experiences.  I'm not against shortcodes per se, but more in favor of supporting existing ones versus introducing new ones.

Also worth noting that I'm not necessarily a primary decision maker for the 2FA plugin, so happy to be overruled by others.

Issue #425, opened 2022-01-05
Use hash_equals for email


Rahe:
Hello,

On https://github.com/WordPress/two-factor/blob/cf55e92bef8f44ffeed2d4f144dc177a4cb742ca/providers/class-two-factor-email.php#L190 the hashes are compared with `!==`.
It seems better to compare them with the function `hash_equals` instead.

Nicolas,

kasparsd:
Fixed in #426

Timmieke:
Could it be, that something went wrong in this part. Plugin functions normal, but there are no emails sent anymore.

kasparsd:
@Timmieke I've verified on my personal blog that the 2FA over email works as expected -- I'm able to receive the codes and login using them.

This particular change is for the code comparison logic so it shouldn't impact the delivery of emails or the generation of the tokens.

Timmieke:
Ok, sorry... i will retest it, thanks!

PR #426, opened 2022-01-11
Implemented Hash Equal Function instead of !== based on issue #425

Mati02K:
Implemented the safe string hash function comparison between two hashes based on issue #425

### DIFF
--- a/providers/class-two-factor-email.php
+++ b/providers/class-two-factor-email.php
@@ -187,7 +187,7 @@ class Two_Factor_Email extends Two_Factor_Provider {
 		$hashed_token = $this->get_user_token( $user_id );
 
 		// Bail if token is empty or it doesn't match.
-		if ( empty( $hashed_token ) || ( wp_hash( $token ) !== $hashed_token ) ) {
+		if ( empty( $hashed_token ) || ( !hash_equals( wp_hash( $token ), $hashed_token) ) ) {
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


providers/class-two-factor-email.php
```diff
@@ -187,7 +187,7 @@ public function validate_token( $user_id, $token ) {
 		$hashed_token = $this->get_user_token( $user_id );
 
 		// Bail if token is empty or it doesn't match.
-		if ( empty( $hashed_token ) || ( wp_hash( $token ) !== $hashed_token ) ) {
+		if ( empty( $hashed_token ) || ( !hash_equals( wp_hash( $token ), $hashed_token) ) ) {
```
kasparsd, line 190: Do we still need the wrapping parentheses now that is a function call?

```suggestion
		if ( empty( $hashed_token ) || ! hash_equals( wp_hash( $token ), $hashed_token ) ) {
```

kasparsd (commented):
Thanks for the pull request! This looks good! Had one suggestion about the formatting.

providers/class-two-factor-email.php
Mati02K: No need Of Wrapping parentheses, I have added them just to make the function and empty function look like separate ones. Will remove it to avoid confusion.

kasparsd (approved):
This is great! Thank so much!

Issue #232, opened 2018-05-10
Support W3C Web Authentication


jcjones:
WebAuthn is [supported in Firefox 60, and supported shortly in Chrome 67 and Edge 18](https://caniuse.com/#feat=webauthn), providing a cross-browser support that is the spiritual successor to U2F. 

It's possible to [migrate U2F users and their credentials to WebAuthn](https://www.w3.org/TR/webauthn/#sctn-appid-extension), but also possible to run U2F and WebAuthn side-by-side as separate modules.

MDN: https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API
Spec: https://www.w3.org/TR/webauthn/

pwaring:
@PHPGangsta That's not the case. You can use Google Authenticator with any browser, and you can use U2F with Firefox (I do, every single day) - although at the moment it's not enabled by default.

PHPGangsta:
@pwaring So you can use the "Two Factor" Wordpress plugin with a plain Firefox and webauthn/U2F?

I'm using "WordPress 5.0.3" and "Two Factor 0.3.0" and Firefox 64.

I get this in my Wordpress if I want to activate U2F for a Wordpress user:
![image](https://user-images.githubusercontent.com/608408/51175938-f206e580-18bb-11e9-948f-fd35db37eeb4.png)

Which is clear: This Plugin only supports Chrome as a U2F browser, not all (via webauthn).

I'm asking that webauthn is added to the "Two Factor" Wordpress plugin, so you can use U2F in Firefox without any plugins.

Maybe you have installed the "U2F" Firefox-Addon which adds the old non-standard U2F functionality that Chrome supported for a few years?

pwaring:
@PHPGangsta What you said was: "Currently only Chrome is supported by your plugin". That's not the case - you can use this plugin with Firefox and other browsers, either by using TOTP (through Google Authenticator or a similar app) or by enabling U2F in Firefox.

You can enable U2F in Firefox by going to `about:config` and switching `security.webauth.u2f` to `true`. This definitely works in 64.0, not sure when it was added.

PHPGangsta:
[...]

PR #427, opened 2022-01-21
Support WebAuthn

### LINKED ISSUE #423: U2F support in the future versions of Chrome
Hi,

Today I have seen in my JS console the following warning while I was logging in using my Yubikey:

> The U2F Security Key API is deprecated and will be removed soon. If you own this website, please migrate to the Web Authentication API. For more information see https://groups.google.com/a/chromium.org/g/blink-dev/c/xHC3AtU_65A/m/yg20tsVFBAAJ

It means that in the next year security keys won't work at all with your plugin?

mcguffin:
As announced in #423 here's my PR.

**Tested with:**
 - Windows 10 (VirtualBox)
 - Firefox macOS 10.15 using [SoftU2F](https://github.com/github/SoftU2F)
 - Chrome macOS using Chrome's native OS support
 - Firefox Android 10
 - Safari on iOS 14.4

**Todo (any help on these is very welcome):**
 - Test more browsers and more devices
 - Accessibility
 - Make UI less mcguffin and more WordPress


includes/WebAuthn/class-webauthn-keystore.php
```diff
@@ -0,0 +1,128 @@
...
+	 * @return bool
+	 */
+	public function delete_key( $user_id, $keyLike ) {
+		global $wpdb;
+		return $wpdb->query( $wpdb->prepare(
+			"DELETE FROM $wpdb->usermeta WHERE user_id=%d AND meta_key=%s AND meta_value LIKE %s",
+			$user_id,
+			self::PUBKEY_USERMETA_KEY,
+			'%' . $keyLike . '%'
+		) ) !== 0;
```
sjinks, line 124: Ideally, you need to fetch the record first and then delete it by ID. You need the "fetch first" thing to update the meta cache, the same way that [`delete_metadata()`](https://github.com/WordPress/WordPress/blob/5.8.2/wp-includes/meta.php#L447) does. Otherwise, [get_keys()](https://github.com/WordPress/two-factor/pull/427/files#diff-55d4450abd4b0b020d83b93cf01b80e2da009d0f00f08ccb872a3d522a59d9a2R55) may yield unexpected result if the site uses an object cache.

providers/class-two-factor-webauthn.php
```diff
@@ -0,0 +1,642 @@
...
+
+	/**
+	 * Return the U2F AppId. WebAuthn requires the AppID
+	 * to be the current domain or a suffix of it.
+	 *
+	 * @return string AppID FQDN
+	 */
+	public function get_app_id() {
+
+		$fqdn = wp_parse_url( network_site_url(), PHP_URL_HOST );
```
sjinks, line 120: It is probably not critical, but WebAuthn requires an origin here, not just the domain. That is if the webserver for some reason uses a non-standard port for HTTPS (say, 8443), it needs to be included in the AppID.

The best way is probably to copy the logic from [`get_u2f_app_id()`](https://github.com/WordPress/two-factor/blob/e72192eef199285e8edc954e049ee888789179d8/providers/class-two-factor-fido-u2f.php#L103) (just omit the protocol).

providers/class-two-factor-webauthn.php
```diff
@@ -0,0 +1,642 @@
...
+				wp_send_json_error( new WP_Error( 'webauthn', $this->webauthn->getLastError() ) );
+			}
+			/* translators: %s webauthn app id (domain) */
+			$key->label     = sprintf( esc_html__( 'New Device - %s', 'two-factor' ), $this->get_app_id() );
+			$key->md5id     = md5( implode( '', array_map( 'chr', $key->id ) ) );
+			$key->created   = time();
+			$key->last_used = false;
+			$key->tested    = false;
+
+			if ( false !== $this->key_store->find_key( $user_id, $key->md5id ) ) {
```
sjinks, line 399: This is probably not critical, but…

According to the [Registration Ceremony](https://www.w3.org/TR/webauthn-2/#sctn-registering-a-new-credential), Item 22, the credential must be unique site-wide:

> Check that the credentialId is not yet registered to any other user. If registration is requested for a credential that is already registered to a different user, the Relying Party SHOULD fail this registration ceremony, or it MAY decide to accept the registration, e.g. while deleting the older registration.

mcguffin:
@sjinks Thanks for review!

jeffpaul:
@sjinks mind re-reviewing after the latest changes from @mcguffin?

sjinks (approved):
LGTM, but I haven't tried to run this locally.

includes/WebAuthn/class-cbor-decoder.php
```diff
@@ -0,0 +1,300 @@
+<?php
+
+if ( ! defined('ABSPATH') ) {
+	die('Bye!');
+}
```
sjinks, line 5: A minor nitpick: this file contains only class definitions; therefore, even if this file is accessed directly, there will be no errors/warnings. That's why it is safe to remove these lines.

includes/WebAuthn/class-webauthn-keystore.php
```diff
@@ -0,0 +1,149 @@
+<?php
+
+if ( ! defined('ABSPATH') ) {
+	die('Bye!');
+}
```
sjinks, line 5: Same as above

providers/js/webauthn-admin.js
```diff
@@ -0,0 +1,370 @@
+( function( $ ) {
+
+	/**
+	 *	Borrowed from https://github.com/davidearl/webauthn
+	 */
+	function webauthnRegister( key, callback ) {
+
+		let publicKey = Object.assign( {}, key.publicKey );
```
sjinks, line 8:
```suggestion
		const publicKey = Object.assign( {}, key.publicKey );
```

providers/js/webauthn-admin.js
```diff
@@ -0,0 +1,370 @@
...
+	 *	Borrowed from https://github.com/davidearl/webauthn
+	 */
+	function webauthnAuthenticate( pubKeyAuth, callback ) {
+
+		const originalChallenge = pubKeyAuth.challenge;
+		const pk = Object.assign( {}, pubKeyAuth );
+
+		pk.challenge = new Uint8Array( pubKeyAuth.challenge );
+		pk.allowCredentials = pk.allowCredentials.map( k => {
+			let ret = Object.assign( {}, k );
```
sjinks, line 68:
```suggestion
			const ret = Object.assign( {}, k );
```

providers/js/webauthn-admin.js
```diff
@@ -0,0 +1,370 @@
...
+		document.execCommand( 'selectAll', false, null );
+	};
+
+	$( document ).on( 'click', '#webauthn-register-key', e => {
+
+		e.preventDefault();
+
+		$( e.target ).next( '.webauthn-error' ).remove();
+
+		let $btn = $( e.target ).addClass( 'busy' );
```
sjinks, line 268:
```suggestion
		const $btn = $( e.target ).addClass( 'busy' );
```

providers/js/webauthn-admin.js
```diff
@@ -0,0 +1,370 @@
...
+
+		const opts = JSON.parse( $( e.target ).attr( 'data-create-options' ) );
+
+		register( opts, response => {
+			$btn.removeClass( 'busy' );
+			if ( response.success ) {
+				let $keyItem = $( response.html ).appendTo( '#webauthn-keys' );
+				let $keyLabel = $keyItem.find( '.webauthn-label' );
```
sjinks, line 276:
```suggestion
				const $keyItem = $( response.html ).appendTo( '#webauthn-keys' );
				const $keyLabel = $keyItem.find( '.webauthn-label' );
```

providers/js/webauthn-login.js
```diff
@@ -0,0 +1,114 @@
...
+	 *	Borrowed from https://github.com/davidearl/webauthn
+	 */
+	function webauthnAuthenticate( pubKeyAuth, callback ) {
+
+		const originalChallenge = pubKeyAuth.challenge;
+		const pk = Object.assign( {}, pubKeyAuth );
+
+		pk.challenge = new Uint8Array( pubKeyAuth.challenge );
+		pk.allowCredentials = pk.allowCredentials.map( k => {
+			let ret = Object.assign( {}, k );
```
sjinks, line 12:
```suggestion
			const ret = Object.assign( {}, k );
```

jeffpaul:
@sjinks thanks, MUCH appreciated!

@mcguffin any thoughts on the feedback above / ability to work on updates accordingly?

mcguffin:
@sjinks And a capital THANKS from me too!

iandunn:
I tested this (as of 4b5513f) w/ Firefox Dev Edition `106` and Chrome `105`, with a Yubikey 5C NFC and a Yubikey 5C Nano. It worked in all cases 👍🏻

jeffpaul:
@kasparsd @georgestephanis any further changes / testing you'd like here before merging in for the 0.8.0 release?

georgestephanis:
@jeffpaul If you're happy with it, I trust your judgement.

iandunn:
I apologize for raising this late in the process, but I've been looking at this closer and have some concerns:

* There aren't any tests (here or upstream) for the [2tvenom/cborencode](https://github.com/2tvenom/CBOREncode) and [davidearl/webauthn](https://github.com/davidearl/webauthn) libraries that were forked into this PR. They appear to be mostly inactively, too. The [madwizard-org/webauthn-server](https://github.com/madwizard-org/webauthn-server/) library that's used in [sjinks/wp-two-factor-provider-webauthn](https://github.com/sjinks/wp-two-factor-provider-webauthn) seems like a safer choice.
* There aren't any tests for first-party code.
* Forking the libraries could mean we miss out on Composer/GitHub notifications of any future security issues.

Those issues make me a bit nervous to put this into production. I can see a couple potential solutions, but first I'm curious to hear what everyone else thinks. I‘m also happy to help with any work that may be needed.

jeffpaul:
Better to raise late than not and run into issues later, so thanks for calling this out @iandunn.  I share the concerns you raise in reading through, would be curious what alternatives you have in mind to weigh against the current PR implementation to try and keep the plugin on a path towards adding WebAuthn support.

iandunn:
I agree that getting `0.8` out the door soon is important 👍🏻 

If we want to [migrate U2F keys](https://github.com/WordPress/two-factor/pull/439#issuecomment-1279366113), then the bare minimum we'll need to do is switch to `madwizard-org/webauthn-server`. That alone would resolve most of my concerns. Adding first-party tests could be a stretch goal, or be done after the release (maybe as part of #468).

We could probably borrow a lot of code from `sjinks/wp-two-factor-provider-webauthn` to make the library switch go quicker, but I'd want to keep the leaner nature of this PR.

That's still a decent chunk of work, but it seems doable. I'm happy to take that on if folks like that approach (_but I'm also happy to just help w/ testing/review if someone else wants to do it; I'm not trying to take over_).

jeffpaul:
@kasparsd @georgestephanis any opinions (strong or weak) on Ian's proposal above for a shift in approach on this PR?

mcguffin:
@iandunn I made some minor improvements and added a basic test class.

Concerning U2F-Key I hope I can relieve some of from your concerns.  
I played around with key migration today and finaly got it to work:
https://github.com/mcguffin/two-factor/blob/migrate-u2f/includes/WebAuthn/class-webauthn-key-migrator.php
It's basically a rephrased version `sjinks/wp-two-factor-provider-webauthn` + `madwizard-org/webauthn-server`.

Don't know whether to move this to another PR or merge into this one...?

iandunn:
Migrating the keys is awesome 🎉 A separate PR seems better to me, to help keep things manageable 👍🏻 

What are your thoughts on removing the `2tvenom/cborencode` and `davidearl/webauthn` libraries, and using `madwizard-org/webauthn-server` for everything instead? And about including the library via Composer instead of forking it?

mcguffin:
@iandunn I've chosen `2venom/cborencode` and `davidearl/webauthn` mainly because they were easy to rephrase. I didn't want to poke into the two-factor codebase and infrastructure too deeply and keep it the way I found it: lean and self-contained.

I'm having similar concerns about `webauth`and `cborencode` like you and replacing them with a sustainable, well aintained and lean 3rd party solution would be the right thing to do. I just have my doubts if (1) now is the right time, (2) `madwizard-org/webauthn-server` is the right library and (3) the change in architecture and the follow-up-work is worth it.

The README of `madwizard-org/webauthn-server` declares the current plugin state to be:
  > Pretty stable but the API may still change slightly until the 1.0 release.  

It is bringing 11 direct dependencies resulting in 24 packages being installed. The `vendor/` dir alone weights 3.4M. https://github.com/lbuchs/WebAuthn has zero dependecies and a codebase of 132k. (Although not a recommendation — no unit tests either, < 1 commit/month and an open php file included in the distribution.)  

`madwizard-org` looks like a one-person-show. There's a certain risk of the package being abandoned. (Not pointing at anyone but myself—there are tons of abandoned repos in my own github.)

WordPress has no central `vendor/` dir; every plugin brings its own instead. To avoid conflicting package versions, we would need to prefix the namespace of all depending classes.
     - `sjinks/wp-two-factor-provider-webauthn` is using [`typisttech/imposter-plugin`](https://github.com/typisttech/imposter-plugin) (probably abandoned).  
     - [WordPress SEO](https://github.com/Yoast/wordpress-seo) has added [PHP Scoper](https://github.com/humbug/php-scoper) to their build process.
From a system resources point of view (CPU, RAM, disk-operations, traffic, ...) it's a not a good idea to load the exact same code multiple times into memory. On high-traffic sites system resources are already a big problem. With climate desaster and energy crisis in mind, the whole resource issue may gain even more significance in the very near future.

With the upcoming [Plugin Dependencies](https://core.trac.wordpress.org/ticket/22316) feature (kudos @afragen!), there could be a WordPress-specific way to handle 3rd party libs—saving all of us from clunky composer constructs.

afragen:
@mcguffin while I appreciate the props I don’t Plugin Dependencies only handles plugins as dependencies. 

It cannot handle libraries or frameworks as dependencies unless these are packaged as plugins. Then it might be possible to add them similarly to adding a non-dot org plugin as a dependency.

mcguffin:
@afragen 
> [...] unless these are packaged as plugins.
This is exactly what's fueling my optimism!  

Maybe I'm wrong, but I would expect some kind of dependency-only plugins to appear wp.org, providing certain commonly used PHP libraries like guzzle, phpseclib, crypto, …

In an ideal world as a plugin author, I would simply add all the PHP-Packages I need to the "Requires Plugins" header, instead of `composer require`ing them and thus bloating my plugin with a `vendor/` directory.

afragen:
I don't believe the plugins team any longer allows for frameworks, etc. 

You could, however, host them on GitHub and hook into the Plugin Dependencies just like I do with the example plugin for non dot org dependencies. 

Happy to help get you started with this.

mcguffin:
@afragen thanks for the clearification. 
I guess you're refering to `afragen/wp-dependency-installer`, right?

afragen:
Actually referring to https://github.com/afragen/add-plugin-dependency-api

It a composer requirement in the Plugin Dependencies feature plugin as an example.

iandunn:
@mcguffin you've raised some good points 🤔 

I also wish that there was a more mature & popular library, but unfortunately it doesn't look like that exists yet. It seems like `madwizard/webauthn-server` is the most mature, though, despite the imperfections.

I also wish it was leaner, but I don't think it'll be a big issue in practice, since the code is autoloaded and only used in a few flows. That's also why I think it's ok to to just include it without any prefixing. Not ideal, but it shouldn't cause any tangible problems in practice.

For me, it ultimately comes down not having enough confidence that the other libraries are secure enough to run in production on my sites, which is a dealbreaker. I understand that's subjective, though, and I can definitely understand your perspective.

mcguffin:
@iandunn What do you think of [web-auth/webauthn-framework](https://github.com/web-auth/webauthn-framework)? Or at least parts of it like `cose-lib` and `webauthn-lib`? There is also a class for U2F key conversion.
The downside: We would likely have to use v3.3.x instead of 4.x because of the PHP requirement.

iandunn:
That one does look mature and well tested too 🤔  IIRC @sjinks tried it out for his plugin and found that [the easy way](https://webauthn-doc.spomky-labs.com/v/v3.3/the-webauthn-server/the-easy-way) didn't support U2F migration, and [the hard way](https://webauthn-doc.spomky-labs.com/v/v3.3/the-webauthn-server/the-hard-way) was pretty difficult to work with.

It looks like the 4.0 docs don't have that distinction, though, so maybe they made it easier to use? If that's the case, then I think it'd definitely be worth considering. IMO it'd be fine to require PHP 7.x for this particular provider (while still supporting 5.6 in the other providers), per the discussion in #461.

@sjinks, do you have any thoughts on `web-auth/webauthn-framework`?

sjinks:
I didn't play with it much, tbh. As far as I remember, it had a lot of dependencies (circa 20 MB, if I am not mistaken). IIRC, that was more than wordpress.org allowed (10 MB).

iandunn:
Ah, good point, it looks like even just the `.php` files is still `17MB` 🙈 (`
composer install --no-dev && find . -name "*.php" | xargs du -sch`). `madwizard/webauthn-server` is `4.4MB` in comparison.

I was also assuming that the `4.x.x` branches required PHP `7.x`, but it looks like it's actually `8.1`, which would probably be  too strict even if it was only for this provider. So we'd probably would need to use the `3.3.x` branch like Jörn said. But then, we probably can't assume that they'll do security patches for that branch, so that doesn't seem viable to me.

paulschreiber:
What's the ETA on getting this merged?

iandunn:
>  IIRC, that was more than wordpress.org allowed (10 MB).

I looked for the details on that, and didn't see anything in [the handbook](https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/) or on [make/Plugins](https://make.wordpress.org/plugins/). Jetpack is `222MB` unzipped and WordFence is `63MB`, but Two Factor is only `1.7MB`, so I think we're probably safe to add whatever library we need. We should still prune any files that won't be used, though.

I'm still leaning towards `madwizard-org/webauthn-server` being the best one that supports PHP `7.x`.

providers/css/webauthn-admin.css
```diff
@@ -0,0 +1,166 @@
...
+}
+.webauthn-key {
+	display: -webkit-box;
+	display: -ms-flexbox;
+	display: flex;
+	-ms-flex-wrap: wrap;
+	flex-wrap: wrap;
+	-webkit-box-align: last baseline;
+	-ms-flex-align: last baseline;
+	align-items: last baseline;
```
StevenDufresne, line 49: We can probably center align this row.

```suggestion
	align-items: center;
```
| State | Before | After | 
| --- | --- |--- |
| | Screenshot 2023-04-21 at 11 50 25 AM | Screenshot 2023-04-21 at 11 50 16 AM |
 | With Notice | Screenshot 2023-04-21 at 11 50 35 AM | Screenshot 2023-04-21 at 11 50 51 AM |

providers/css/webauthn-admin.css
```diff
@@ -0,0 +1,166 @@
...
+	color: #0085ba;
+}
+.webauthn-key [data-tested="fail"] {
+	color: #dc3232;
+}
+.webauthn-key [data-tested="fail"]::before {
+	content: "";
+}
+.webauthn-key [data-tested="untested"] {
+	color: #ccc;
```
StevenDufresne, line 113:
```suggestion
	color: #ccc;
	vertical-align: bottom;
```
Align the icon better with text.

| Before | After |
| --- | --- |
| Screenshot 2023-04-21 at 11 55 36 AM | Screenshot 2023-04-21 at 11 55 12 AM |

providers/js/webauthn-login.js
```diff
@@ -0,0 +1,114 @@
...
+					callback( false, err.name + ': ' + err.message );
+				} else {
+					callback( false, err.toString() );
+				}
+			});
+	};
+
+	const login = ( opts, callback ) => {
+
+		const { action, payload, _wpnonce } = opts;
```
StevenDufresne, line 70: `action` & `_wpnonce` are deconstructured but I don't any references.

providers/js/webauthn-admin.js
```diff
@@ -0,0 +1,370 @@
...
+		});
+	};
+
+	const sendRequest = ( opts, callback ) => {
+
+		$.ajax( {
+			url: wp.ajax.settings.url,
+			method: 'post',
+			data: opts,
+			success:callback
```
StevenDufresne, line 193:
```suggestion
			success: callback
```

providers/css/webauthn-admin.css
lkraav: Needing `vertical-align` is usually a signal to refactor torwards flexbox. We've come a long way, but browsers still render `vertical-align` differently in various cases. Life is too short for that.

ravinderk (changes requested):
@mcguffin I left a few suggestions on code. Feel free to ask questions.

I noticed that new files need to be formatted correctly. You cannot catch formatting issues because `includes` directory is excluded. I think you can remove [this line](	<exclude-pattern>*/includes/*</exclude-pattern>) to get the correct formatting results locally.

includes/WebAuthn/class-webauthn-keystore.php
```diff
@@ -0,0 +1,141 @@
...
+	 */
+	public function key_exists( $keyLike ) {
+
+		global $wpdb;
+
+		$num_keys = $wpdb->get_var( $wpdb->prepare(
+			"SELECT COUNT(*) FROM $wpdb->usermeta WHERE meta_key=%s AND meta_value LIKE %s",
+			self::PUBKEY_USERMETA_KEY,
+			'%' . $wpdb->esc_like( serialize( $keyLike ) ) . '%'
+		) );
```
xknown, line 90: Would it make sense to only use core functions like `get_user_meta( $user_id, self::PUBKEY_USERMETA_KEY )` and then doing the lookup on PHP? This would avoid uncached queries for sites that make use of object caching plugins.

jornfranke:
@kasparsd  could you review it or assign someone else for review to unblock this?

georgestephanis:
I think the main blocker on the current PR is getting context or resolution on the questions raised in a few of the above reviews.

providers/class-two-factor-webauthn.php
```diff
@@ -0,0 +1,646 @@
...
+
+	/**
+	 * Return the U2F AppId. WebAuthn requires the AppID
+	 * to be the current domain or a suffix of it.
+	 *
+	 * @return string AppID FQDN
+	 */
+	public function get_app_id() {
+
+		$url_parts = wp_parse_url( network_site_url() );
```
joshbetz, line 120: What is the ideal workflow on a network that has sites with different domains?

joshbetz:
It's not working in Safari for me 🤔 I'm able to register my browser with Touch ID, but logging in with Touch ID is not working for me.

adampl:
This is ridiculous. Is this ever going to be finished?

jeffpaul:
@georgestephanis @kasparsd (and anyone else following this with actual interest, availability, and capability in getting this to `done`) would it work for you if I scheduled a Google Meet video chat to review what's needed here (and in any other larger, functional PRs) so we can get that more clearly defined, documented, and identify folks who can help resolve those?

rowan-m:
I'd be happy to help out here with an eye on closing out #232 and moving on to #779.

> @georgestephanis @kasparsd (and anyone else following this with actual interest, availability, and capability in getting this to `done`) would it work for you if I scheduled a Google Meet video chat to review what's needed here (and in any other larger, functional PRs) so we can get that more clearly defined, documented, and identify folks who can help resolve those?

masteradhoc:
Hey @rowan-m 

Our next byweekly meeting is on wednesday (next week) 4PM UTC in #core-passwords slack channel. Feel free to join. We'd be happy to discuss!

Issue #364, opened 2020-06-16
Fix Coveralls reporting


kasparsd:
Currently the following error is generated when trying to upload the coverage report to Coveralls during the Travis run:

https://travis-ci.com/github/WordPress/two-factor/jobs/349903189

```
$ source "$DEV_LIB_PATH/travis.after_script.sh"
if can_generate_coverage_clover && [ -s "$TEMP_DIRECTORY/paths-scope-php" ] && [ -n "$INITIAL_DIR" ]; then
	cd "$INITIAL_DIR"
	# Coveralls 1.x.x path
	COVERALLS_BIN="vendor/bin/coveralls";
	# Coveralls 2.x.x path
	if [ ! -e "$COVERALLS_BIN" ]; then
		COVERALLS_BIN="vendor/bin/php-coveralls"
	fi
	php $COVERALLS_BIN -v
	cd - > /dev/null
fi
In Configurator.php line 165:
                                                                                 
  [Symfony\Component\Config\Definition\Exception\InvalidConfigurationException]  
  coverage_clover XML file is not readable                                       
                                                                                 
Exception trace:
  at /tmp/wordpress/src/wp-content/plugins/two-factor/vendor/php-coveralls/php-coveralls/src/Satooshi/Bundle/CoverallsV1Bundle/Config/Configurator.php:165
 Satooshi\Bundle\CoverallsV1Bundle\Config\Configurator->getGlobPaths() at /tmp/wordpress/src/wp-content/plugins/two-factor/vendor/php-coveralls/php-coveralls/src/Satooshi/Bundle/CoverallsV1Bundle/Config/Configurator.php:191
 Satooshi\Bundle\CoverallsV1Bundle\Config\Configurator->getGlobPathsFromStringOption() at /tmp/wordpress/src/wp-content/plugins/two-factor/vendor/php-coveralls/php-coveralls/src/Satooshi/Bundle/CoverallsV1Bundle/Config/Configurator.php:141
 Satooshi\Bundle\CoverallsV1Bundle\Config\Configurator->ensureCloverXmlPaths() at /tmp/wordpress/src/wp-content/plugins/two-factor/vendor/php-coveralls/php-coveralls/src/Satooshi/Bundle/CoverallsV1Bundle/Config/Configurator.php:119
[...]

PR #370, opened 2020-06-26
Use local tooling for reporting the phpunit coverage

kasparsd:
Fixes #364.

PR #436, opened 2022-03-24
Move to GitHub actions

### LINKED ISSUE #435: Switch to GitHub Actions for CI runs
The Travis CI runner for the WP org appears to be cancelled https://app.travis-ci.com/github/WordPress/two-factor

kasparsd:
Fixes #435.

- Remove `wp-dev-lib` dependency in favour of `wp-env`.
- Introduce local phpunit setup which runs inside `wp-env`.
- Use the markdown README for contributors and keep `readme.txt` for WP.org.
- Lock to Node 16 for all tooling.


.eslintrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": [ "plugin:@wordpress/eslint-plugin/es5" ],
+  "rules": {
+    "comma-dangle": "off"
```
kasparsd, line 4: We keep this for IE compat as that's what we currently have.

.github/workflows/test.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: npm install
+
+      - name: Lint JS
+        run: npm run lint:js
+
+      - name: Lint CSS
+        run: npm run lint:css
+
+      - name: Lint PHP
+        run: npm run lint:php || true # Ignore for now.
```
kasparsd, line 36: There are a bunch of WordPress linter errors that need to be fixed.

jeffpaul:
@kasparsd as noted in https://github.com/WordPress/two-factor/pull/370#issuecomment-1077686004, should this also fix #364?

PR #438, opened 2022-03-24
Add automatic deployments to WP.org

kasparsd:
- The `master` branch goes to `trunk` on every `master` branch and tag commit.
- Tags commits go to tags.


readme.txt
```diff
@@ -4,7 +4,7 @@ Tags: two factor, two step, authentication, login, totp, fido u2f, u2f, email, b
 Requires at least: 4.3
 Tested up to: 5.9
 Requires PHP: 5.6
-Stable tag: trunk
+Stable tag: 0.7.1
```
kasparsd, line 7: Specify versions here to allow trunk being decoupled from stable tags.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,45 @@
+name: Deploy
+
+on:
+  push:
+    # TODO: Limit to master branch for trunk deployments and all tags.
```
jeffpaul, line 5: My usual approach is to limit to this to:
> on:
>  release:
>    types: [published]

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '5.6'
+          tools: composer:v2
+
+      - name: Install NPM dependencies
+        run: npm install
+
+      #- name: Authenticate with WordPress.org SVN
+      #  run: svn co --username ${{ secrets.DEPLOY_SVN_USERNAME }} --password "${{ secrets.DEPLOY_SVN_PASSWORD }}" https://plugins.svn.wordpress.org/two-factor/
```
kasparsd, line 35: @jeffpaul Do you have suggestions for credentials we could use here? Is there a bot user account that is used for ["beta" plugins](https://wordpress.org/plugins/browse/beta/) like this one?

I see that the performance plugin appears to be using [a set of credentials](https://github.com/WordPress/performance/blob/1410807ccc071c0b00540369f0c1b26debb0032b/.github/workflows/deploy-dotorg.yml#L18-L19) that map to the `performanceteam` user.

jeffpaul: @kasparsd yes, I can coordinate with folks over in Meta to get something set up to be used.  I'll look to get to that probably post 6.1 Beta next week when I'll have a bit more focus I can direct here.

jeffpaul: Requested secrets here: https://meta.trac.wordpress.org/ticket/6492

dd32: @kasparsd SVN_USERNAME + SVN_PASSWORD have been setup, mapping to the `githubsync` WordPress.org user which we use for a few beta plugins with `10up/action-wordpress-plugin-deploy@stable` (Which I would have preferred to see here, but since you have an existing deploy code that already must work.. no objections to using it)

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Install NPM dependencies
+        run: npm install
+
+      - name: Authenticate with WordPress.org SVN
+        run: svn info --non-interactive --username "${{ secrets.DEPLOY_SVN_USERNAME }}" --password "${{ secrets.DEPLOY_SVN_PASSWORD }}" https://plugins.svn.wordpress.org/two-factor/
+
+      - name: Deploy to WordPress.org SVN
+        env:
+          DEPLOY_SVN_USERNAME: ${{ secrets.DEPLOY_SVN_USERNAME }}
```
dd32, line 41: Per https://github.com/WordPress/two-factor/pull/438#discussion_r974860345 this is all I think needs changing to use the org provisioned secrets:

```suggestion
        run: svn info --non-interactive --username "${{ secrets.SVN_USERNAME }}" --password "${{ secrets.SVN_PASSWORD }}" https://plugins.svn.wordpress.org/two-factor/

      - name: Deploy to WordPress.org SVN
        env:
          DEPLOY_SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,46 @@
+name: Deploy
+
+on:
+  push:
+    branches:
+      - master
+      # TODO: Uses for testing this pull request only. Remove before merge!
```
jeffpaul, line 7: Reminder that we'll want to remove this comment and related line of code before merging this in.

jeffpaul (changes requested):
Two changes and then I think this is likely ready to merge.

Gruntfile.js
```diff
@@ -33,10 +33,14 @@ module.exports = function( grunt ) {
...
 				build_dir: '<%= dist_dir %>',
 				assets_dir: 'assets',
-				deploy_tag: false,
 			},
-			trunk: {
-				deploy_trunk: true,
+			wporg: {
+				options: {
+					skip_confirmation: process.env.DEPLOY_SKIP_CONFIRMATION,
+					svn_user: process.env.DEPLOY_SVN_USERNAME,
```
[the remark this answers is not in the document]
dd32: No, DEPLOY_SVN_USERNAME is the ENV var passed in https://github.com/WordPress/two-factor/pull/438/commits/fb4689cf26a44b363a094471a092f38f9dd61fa6

PR #453, opened 2022-09-08
Hash the login nonce for storage to avoid leaking it via DB access

kasparsd:
Hash the second factor login screen nonce in storage to avoid leaking it via DB access but pass in unhashed to the second factor input form.

- [x] Add tests to verify the nonce checks.
- [x] Ensure every failed nonce check requires a fresh nonce.
- [x] Updating dev deps to the latest versions.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -793,6 +793,9 @@ class Two_Factor_Core {
 	public static function verify_login_nonce( $user_id, $nonce ) {
 		$login_nonce = get_user_meta( $user_id, self::USER_META_NONCE_KEY, true );
 
+		// Require a fresh nonce for each request.
+		self::delete_login_nonce( $user_id );
+
 		if ( ! $login_nonce || empty( $login_nonce['key'] ) || empty( $login_nonce['expiration'] ) ) {
 			return false;
 		}
@@ -800,8 +803,6 @@ class Two_Factor_Core {
 		if ( hash_equals( $login_nonce['key'], self::hash_login_nonce( $nonce ) ) && time() < $login_nonce['expiration'] ) {
 			return true;
 		}
-
-		self::delete_login_nonce( $user_id );
 
 		return false;
 	}
@@ -864,8 +865,6 @@ class Two_Factor_Core {
 			self::login_html( $user, $login_nonce['key'], $_REQUEST['redirect_to'], esc_html__( 'ERROR: Invalid verification code.', 'two-factor' ), $provider );
 			exit;
 		}
-
-		self::delete_login_nonce( $user->ID );
 
 		$rememberme = false;
 		if ( isset( $_REQUEST['rememberme'] ) && $_REQUEST['rememberme'] ) {
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -759,7 +759,7 @@ class Two_Factor_Core {
 		}
 
 		// Store the nonce hashed to avoid leaking it via database access.
-		$login_nonce_stored = $login_nonce;
+		$login_nonce_stored        = $login_nonce;
 		$login_nonce_stored['key'] = self::hash_login_nonce( $login_nonce['key'] );
 
 		if ( ! update_user_meta( $user_id, self::USER_META_NONCE_KEY, $login_nonce_stored ) ) {
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -550,7 +550,7 @@ class Two_Factor_Core {
 		}
 
 		if ( true !== self::verify_login_nonce( $user->ID, $nonce ) ) {
-			wp_safe_redirect( get_bloginfo( 'url' ) );
+			wp_safe_redirect( home_url() );
 			exit;
 		}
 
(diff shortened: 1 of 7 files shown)


class-two-factor-core.php
```diff
@@ -779,12 +796,13 @@ public static function verify_login_nonce( $user_id, $nonce ) {
 			return false;
 		}
 
-		if ( $nonce !== $login_nonce['key'] || time() > $login_nonce['expiration'] ) {
-			self::delete_login_nonce( $user_id );
-			return false;
+		if ( hash_equals( self::hash_login_nonce( $nonce ), $login_nonce['key'] ) && time() < $login_nonce['expiration'] ) {
```
kasparsd, line 799: Use a positive check for readability.

class-two-factor-core.php
```diff
@@ -737,15 +748,21 @@ public static function login_url( $params = array(), $scheme = 'login' ) {
...
 			$login_nonce['key'] = bin2hex( random_bytes( 32 ) );
 		} catch ( Exception $ex ) {
 			$login_nonce['key'] = wp_hash( $user_id . wp_rand() . microtime(), 'nonce' );
 		}
-		$login_nonce['expiration'] = time() + HOUR_IN_SECONDS;
 
-		if ( ! update_user_meta( $user_id, self::USER_META_NONCE_KEY, $login_nonce ) ) {
+		// Store the nonce hashed to avoid leaking it via database access.
+		$login_nonce_stored = $login_nonce;
+		$login_nonce_stored['key'] = self::hash_login_nonce( $login_nonce['key'] );
```
kasparsd, line 763: Here we hash the nonce for storage and pass the original to the form.

class-two-factor-core.php
```diff
@@ -737,15 +748,21 @@ public static function login_url( $params = array(), $scheme = 'login' ) {
...
+
+		if ( ! update_user_meta( $user_id, self::USER_META_NONCE_KEY, $login_nonce_stored ) ) {
```
kasparsd, line 765: Store the nonce with the key hashed.

class-two-factor-core.php
```diff
@@ -775,16 +792,19 @@ public static function delete_login_nonce( $user_id ) {
 	 */
 	public static function verify_login_nonce( $user_id, $nonce ) {
 		$login_nonce = get_user_meta( $user_id, self::USER_META_NONCE_KEY, true );
-		if ( ! $login_nonce ) {
+
+		// Require a fresh nonce for each request.
+		self::delete_login_nonce( $user_id );
```
kasparsd, line 797: Is there a reason to not require a new nonce for every login attempt?

class-two-factor-core.php
```diff
@@ -846,8 +866,6 @@ public static function login_form_validate_2fa() {
 			exit;
 		}
 
-		self::delete_login_nonce( $user->ID );
```
kasparsd, line 849: Nonces are now deleted whenever they are validated.

class-two-factor-core.php
```diff
@@ -737,15 +748,21 @@ public static function login_url( $params = array(), $scheme = 'login' ) {
 	 * @return array
 	 */
 	public static function create_login_nonce( $user_id ) {
-		$login_nonce = array();
+		$login_nonce = array(
+			'expiration' => time() + HOUR_IN_SECONDS,
```
kasparsd, line 752: Do we really need a nonce that is valid for an hour? How about we reduce it to 10 minutes or something like that?

composer.json
```diff
@@ -10,27 +10,32 @@
...
     },
     "allow-plugins": {
-      "roots/wordpress-core-installer": true,
-      "dealerdirect/phpcodesniffer-composer-installer": true
+      "dealerdirect/phpcodesniffer-composer-installer": true,
+      "roots/wordpress-core-installer": true
     }
   },
+  "extra": {
+    "wordpress-install-dir": "wordpress"
```
kasparsd, line 23: This was needed after updating to `roots/wordpress-full` which has the core themes bundled.

georgestephanis (approved):

calvinalkan:
@jeffpaul @kasparsd 

This looks good at first glance

Although I would like the $user_id and $expiration to be part of the MAC to absolutely guarantee that a given "challenge_token" can only ever be used for the user that created it and that the expiration can't be tampered with either.

```php

$data = [
   'key' => bin2hex(random_bytes(32)), 
   'user_id' => $user->ID,
   'expiration' => $expiration
]

$hashed_validator = hash_hmac('sha256', json_encode($data, JSON_THROW_ON_ERROR), wp_salt('nonce'));

$store_me = [
 'expiration => $expiration, 
 'hashed_validator => $hashed_validator
]

```

On the validation side you just reconstruct the validator from the single parts

jeffpaul:
@calvinalkan we're working on another release in the (hopefully) near-term, so if you are able to push up a PR for your suggestion then I can coordinate to get review/merge on that... thanks!

class-two-factor-core.php
iandunn: Addressed in #473 👍🏻

PR #462, opened 2022-10-04
Add logging function when failed to authenticate

### LINKED ISSUE #459: Log or alert on failed 2FA codes
Logging a placeholder issue from insight shared from @georgestephanis after finding a [related tweet on this topic](https://twitter.com/SarahDVictoria/status/1570767155490324481)... We should fire off a log or alert to site admins on any failed 2FA code.  Or an error_log or something.  So if someone has a password but is trying to brute force a code it can get caught.

Lucisu:
I've added the method `is_valid_authcode` to the class `Two_Factor_Provider` so providers can use it when failing to validate the authentication.

By default, it uses the `error_log` function, but there are actions and filters that can be used to extend it.

Related issue: #459

### DIFF
--- a/providers/class-two-factor-backup-codes.php
+++ b/providers/class-two-factor-backup-codes.php
@@ -305,7 +305,13 @@ class Two_Factor_Backup_Codes extends Two_Factor_Provider {
 	 */
 	public function validate_authentication( $user ) {
 		$backup_code = isset( $_POST['two-factor-backup-code'] ) ? sanitize_text_field( wp_unslash( $_POST['two-factor-backup-code'] ) ) : false;
-		return $this->validate_code( $user, filter_var( $backup_code, FILTER_SANITIZE_STRING ) );
+		$success = $this->validate_code( $user, filter_var( $backup_code, FILTER_SANITIZE_STRING ) );
+
+		if ( ! $success ) {
+			$this->log_failure( $user, $backup_code );
+		}
+
+		return $success;
 	}
 
 	/**
--- a/providers/class-two-factor-provider.php
+++ b/providers/class-two-factor-provider.php
@@ -93,7 +93,7 @@ abstract class Two_Factor_Provider {
 		do_action( 'two_factor_user_login_failed', $user, $code, $this );
 
 		/* translators: %1$d: the user's ID %2$s: the code used to authenticate */
-		$log_message = sprintf( esc_html__( 'The user with ID %1$d failed to login using the code "%2$s"', 'two-factor' ), $user->ID, esc_html( $code ) );
+		$log_message = sprintf( esc_html__( 'The user %1$s (ID: %2$d) failed to login using the code "%3$s"', 'two-factor' ), esc_html( $user->user_login ), $user->ID, esc_html( $code ) );
 
 		/**
 		 * This filter is triggered to checke whether it's needed to log the authentication failure.
--- a/providers/class-two-factor-provider.php
+++ b/providers/class-two-factor-provider.php
@@ -73,6 +73,39 @@ abstract class Two_Factor_Provider {
 	abstract public function validate_authentication( $user );
 
 	/**
+	 * Logs the failed authentication.
+	 *
+	 * @param WP_User      $user WP_User object of the user trying to login.
+	 * @param string|false $code The code used to authenticate, if available.
+	 *
+	 * @return void
+	 */
+	public function log_failure( $user, $code = false ) {
+		/**
+		 * This action is triggered when a Two Factor validation fails.
+		 *
+		 * @param WP_User      $user WP_User object of the user trying to login.
+		 * @param string|false $code The code used to authenticate, if available.
+		 */
+		do_action( 'two_factor_user_login_failed', $user, $code );
+
+		/* translators: %1$d: the user's ID %2$s: the code used to authenticate */
+		$log_message = sprintf( esc_html__( 'The user with ID %1$d failed to login using the code "%2$s"', 'two-factor' ), $user->ID, esc_html( $code ) );
+
+		/**
+		 * This action is triggered when a Two Factor validation fails.
+		 *
+		 * @param boolean      $should_log  Whether or not the authentication failure should be logged.
+		 * @param WP_User      $user        WP_User object of the user trying to login.
+		 * @param string|false $code        The code used to authenticate, if available.
+		 * @param string       $log_message The generated log message.
+		 */
+		if ( apply_filters( 'two_factor_log_failure', true, $user, $code, $log_message ) ) {
+			error_log( $log_message );
+		}
+	}
+
+	/**
 	 * Whether this Two Factor provider is configured and available for the user specified.
 	 *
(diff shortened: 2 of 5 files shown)


Lucisu:
It's done.

For me, it would make more sense to call a method inside the Two_Factor_Provider class to validate the authentication, which would call the child provider method `validate_authentication`, getting its result and logging the failure in the parent method if needed, instead of adding to every new provider the code:

```php
if ( ! $success ) {
	$this->log_failure( $user, $response );
}
return $success;
```

Please, let me know what you think about this.

kasparsd (commented):
How about we move the logging to the main conditional here:

https://github.com/WordPress/two-factor/blob/a112a0a1001fd0ee7ffa4b6375a82d572c15dc0f/class-two-factor-core.php#L857-L867

so that (1) we don't need to modify each provider to add this and (2) providers are not responsible for logging.

We could also update the `validate_authentication()` methods to return `WP_Error` objects which could provide their own error messages which can then be used by the logger as necessary.

providers/class-two-factor-provider.php
```diff
@@ -72,6 +72,43 @@ public function pre_process_authentication( $user ) {
...
+		 * This filter is triggered to checke whether it's needed to log the authentication failure.
+		 *
+		 * @param boolean      $should_log  Whether or not the authentication failure should be logged.
+		 * @param WP_User      $user        WP_User object of the user trying to login.
+		 * @param string|false $code        The code used to authenticate, if available.
+		 * @param string       $log_message The generated log message.
+		 * @param Two_Factor_Provider $this The Provider class used during the authentication.
+		 */
+		if ( apply_filters( 'two_factor_log_failure', true, $user, $code, $log_message, $this ) ) {
+			error_log( $log_message );
```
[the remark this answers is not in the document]
kasparsd: @jeffpaul, @georgestephanis What are your thoughts on this?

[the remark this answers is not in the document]
georgestephanis: My only uncertainty is if it's a good idea to include the bad code in this.  I don't have any strong opposition, but I'm slightly concerned that if it's one-digit off, or enough times combined with the timestamp being one tick off, something may be able to be reversed.

But I'm fine with email or not (someone trying to brute force the two factor code may be ... uhh ... a lot of emails though) -- but I definitely think error_log should be included.

Also, appreciate the filter so a third-party catcher could also listen for failures for alternate logs.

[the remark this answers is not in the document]
jeffpaul: Maybe support the admin_email but have it disabled and filterable to enable?  I think I'd want it enabled on sites I'm using, but I could see where that could get VERY chatty for someone with a large amount of users?

georgestephanis:
That sounds fair.

iandunn:
Once #427 lands, logging should be added there too.

providers/class-two-factor-provider.php
```diff
@@ -72,6 +72,43 @@ public function pre_process_authentication( $user ) {
...
+		 * This action is triggered when a Two Factor validation fails.
+		 *
+		 * @param WP_User             $user WP_User object of the user trying to login.
+		 * @param string|false        $code The code used to authenticate, if available.
+		 * @param Two_Factor_Provider $this The Provider class used during the authentication.
+		 */
+		do_action( 'two_factor_user_login_failed', $user, $code, $this );
+
+		/* translators: %1$d: the user's ID %2$s: the code used to authenticate */
+		$log_message = sprintf( esc_html__( 'The user with ID %1$d failed to login using the code "%2$s"', 'two-factor' ), $user->ID, esc_html( $code ) );
```
iandunn, line 96: Would username be better here? Otherwise it's not very informative unless the recepient has IDs memorized, or manually look it up.

providers/class-two-factor-provider.php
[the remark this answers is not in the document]
iandunn: It might also depend on the provider; for TOTP it's probably not uncommon to make a typo, and getting an email for that seems like overkill. Emailing after 3 failed attempts could be more reasonable (as long as there's a cap on how often it's sent), but IMO #477 seems like a better way to address brute force attacks.

If a backup code fails, though, I can see it being more useful to email, because it's very out of the ordinary. Related #476. 

WebAuthn keys (#427) could also be worth emailing about, since they should be less prone to accidental failure, especially with user verification/PINs disabled.

-------------------------------

[I don't think gaining access to several TOTPs would let you reverse the secret](https://security.stackexchange.com/a/162367/8467), but I'm not sure how useful including it is either.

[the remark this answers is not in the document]
Lucisu: I'm afraid that moving the logging function out of the providers will not allow many customizations from them. I can think of some examples where it's useful to allow providers to set their own log data:

1. An email provider can send the email address used to log in instead of the user ID.
2. A provider could define their own "failed amount" before logging in.

providers/class-two-factor-provider.php
Lucisu: Makes sense! I just added it [here](https://github.com/WordPress/two-factor/pull/462/commits/bf6bd6a8d5ef5b9ed5f8d31b20d7c9bddf086cba)

iandunn: Nice 👍🏻

providers/class-two-factor-provider.php
```diff
@@ -72,6 +72,43 @@ public function pre_process_authentication( $user ) {
...
+	 * Logs the failed authentication.
+	 *
+	 * @param WP_User      $user WP_User object of the user trying to login.
+	 * @param string|false $code The code used to authenticate, if available.
+	 *
+	 * @since 0.9.0
+	 *
+	 * @return void
+	 */
+	public function log_failure( $user, $code = false ) {
```
ravinderk, line 85: We should support only `string` as a second param. Is there any benefit we are getting by supporting `boolean` as an alternative value?

providers/class-two-factor-backup-codes.php
```diff
@@ -305,7 +305,13 @@ public function authentication_page( $user ) {
 	 */
 	public function validate_authentication( $user ) {
 		$backup_code = isset( $_POST['two-factor-backup-code'] ) ? sanitize_text_field( wp_unslash( $_POST['two-factor-backup-code'] ) ) : false;
-		return $this->validate_code( $user, filter_var( $backup_code, FILTER_SANITIZE_STRING ) );
+		$success = $this->validate_code( $user, filter_var( $backup_code, FILTER_SANITIZE_STRING ) );
```
ravinderk, line 308: I think we place logging logic in the core class where we validate the authentication provider: https://github.com/WordPress/two-factor/blob/c725c9b2affb3a0092427c09f544febb3f721ad9/class-two-factor-core.php#L1174

Are we getting benefits from adding logging logic here?

ravinderk: We should not worry about the format of logging data or internal; the log function should dump logging data in an error log file.

ravinderk (changes requested):
@Lucisu, I left a few suggestions. Feel free to ask questions.

PR #465, opened 2022-10-07
Add PHP8 linter and fix linter issues

kasparsd:
- Add a Composer script to lint for PHP 8.2 support.
- Add the check to CI, too.
- Fix the CSS linter issues introduced by #387.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -987,7 +987,7 @@ class Two_Factor_Core {
 								<th scope="row"><input id="enabled-<?php echo esc_attr( $class ); ?>" type="checkbox" name="<?php echo esc_attr( self::ENABLED_PROVIDERS_USER_META_KEY ); ?>[]" value="<?php echo esc_attr( $class ); ?>" <?php checked( in_array( $class, $enabled_providers, true ) ); ?> /></th>
 								<th scope="row"><input type="radio" name="<?php echo esc_attr( self::PROVIDER_USER_META_KEY ); ?>" value="<?php echo esc_attr( $class ); ?>" <?php checked( $class, $primary_provider_key ); ?> /></th>
 								<td>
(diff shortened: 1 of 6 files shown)


class-two-factor-core.php
```diff
@@ -672,7 +672,7 @@ public static function login_html( $user, $login_nonce, $redirect_to, $error_msg
 						?>
 						<li>
 							<a href="<?php echo esc_url( $login_url ); ?>">
-								<?php $backup_provider->print_label(); ?>
+								<?php echo esc_html( $backup_provider->get_label() ); ?>
```
kasparsd, line 675: Don't trust the provider to return the label escaped.

class-two-factor-core.php
```diff
@@ -987,8 +987,7 @@ public static function user_two_factor_options( $user ) {
 								<th scope="row"><input id="enabled-<?php echo esc_attr( $class ); ?>" type="checkbox" name="<?php echo esc_attr( self::ENABLED_PROVIDERS_USER_META_KEY ); ?>[]" value="<?php echo esc_attr( $class ); ?>" <?php checked( in_array( $class, $enabled_providers, true ) ); ?> /></th>
 								<th scope="row"><input type="radio" name="<?php echo esc_attr( self::PROVIDER_USER_META_KEY ); ?>" value="<?php echo esc_attr( $class ); ?>" <?php checked( $class, $primary_provider_key ); ?> /></th>
 								<td>
-									<label for="enabled-<?php echo esc_attr( $class ); ?>"><?php $object->print_label(); ?></label>
-
+									<label class="two-factor-method-label" for="enabled-<?php echo esc_attr( $class ); ?>"><?php esc_html( $object->get_label() ); ?></label>
```
kasparsd, line 990: Use a class to target the label for the method selector to avoid generic `label` styles leaking. This looks good now:

methods

PR #466, opened 2022-10-07
[DO NOT MERGE] Test PHP8 linter

kasparsd:
Add back a constant that has been depracted in PHP 8 to confirm that PHPCompatibilityWP is actually working.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -830,7 +830,7 @@ class Two_Factor_Core {
 			exit;
 		}
 
-		$provider = filter_input( INPUT_POST, 'provider', FILTER_CALLBACK, array( 'options' => 'sanitize_text_field' ) );
+		$provider = filter_input( INPUT_POST, 'provider', FILTER_SANITIZE_STRING );
 		if ( $provider ) {
 			$providers = self::get_available_providers_for_user( $user );
 			if ( isset( $providers[ $provider ] ) ) {


class-two-factor-core.php
```diff
@@ -830,7 +830,7 @@ public static function login_form_validate_2fa() {
 			exit;
 		}
 
-		$provider = filter_input( INPUT_POST, 'provider', FILTER_CALLBACK, array( 'options' => 'sanitize_text_field' ) );
+		$provider = filter_input( INPUT_POST, 'provider', FILTER_SANITIZE_STRING );
```
kasparsd, line 833: Why is this not being reported?

composer.json
```diff
@@ -22,14 +22,15 @@
...
+  "minimum-stability": "dev",
   "require": {
     "php": ">=5.6"
   },
   "require-dev": {
     "automattic/vipwpcs": "^2.3",
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7.2",
     "php-coveralls/php-coveralls": "^2.5",
-    "phpcompatibility/phpcompatibility-wp": "^2.1",
+    "phpcompatibility/php-compatibility": "@dev",
```
kasparsd, line 33: Turns out even the dev version of `phpcompatibility/phpcompatibility-wp` is locked at `"phpcompatibility/php-compatibility" : "^9.0",` https://github.com/PHPCompatibility/PHPCompatibilityWP/blob/dbb88b4f397fe2d52af984cae7714d7cd28d5bc3/composer.json#L26 while the PHP8 compatibility lints are only part of the untagged dev version and the latest stable 9.3.5 (released in December 2019) https://github.com/PHPCompatibility/PHPCompatibility/releases/tag/9.3.5 doesn't include them.

kasparsd: Or use this hack https://github.com/PHPCompatibility/PHPCompatibilityWP/issues/40#issuecomment-1139623926

package.json
```diff
@@ -20,7 +20,8 @@
     "format:js": "npm run lint:js -- --fix",
     "format:css": "npm run lint:css -- --fix",
     "test": "npm-run-all test:*",
-    "test:php": "wp-env run phpunit 'composer test --working-dir=/var/www/html/wp-content/plugins/two-factor'",
+    "test:php": "npm run composer test",
+    "composer": "wp-env run phpunit composer --working-dir=/var/www/html/wp-content/plugins/two-factor",
```
kasparsd, line 24: Introduce a new script helper to run any Composer commands inside the dev environment.

Issue #468, opened 2022-10-10
Test coverage seems low


iandunn:
I ran an Xdebug report locally and the overall coverage was `34%` before #427, and `23%` after.

I'm not a coverage zealot, and don't think `100%` is a reasonable, but `100%` over the _critical_ areas seems reasonable for a security plugin, especially before it becomes a canonical plugin or merges into Core.

[`sjinks/wp-two-factor-provider-webauthn` has some e2e tests](https://github.com/sjinks/wp-two-factor-provider-webauthn/tree/8e5f04f/tests/e2e/) that might provide a head start. Adding more unit/integration tests would be nice too.

Using `@codeCoverageIgnore` on the untestable and low-severity areas would also help make it more obvious if everything that _should_ be covered actually is.

Related: #467

iandunn:
i didn't mean to close this with #469 , that just increased TOTP. I think it's worth examining the rest of the plugin as well (except U2F).

iandunn:
Related #497

PR #469, opened 2022-10-12
Increase TOTP coverage

iandunn:
This adds some tests to fill small gaps in coverage, and marks some functions as ignored because they're not testable in a unit test, or aren't worth testing. This increases the coverage of `class-two-factor-totp.php` to `92%`, making it obvious that it is well-tested.

To aid in writing tests, this also adds `phpunit-watcher`, similar to Gutenberg, WordCamp.org, and some other WordPress repos.

See #468

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -14,12 +14,18 @@ Please [report issues](https://github.com/WordPress/two-factor/issues) and [open
 
 Join the `#core-passwords` channel [on WordPress Slack](http://wordpress.slack.com) ([sign up here](http://chat.wordpress.org)).
 
-Here is how to get started:
+To use the provided development environment, you'll first need to install and launch Docker. Once it's running, the next steps are:
 
     $ git clone https://github.com/wordpress/two-factor.git
+    $ cd two-factor
+    $ composer install
     $ npm install
+    $ npm run build
+    $ npm run env start
 
-Then open [a pull request](https://help.github.com/articles/creating-a-pull-request-from-a-fork/) with the suggested changes.
+See `package.json` for other available scripts you might want to use during development, like linting and testing.
+
+When you're ready, open [a pull request](https://help.github.com/articles/creating-a-pull-request-from-a-fork/) with the suggested changes.
 
 ## Deployments
 
--- a/providers/class-two-factor-totp.php
+++ b/providers/class-two-factor-totp.php
@@ -38,7 +38,7 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 	const DEFAULT_TIME_STEP_ALLOWANCE = 4;
 
 	/**
-	 * Chracters used in base32 encoding.
+	 * Characters used in base32 encoding.
 	 *
 	 * @var string
 	 */
@@ -46,6 +46,8 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 
 	/**
 	 * Class constructor. Sets up hooks, etc.
+	 *
+	 * @codeCoverageIgnore
 	 */
 	protected function __construct() {
 		add_action( 'two_factor_user_options_' . __CLASS__, array( $this, 'user_two_factor_options' ) );
@@ -58,6 +60,8 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 
 	/**
 	 * Ensures only one instance of this class exists in memory at any one time.
+	 *
+	 * @codeCoverageIgnore
 	 */
 	public static function get_instance() {
 		static $instance;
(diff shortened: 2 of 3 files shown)


iandunn:
I'll take a closer look tomorrow at the lint-php8 failure tomorrow.

kasparsd:
The PHP8 linter issues are fixed in #466.

composer.json
```diff
@@ -11,8 +11,9 @@
   },
   "config": {
     "sort-packages": true,
+    "process-timeout": 0,
```
kasparsd, line 14: This is set to 300 seconds or 5 minutes by default. I think we should keep it at that in order to avoid long running CI tasks. All cases with linters or tests running for more than 5 minutes should probably be investigated and made to run faster. Keeping the limit in place will ensure that we notice when this happens.

readme.md
```diff
@@ -14,12 +14,18 @@ Please [report issues](https://github.com/WordPress/two-factor/issues) and [open
 
 Join the `#core-passwords` channel [on WordPress Slack](http://wordpress.slack.com) ([sign up here](http://chat.wordpress.org)).
 
-Here is how to get started:
+To use the provided development environment, you'll first need to install and launch Docker. Once it's running, the next steps are:
```
kasparsd, line 17: Thanks for updating these instructions!

composer.json
```diff
@@ -11,8 +11,9 @@
...
     "platform": {
-      "php": "5.6.20"
```
kasparsd, line 15: Maybe leave this to a separate pull request as we'll probably want to bump the versions of phpunit when doing this. The `phpunit-polyfills` are already present but we'll need to update the actual test files to match the new parent class signatures for the phpunit base classes. 

Overall, we could have this platform locked at the version we're using/expecting for all the dev tooling while the phpcompatability checker is used for ensuring that all code is compatible with the lowest version of PHP supported by the WP core, right?

composer.json
```diff
@@ -44,6 +45,7 @@
     "lint": "phpcs",
     "lint-php8": "phpcs -p --standard=PHPCompatibilityWP --runtime-set testVersion 8.0- --extensions=php --ignore='vendor/,wordpress/,node_modules/' .",
     "test": "phpunit",
+    "test:watch": "phpunit-watcher watch < /dev/tty",
```
kasparsd, line 48: It seems that `< /dev/tty` might not be necessary. I was able to start the watcher without it and it worked fine.

package.json
```diff
@@ -19,8 +19,8 @@
     "format:php": "composer format",
     "format:js": "npm run lint:js -- --fix",
     "format:css": "npm run lint:css -- --fix",
-    "test": "npm-run-all test:*",
-    "test:php": "npm run composer test",
+    "test": "npm run composer test",
+    "test:watch": "npm run composer test:watch -- -- --timeout=0",
```
kasparsd, line 23: While resolving the merge conflict, I was able to move the timeout logic here where I assume it was needed. Please correct me if I'm wrong.

phpunit-watcher.yml.dist
```diff
@@ -0,0 +1,9 @@
+watch:
+  directories:
+    - ./providers/
```
kasparsd, line 3: Should we add the root directory to this list to ensure that changes to those files also trigger the runs?

.github/workflows/test.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Setup PHP and Composer
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '5.6'
+          php-version: '7.2'
```
kasparsd, line 23: The php8 compat packages need a later version of PHP. All the linters are running correctly now.

composer.json
iandunn: In the past it's caused problems with the `watch` task, since it's designed to run for hours, but I'll test to make sure that's still the case, and if there's another way around it.

kasparsd: I was able to fix it like so 466370bc928fffa71e6d80f0009cb0ee0ce75bb2. Does that solve it?

package.json
iandunn: Ah that's a great idea, I'll test that out 👍🏻

phpunit-watcher.yml.dist
iandunn: We could, as long as we exclude `vendor` etc. I'll look into that 👍🏻

iandunn:
I removed the commits related to `phpunit-watcher` and rebased against `master`, so I think this is ready for review again. I'll open a new PR w/ the other commits and do some of the other PHP tooling work there.

kasparsd (approved):
Looks great!

kasparsd:
Yes, please merge it in!

PR #472, opened 2022-10-12
Update PHPUnit, Install phpunit-watcher

iandunn:
PHPUnit was updated to the latest version (`9.5`), and function signatures/assertions were updated to match. I also used WP's currently recommended setup/teardown methods and fixtures.

This also installs `phpunit-watcher`, since it makes developing tests much easier. It's used in Gutenberg, WordCamp.org, and several other WP repos.

This was branched off from #469. I belive all of the comments raised there have been addressed:
* Removing the redirect to `/dev/tty` seems to work fine in this environment, but there's more information in https://github.com/spatie/phpunit-watcher/pull/112 if we ever run into it. 
* Watching the root files is working. It seems a little bit slower, but not by a significant amount.
* The `--timeout=0` flag didn't work for me, Composer would still time out after 5 minutes. I was able to get it working by including `Composer\\Config::disableProcessTimeout` as part of the task, though. The background info on that is at https://github.com/spatie/phpunit-watcher/issues/63#issuecomment-545633709.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Setup PHP and Composer
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '5.6'
+          php-version: '7.2'
           tools: composer:v2
 
       - name: Install NPM dependencies
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Setup PHP and Composer
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '7.2'
+          php-version: '7.4'
           tools: composer:v2
 
       - name: Install NPM dependencies
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,6 @@
   },
   "config": {
     "sort-packages": true,
-    "process-timeout": 0,
     "platform": {
       "php": "7.2"
     },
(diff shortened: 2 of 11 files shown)


iandunn:
I'll need to rebase this after #469 is merged, and maybe apply a few of the updates there as well.

.github/workflows/test.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Setup PHP and Composer
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '5.6'
+          php-version: '7.3'
```
kasparsd, line 23: Is there a specific reason to use 7.3 instead of the latest stable 7.4? See https://www.php.net/supported-versions.php

kasparsd (approved):
This looks great!

.github/workflows/test.yml
iandunn: That was the minimum that PHPUnit 9.5 required, but 7.4 sounds good to me too, I'll update it to that 👍🏻

iandunn: Updated in [8cbcd57](https://github.com/WordPress/two-factor/pull/472/commits/8cbcd579e4c26fe789099c398de2f1225049048a). Looks like the checks are still passing 👍🏻

iandunn:
It doesn't look like there are any updates needed from #469, so I'll go ahead and rebase + merge.

PR #473, opened 2022-10-13
Harden login nonces

iandunn:
I noticed a few comments on a closed PR and didn't want them to fall through the cracks:

* [Shorten expiration to ~10 minutes](https://github.com/WordPress/two-factor/pull/453/files#r965622814)
* [Include user ID and expiration in HMAC](https://github.com/WordPress/two-factor/pull/453#issuecomment-1244171674) - cc @calvinalkan 
* ~~It also looks like a valid nonce can be reused many times. IMO it should be deleted once it's been successfully used once. The unit test [says that it checks for this](https://github.com/WordPress/two-factor/blob/be268ba1644d8f5b35761fce790c3b1de1ce86a2/tests/class-two-factor-core.php#L422), but actually checks that an invalid nonce deletes a valid one.~~

I'd appreciate some extra eyes on this one to make sure everything with the HMAC is correct.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -749,7 +749,7 @@ class Two_Factor_Core {
 	 */
 	public static function create_login_nonce( $user_id ) {
 		$login_nonce = array(
-			'expiration' => time() + HOUR_IN_SECONDS,
+			'expiration' => time() + ( 10 * MINUTE_IN_SECONDS ),
 		);
 
 		try {
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -731,12 +731,20 @@ class Two_Factor_Core {
 	/**
 	 * Get the hash of a nonce for storage and comparison.
 	 *
-	 * @param string $nonce Nonce value to be hashed.
+	 * @param array $nonce Nonce array to be hashed. ⚠️ This must contain user ID and expiration,
+	 *                     to guarantee the nonce only works for the intended user during the
+	 *                     intended time window.
 	 *
-	 * @return string
+	 * @return string|false
 	 */
 	protected static function hash_login_nonce( $nonce ) {
-		return wp_hash( $nonce, 'nonce' );
+		$message = wp_json_encode( $nonce );
+
+		if ( ! $message ) {
+			return false;
+		}
+
+		return wp_hash( $message, 'nonce' );
 	}
 
 	/**
@@ -745,10 +753,11 @@ class Two_Factor_Core {
 	 * @since 0.1-dev
 	 *
 	 * @param int $user_id User ID.
-	 * @return array
+	 * @return array|false
 	 */
 	public static function create_login_nonce( $user_id ) {
 		$login_nonce = array(
+			'user_id'    => $user_id,
 			'expiration' => time() + ( 10 * MINUTE_IN_SECONDS ),
 		);
 
@@ -759,14 +768,20 @@ class Two_Factor_Core {
 		}
 
 		// Store the nonce hashed to avoid leaking it via database access.
-		$login_nonce_stored        = $login_nonce;
-		$login_nonce_stored['key'] = self::hash_login_nonce( $login_nonce['key'] );
+		$hashed_key = self::hash_login_nonce( $login_nonce );
 
-		if ( ! update_user_meta( $user_id, self::USER_META_NONCE_KEY, $login_nonce_stored ) ) {
-			return false;
+		if ( $hashed_key ) {
+			$login_nonce_stored = array(
+				'expiration' => $login_nonce['expiration'],
+				'key'        => $hashed_key,
+			);
+
+			if ( update_user_meta( $user_id, self::USER_META_NONCE_KEY, $login_nonce_stored ) ) {
+				return $login_nonce;
+			}
 		}
 
-		return $login_nonce;
+		return false;
 	}
 
 	/**
@@ -797,7 +812,16 @@ class Two_Factor_Core {
 			return false;
 		}
 
-		if ( hash_equals( $login_nonce['key'], self::hash_login_nonce( $nonce ) ) && time() < $login_nonce['expiration'] ) {
+		$unverified_nonce = array(
+			'user_id'    => $user_id,
+			'expiration' => $login_nonce['expiration'],
+			'key'        => $nonce,
+		);
+
+		$unverified_hash = self::hash_login_nonce( $unverified_nonce );
+		$hashes_match    = $unverified_hash && hash_equals( $login_nonce['key'], $unverified_hash );
+
+		if ( $hashes_match && time() < $login_nonce['expiration'] ) {
 			return true;
 		}
 
--- a/tests/class-two-factor-core.php
+++ b/tests/class-two-factor-core.php
@@ -380,7 +380,43 @@ class Test_ClassTwoFactorCore extends WP_UnitTestCase {
 	}
 
 	/**
+	 * @covers Two_Factor_Core::create_login_nonce()
+	 * @covers Two_Factor_Core::hash_login_nonce()
+	 */
+	public function test_invalid_hash_input_fails() {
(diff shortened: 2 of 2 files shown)


calvinalkan:
HMAC logic looks good to be, only change would be to fail hard If wp_json_encode returns false.

The return type is string|false.

Tokens not being deleted should be fixed ASAP.

iandunn:
Doh, good catch, thanks! Fixed in 87fdce0

class-two-factor-core.php
```diff
@@ -731,12 +731,20 @@ public static function login_url( $params = array(), $scheme = 'login' ) {
...
 	 */
 	protected static function hash_login_nonce( $nonce ) {
-		return wp_hash( $nonce, 'nonce' );
+		$message = wp_json_encode( $nonce );
+
+		if ( ! $message ) {
+			return false;
+		}
+
+		return hash_hmac( 'sha256', $message, wp_salt( 'nonce' ) );
```
kasparsd, line 747: What is is the PHP support for `hash_hmac` with this specific hash function? Is this a breaking change? What are the benefits compared to using `wp_hash()` helper for this?

class-two-factor-core.php
```diff
@@ -745,11 +753,12 @@ protected static function hash_login_nonce( $nonce ) {
...
 	 *
 	 * @param int $user_id User ID.
-	 * @return array
+	 * @return array|false
 	 */
 	public static function create_login_nonce( $user_id ) {
 		$login_nonce = array(
-			'expiration' => time() + HOUR_IN_SECONDS,
+			'user_id'    => $user_id,
+			'expiration' => time() + ( 10 * MINUTE_IN_SECONDS ),
```
kasparsd, line 761: I'm wondering if we should store the creation time here and verify the expiration at the login state? This would allow adjusting the expiration logic based on certain conditionals via filters, for example.

class-two-factor-core.php
```diff
@@ -797,14 +812,24 @@ public static function verify_login_nonce( $user_id, $nonce ) {
...
+		$unverified_hash = self::hash_login_nonce( $unverified_nonce );
+		$hashes_match    = $unverified_hash && hash_equals( $login_nonce['key'], $unverified_hash );
+
+		$verified = false;
+		if ( $hashes_match && time() < $login_nonce['expiration'] ) {
+			$verified = true;
 		}
 
-		// Require a fresh nonce if verification fails.
+		// Require a fresh nonce if verification fails, and don't allow valid nonces to be reused.
```
kasparsd, line 829: Valid nonces are already being purged here if I'm not mistaken:

https://github.com/WordPress/two-factor/blob/79c364169e4e7fc430c4ed05c9cc81402955785c/class-two-factor-core.php#L869

So this was specifically dealing with the invalid nonces as it wouldn't reach that point during the invalid logic flow.

kasparsd (commented):
Left some initial feedback. Looks good overall!

class-two-factor-core.php
iandunn: Ah, you're right, I was looking for that in the wrong place. I'll update it. Thanks!

class-two-factor-core.php
iandunn: [It works all the way down to PHP 5.2](https://3v4l.org/Wr8c4#v5.2.0)

class-two-factor-core.php
iandunn: I removed that in 5f37351, but kept the part where the test is renamed for accuracy.

It doesn't look like `login_form_validate_2fa()` is testable, because it's not possible to mock `INPUT_POST`, but that might be something we could test in the future w/ e2e tests. #427 might bring in some from `sjinks/wp-two-factor-provider-webauthn`

class-two-factor-core.php
iandunn: 🤔 I don't have a strong opinion, except that that might be a bit outside the scope of this PR, and there're a lot of other things to do for `0.8`. Would it be better to open a ticket for that as a future iteration?

Alternatively, something quick like this might cover the use cases you were thinking of, and be forwards-compatible for future expansion:

```php
function __construct() {
	self::LOGIN_NONCE_EXPIRATION = apply_filters( 'two_factor_login_nonce_expiration', 10 * MINUTE_IN_SECONDS );
}

public static function create_login_nonce( $user_id ) {
	$login_nonce = array(
		'user_id'    => $user_id,
		'expiration' => time() + self::LOGIN_NONCE_EXPIRATION,
	);
}
```

I'm open to either, though. Do you have a preference?

dd32: Core uses `expiration` in the user auth cookies, in preference for created + expiration being filtered after-the-fact.

class-two-factor-core.php
dd32: Not quite, the WordPress included `hash_hmac()` fallback only supports `md5` and `sha1`.

https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/compat.php#L253-L271

For this reason, core uses sha256 and falls back to sha1 in auth:

https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/pluggable.php#L761-L767

The compat is still included as [the hash extension can be disabled until PHP 7.4](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/compat.php#L228-L233).

class-two-factor-core.php
kasparsd: OK, let's leave it as is for this changeset.

class-two-factor-core.php
kasparsd: Should we stick to `wp_hash()` here since the general idea is to (1) _not_ have the token in plain text and (2) have it work on as many systems as possible? These are short lived tokens and it is seems to be more important that they get hashed than which hash function is used. Before this change we used `wp_hash()` which "guaranteed" that it will be hashed as expected.

The addition of user ID and the timestamp to the hashed payload is the important part here, right?

iandunn: Ah, great point Dion 👍🏻 

🤔 , I definitely prefer using Core functions in general, this just seemed like an exception given how insecure MD5 is.  Now that I read about it more, though, it does seem like it's safe-ish in an HMAC context (if I'm understanding everything correctly).

* https://crypto.stackexchange.com/a/9340/9297
* https://crypto.stackexchange.com/a/6753/9297 - in our case the `NONCE_SALT` key length is only 64 chars, but [that's still impractical against a remote network](https://www.wolframalpha.com/input?i=2%5E63+%2F+150000)

I'm still leaning towards erring on the side of caution, though; especially given how easy it is to use [Core's progressive enhancement technique](https://github.com/WordPress/wordpress-develop/blob/3977b6b06d1efd3f6cadb8b31bc8ba55e09486d5/src/wp-includes/pluggable.php#L764). That's partially just because I'm far from an expert here, and I'd rather be safe than sorry.

calvinalkan: You can also use sodium_crypto_generic_hash

dd32: The primary reason the Auth cookies above use `hash_hmac()` over `wp_hash()` is that they need to use a custom `$key` that changes per-user in addition to per-site (which is why the key also uses `wp_hash()` in it), rather than `wp_hash()` which just uses the static per-site key.

In this case, `wp_hash()` (which uses `hash_mac( 'md5' )`) is secure IMO for this purposes. I don't think this is a scenario where `md5 is insecure so we shouldn't use it` holds true.

iandunn: That works for me, I moved it back to `wp_hash()` in [b153455](https://github.com/WordPress/two-factor/pull/473/commits/b153455dfab0ee7b2624d44e9c5510e783b47dd9)

iandunn:
I'm gonna go ahead and merge this since it seems like everything has been resolved. We can definitely iterate if anyone has any more thoughts though.

PR #474, opened 2022-10-13
Tests: Explicitly ignore directories for performance

iandunn:
These should be implicitly ignored because of the `ignoreDotFiles`, `ignoreVCS`, and `ignoreVCSIgnored` settings, but some of them weren't for some reason. That was causing a noticable delay between when a file was saved and when the tests relaunched.

Previously #472


kasparsd (approved):
Nice!

Issue #467, opened 2022-10-07
Document code coverage report


iandunn:
I tried to find the report by guessing what the coveralls.io URL would be, but wasn't successful. Adding it to the readme could be helpful. I'm happy to open a PR if someone can share the URL.

iandunn:
Thanks! It doesn't seem to be showing any reports, so I guess documenting should probably wait until it's configured 👍🏻 

Related #468

iandunn:
Related #364

jeffpaul:
Are unit tests and related code coverage reports as useful as say documented critical flows and associated e2e tests with some basic test report formatting?

PR #475, opened 2022-10-13
Tests: Migrate config to 9.5 format

iandunn:
I missed this in #472, and this was being shown when running tests:

> Warning:       Your XML configuration validates against a deprecated schema.
> Suggestion:    Migrate your XML configuration using "--migrate-configuration"!

I haven't tested Coveralls since it doesn't seem to be configured (#364, #467), but I verified that the log file is still being generated.


kasparsd (approved):
Looks good!

PR #478, opened 2022-10-17
Pass a WP_Error object to the `wp_login_failed` filter.

### LINKED ISSUE #471: Inproper use of the wp_login_failed action (conflict with other plugins)
Hi,

I have found a conflict between Two Factor and Publigo plugins. The problem is caused probably by improper use of the `wp_login_failed` action in Two Factor plugin.

As codex says - `wp_login_failed` uses two params: 

* username (string)
* error (WP_Error instance)

ref. https://developer.wordpress.org/reference/hooks/wp_login_failed/

But Two Factor plugin uses only one param:

https://github.com/WordPress/two-factor/blob/master/class-two-factor-core.php#L858

And it causes a problem if somebody expects 2 params in their custom handling of the `wp_login_failed` action.

For Publigo it causes the following PHP error (sorry for polish logs, but as I suppose the core messages are in english):

```
Błąd typu E_ERROR został spowodowany w linii 56 pliku /[REDACTED]/public_html/wp-content/plugins/wp-idea/includes/events/handlers/class-auth-events-handler.php](/[REDACTED]/public_html/wp-content/plugins/wp-idea/includes/events/handlers/class-auth-events-handler.php). 
Komunikat błędu: Uncaught ArgumentCountError: Too few arguments to function bpmj\wpidea\events\handlers\Auth_Events_Handler::bpmj\wpidea\events\handlers\{closure}(), 1 passed in /[REDACTED]/public_html/wp-includes/class-wp-hook.php](/[REDACTED]/public_html/wp-includes/class-wp-hook.php) on line 307 and exactly 2 expected in /[REDACTED]/public_html/wp-content/plugins/wp-idea/includes/events/handlers/class-auth-events-handler.php:56](/[REDACTED]/public_html/wp-content/plugins/wp-idea/includes/events/handlers/class-auth-events-handler.php:56)
Stack trace:
#0 /[REDACTED]/public_html/wp-includes/class-wp-hook.php(307)](/[REDACTED]/public_html/wp-includes/class-wp-hook.php(307)): bpmj\wpidea\events\handlers\Auth_Events_Handler->bpmj\wpidea\events\handlers\{closure}('usrob')
#1 /[REDACTED]/public_html/wp-includes/class-wp-hook.php(331)](/[REDACTED]/public_html/wp-includes/class-wp-hook.php(331)): WP_Hook->apply_filters('', Array)
#2 /[REDACTED]/public_html/wp-includes/plugin.php(476)](/[REDACTED]/public_html/wp-includes/plugin.php(476)): WP_Hook->do_action(Array)
#3 /[REDACTED]/public_html/wp-content/plugins/two-factor/class-two-factor-core.php(858)](/[REDACTED]/public_html/wp-content/plugins/two-factor/class-two-factor-core.php(858)):
```

dd32:
Fixes #471

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -855,7 +855,7 @@ class Two_Factor_Core {
 
 		// Ask the provider to verify the second factor.
 		if ( true !== $provider->validate_authentication( $user ) ) {
-			do_action( 'wp_login_failed', $user->user_login, new WP_Error( 'two_factor_invalid', esc_html__( 'ERROR: Invalid verification code.', 'two-factor' ) ) );
+			do_action( 'wp_login_failed', $user->user_login, new WP_Error( 'two_factor_invalid', __( 'ERROR: Invalid verification code.', 'two-factor' ) ) );
 
 			$login_nonce = self::create_login_nonce( $user->ID );
 			if ( ! $login_nonce ) {
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -855,7 +855,7 @@ class Two_Factor_Core {
 
 		// Ask the provider to verify the second factor.
 		if ( true !== $provider->validate_authentication( $user ) ) {
-			do_action( 'wp_login_failed', $user->user_login );
+			do_action( 'wp_login_failed', $user->user_login, new WP_Error( 'two_factor_invalid', esc_html__( 'ERROR: Invalid verification code.', 'two-factor' ) ) );
 
 			$login_nonce = self::create_login_nonce( $user->ID );
 			if ( ! $login_nonce ) {


kasparsd (approved):

class-two-factor-core.php
```diff
@@ -855,7 +855,7 @@ public static function login_form_validate_2fa() {
 
 		// Ask the provider to verify the second factor.
 		if ( true !== $provider->validate_authentication( $user ) ) {
-			do_action( 'wp_login_failed', $user->user_login );
+			do_action( 'wp_login_failed', $user->user_login, new WP_Error( 'two_factor_invalid', esc_html__( 'ERROR: Invalid verification code.', 'two-factor' ) ) );
```
kasparsd, line 858: It appears that core doesn’t escape the error message when returning it as part of the error object:

https://github.com/WordPress/wordpress-develop/blob/6.0.2/src/wp-includes/pluggable.php#L625

I think we should also let it be escaped later, if needed by the action hook user method.

dd32: To be honest, Core correctly doesn't escape any static string, and I believe not escaping translations is the correct thing to do.

I only escaped it here to be inline with the line below where I copied it from. I'll switch it over for consistency.

kasparsd:
This is great, thank you!

PR #482, opened 2022-10-18
Core: Reset compromised passwords after 2FA failures

### LINKED ISSUE #477: Throttle second factor attempts
TIL, [brute-forcing TOTP](https://lukeplant.me.uk/blog/posts/6-digit-otp-for-two-factor-auth-is-brute-forceable-in-3-days/) is [much easier](https://pulsesecurity.co.nz/articles/totp-bruting) than I assumed (yay for checking assumptions!). That applies to other providers as well to varying degrees.

It seems like a simple solution would be to throttle attempts with increasing severity:

| # of Failed Attempts  | Delay (seconds) |
| ------------- | ------------- |
| 0  | 0  |
| 1  | 0  |
| 2  | 1  |
| 3  | 3  |
| 4  | 5  |
| 5  | 10  |
| 6  | 20  |

... and so on. There could be a filter on that mapping of `attempts -> delay`, to allow customization. Implementing it should be straightforward with a `usermeta` counter, and a `sleep()` inside `login_form_validate_2fa()`, and some kind of notice to the user.

After ~8 attempts, though, it may be safest to assume that the password has been compromised, and to force a reset and email the user.

iandunn:
Fixes #477 
Builds on top of #510

After a reasonable number of 2nd-factor attempts, it's safe to assume that the password been compromised and an attacker is trying to brute force the 2nd factor.

`get_user_time_delay()` mitigates brute force attempts, but many 2nd factors -- like TOTP and backup codes -- are very weak on their own, so it's not safe to give attackers unlimited attempts.

Since we know that the password is compromised, we have the responsibility to reset it and inform the user. That will guarantee that attackers can't brute force it (unless they compromise the new password).

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -651,8 +651,8 @@ class Two_Factor_Core {
 	/**
 	 * Show the password reset notice if the user's password was reset.
 	 *
-	 * This is needed because they may not have received the email notification that was sent in
-	 * `send_password_reset_email()`.
+	 * They were also sent an email notification in `send_password_reset_email()`, but email sent from a typical
+	 * web server is not reliable enough to trust completely.
 	 *
 	 * @param WP_Error $errors
 	 */
@@ -1186,7 +1186,7 @@ class Two_Factor_Core {
 		/**
 		 * Filters the maximum number of failed attempts on a 2nd factor before the user's
 		 * password will be reset. After a reasonable number of attempts, it's safe to assume
-		 * that the password been compromised and an attacker is trying to brute force the 2nd
+		 * that the password has been compromised and an attacker is trying to brute force the 2nd
 		 * factor.
 		 *
 		 * ⚠️ `get_user_time_delay()` mitigates brute force attempts, but many 2nd factors --
@@ -1204,13 +1204,9 @@ class Two_Factor_Core {
 	/**
 	 * Reset a compromised password.
 	 *
-	 * After a reasonable number of 2nd-factor attempts, it's safe to assume that the password been compromised
-	 * and an attacker is trying to brute force the 2nd factor. `is_user_rate_limited()` mitigates brute force
-	 * attempts, but many 2nd factors -- like TOTP and backup codes -- are very weak on their own, so it's not
-	 * safe to give attackers unlimited attempts.
-	 *
 	 * If we know that the the password is compromised, we have the responsibility to reset it and inform the
-	 * user. That will guarantee that attackers can't brute force it (unless they compromise the new password).
+	 * user. `get_user_time_delay()` mitigates brute force attempts, but this acts as an extra layer of defense
+	 * which guarantees that attackers can't brute force it (unless they compromise the new password).
 	 *
 	 * @param WP_User $user The user who failed to login
 	 */
@@ -1244,7 +1240,7 @@ class Two_Factor_Core {
 		$notify_admin = apply_filters( 'two_factor_notify_admin_user_password_reset', true );
 		$admin_email  = get_option( 'admin_email' );
 
-		if ( $user->user_email !== $admin_email && $notify_admin ) {
+		if ( $notify_admin && $admin_email !== $user->user_email ) {
 			self::notify_admin_user_password_reset( $user );
 		}
 	}
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -50,6 +50,13 @@ class Two_Factor_Core {
 	const USER_FAILED_LOGIN_ATTEMPTS_KEY = '_two_factor_failed_login_attempts';
 
 	/**
+	 * The user meta key to store whether or not the password was reset.
+	 *
+	 * @var string
+	 */
+	const USER_PASSWORD_WAS_RESET_KEY = '_two_factor_password_was_reset';
+
+	/**
 	 * URL query paramater used for our custom actions.
 	 *
 	 * @var string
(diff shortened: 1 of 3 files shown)


pkevan:
Do we want to consider the attempt counter to be using a transient to benefit those sites running with memcache etc, rather than it always being the DB? Looking at the suggested alternatives [metioned](https://github.com/WordPress/two-factor/issues/477#issuecomment-1281724184) that seems like the approach taken for passwords and considering a large site, and the worst case scenario (everything is compromised), then it does open it to resource saturation.

The other thing regarding resetting passwords and emailing the reset link - is this ok? what if the account is compromised via password and the email address is updated? Or am I thinking too far ahead of the scenarios scope?

pkevan:
> Wouldn't the attacker be prevented from changing the address until they fully authenticate (including 2FA) ? [Logging in via the REST API and XML-RPC are disabled by default](https://github.com/WordPress/two-factor/blob/6e7265e41068a8c85ddf10442b373e59d71f4992/class-two-factor-core.php#L462-L480).

Yes, I was considering those sites who had disabled the functionality for various reasons, like syncing content for search, or applications unable to handle a 2FA workflow .

iandunn:
I switched to a transient in 2d5c53bb92282818ab24f614c45ea2a338303d12 👍🏻 

> sites who had disabled the functionality

🤔 If the attacker was using the REST API or XML-RPC, then that wouldn't trigger the reset. They could still trigger it if tried to login via the browser.

If they did update the address, then you're right that this wouldn't be effective. It seems like that's the risk the owner takes when enabling REST/XML, though; the entire plugin is essentially disabled at that point, so I'm not sure the user is any worse off. The user would still get a notification from Core (at their actual address) letting them know that it was changed, though.

Am I missing something, though, or is there something you'd suggest instead?

class-two-factor-core.php
```diff
@@ -932,6 +964,140 @@ public static function login_form_validate_2fa() {
...
+		$subject     = sprintf( 'Compromised password for %s has been reset', esc_html( $user->user_login ) );
+
+		$message = sprintf(
+			'Hello, this is a notice from the Two Factor plugin to inform you that an unusually high number of failed login attempts have been detected on the %1$s account (ID %2$d).
+
+			Those attempts successfully entered the user\'s password, and were only blocked because they entered invalid second authentication factors.
+
+			To protect their account, the password has automatically been reset, and they have been notified that they will need to create a new one.
+
+			If you do not wish to receive these notifications, you can disable them with the `two_factor_notify_admin_password_reset` filter. See %3$s for more information.
```
pkevan, line 1092: `two_factor_notify_admin_password_reset` is different from the filter `two_factor_notify_admin_user_password_reset ` (has `_user_` added) on https://github.com/WordPress/two-factor/pull/482/commits/77a015b43d1c1b1297eed61a6116c37def4d685c#diff-ee04f1d323104504c6bfa38dd11ef43e78d1dbac2883293af0b5c310d16e9519R1035

pkevan:
> Am I missing something, though, or is there something you'd suggest instead?

Probably not 😄 I was just working through some scenarios where this is an edge case (Jetpack content sync) but it seems like the approach is to allow only these specifically, so it depends whether the plugin should handle this (probably not) or it's handled on a per-site basis.

class-two-factor-core.php
iandunn: Great catch, thanks! fixed in b44e7f6

iandunn:
@dd32 [raised some good points](https://github.com/WordPress/two-factor/issues/477#issuecomment-1413299605) that would improve this PR. There's also overlap with #510 that could be incorporated:

* Bump the default trigger from `7` to `~30` so that legit users don't have their passwords reset during normal failures
* Add an admin notice like #510 when a user tries to log in with a reset password, in case they never got the email
* Add filters that'd allow another plugin to use atomic operations like `wp_cache_incr()`, or more sophisticated limiting mechanisms that track IP etc.

If [a hybrid approach](https://github.com/WordPress/two-factor/issues/477#issuecomment-1416255118) is desirable, though, then this PR could also be refactored to build on top of failure tracking in #510, rather than implementing a separate mechanism.

iandunn:
45f2040 refactored this PR to build on top of #510, and implement the improvements mentioned above.

dd32:
> Bump the default trigger from 7 to ~30 so that legit users don't have their passwords reset during normal failures

With this change, I'm supportive of this PR approach.

class-two-factor-core.php
```diff
@@ -1116,6 +1173,154 @@ public static function login_form_validate_2fa() {
...
+			These attempts successfully entered your password, and were only blocked because they failed to enter your second authentication factor. Despite not being able to access your account, this behavior indicates that the attackers have compromised your password. The most common reasons for this are that your password was easy to guess, or was reused on another site which has been compromised.
+
+			To protect your account, your password has been reset, and you will need to create a new one. For advice on setting a strong password, please read %3$s
+
+			To pick a new password, please visit %4$s
+
+			This is an automated notification. If you would like to speak to a site administrator, please contact them directly.',
+			esc_html( $user->user_login ),
+			home_url(),
+			'https://wordpress.org/support/article/password-best-practices/',
```
pkevan, line 1268: this link now redirects to: https://wordpress.org/documentation/article/password-best-practices/

iandunn: Good catch! Fixed in 5f66a3e

PR #487, opened 2022-10-25
Generate QR codes locally with JS

### LINKED ISSUE #42: TOTP: Switch to an internal QR Code provider.

### LINKED ISSUE #443: Maybe add a filter hook for the TOTP QR code URL
I'm evaluating this plugin for use in my client's projects. Unfortunately, generating QR codes using googleapis is a hard no-go for them. Having a filter hook in `get_google_qr_code` would allow me to use a local QR generator instead.

Adding a filter hook in that function should be pretty straightforward, and I don't think it would introduce any other issues?

iandunn:
Fixes #42
Closes #388 - This is an alternate approach, which addresses [the concerns I raised there](https://github.com/WordPress/two-factor/pull/388#issuecomment-1284751347).
Closes #443 - There won't be an external URL to filter anymore. This has filters for the local URL, though

I don't think a PHP fallback is needed, since disabling JS is extremely rare, and the manual secret already serves that purpose.


iandunn:
@kasparsd , @dd32 : I'll be AFK until the 2nd, so don't hesitate to make any changes y'all see fit in my absence. I'm also happy to pick up back up when I return.

providers/class-two-factor-totp.php
```diff
@@ -136,9 +152,40 @@ public function user_two_factor_options( $user ) {
...
+					var url    = 'otpauth://totp/' + label + '?secret=' + secret + '&issuer=' + issuer;
+
+					qr.addData( url );
+					qr.make();
+
+					// ⚠️ Intentionally using GIF here instead of SVG, for security. The benefits
+					// of SVG in this situation are minimal, but it would introduce the possibility
+					// of malicious JS being injected into the SVG and causing XSS.
+					document.getElementById( 'two-factor-qr-placeholder' ).innerHTML = qr.createImgTag( 5 );
+				} );
```
dd32, line 186: IMHO, this is overly cautious code, where it could be much simpler and likely more extendible.

In my testing, I felt the GIF images were slightly blurry, as shown here by GIF and SVG side by side, one is just clearly crisper to my eyes.
Screen Shot 2022-11-08 at 4 21 28 pm

```suggestion
				<?php
				// TODO: Move into the above PHP block after $totp_title.
				$totp_args = array(
					'secret' => rawurlencode( $key ),
					'issuer' => rawurlencode( $site_name )
				);
				$totp_args = apply_filters( 'two_factor_totp_parameters', $totp_args, $user );
	
				$totp_url = 'otpauth://totp/' . rawurlencode( $totp_title );
				$totp_url = add_query_arg( $totp_args, $totp_url );
				$totp_url = apply_filters( 'two_factor_totp_url', $totp_url, $user );
				?>
				window.addEventListener( 'DOMContentLoaded', function( event ) {
					/*
					 * 0 = Automatically select the version, to avoid going over the limit of URL
					 *     length.
					 * L = Least amount of error correction, because it's not needed when scanning
					 *     on a monitor, and it lowers the image size.
					 */
					var qr = qrcode( 0, 'L' );

					qr.addData( <?php echo wp_json_encode( $totp_url ); ?> );
					qr.make();

					document.getElementById( 'two-factor-qr-placeholder' ).innerHTML = qr.createSvgTag( 5 );
				} );
```

I did consider if it was worth moving it into an inline object, `{ type: "0", errorCorrectionLevel: "L", url: ".." }`, but that doesn't gain much, and as you say, I don't think anyone would really want to customise the QR type or errorCorrection level (L = 7%).

dd32:
Additionally, It would be worth changing this from just a QR image, to also linking to the otpauth url.
That allows for those setting up on mobile (...and desktop) to simply click the QR image which loads up their Authenticator app.

Ie.
```diff
-<p id="two-factor-qr-placeholder">QR</p>
+<a id="two-factor-qr-placeholder" href="otpauth://....">QR</a>
```

providers/class-two-factor-totp.php
```diff
@@ -78,6 +79,21 @@ public function get_label() {
...
+	public function enqueue_assets() {
+		$environment_prefix = file_exists( TWO_FACTOR_DIR . '/dist' ) ? '/dist' : '';
+
+		wp_enqueue_script(
+			'two-factor-qr-code-generator',
+			plugins_url( $environment_prefix . '/includes/qrcode-generator/qrcode.js', __DIR__ ),
+			array(),
+			TWO_FACTOR_VERSION,
+			true
+		);
```
dd32, line 94: As this is being loaded on all admin pageviews based on the hook in use, this should be a `wp_register_script()` and `wp_enqueue_script()` called during the render below.

iandunn:
> also linking to the otpauth url

That's a good idea, done in 68ed4f2

providers/class-two-factor-totp.php
iandunn: Huh, those two look identical to me, maybe I'm just old 😆 🤷🏻‍♂️ 

I agree that the scope of vulns that this would avoid is very small, though, and it's not worth holding this up, so I changed it in [6f33e71](https://github.com/WordPress/two-factor/pull/487/commits/6f33e71f889c5618b6dfd59a795dc9414d84ccb0)

providers/class-two-factor-totp.php
```diff
@@ -131,14 +151,62 @@ public function user_two_factor_options( $user ) {
...
+
 			?>
+
 			<p>
 				<?php esc_html_e( 'Please scan the QR code or manually enter the key, then enter an authentication code from your app in order to complete setup.', 'two-factor' ); ?>
 			</p>
-			<p>
-				<img src="<?php echo esc_url( $this->get_google_qr_code( $totp_title, $key, $site_name ) ); ?>" id="two-factor-totp-qrcode" />
+			<p id="two-factor-qr-code">
+				<a href="<?php echo $totp_url; ?>">
```
kasparsd, line 179: This needs to be escaped with `esc_url()`.

iandunn: It is escaped [above](https://github.com/WordPress/two-factor/blob/8e016214209739bd8b47994146e85727168bd898/providers/class-two-factor-totp.php#L171) (same w/ the other). Normally I'd escape as late as possible, but this one needs the extra param for the `otpauth` protocol, which would make those `echo`s fairly long and less readable, and duplicated instead of DRY.

I don't feel strongly either way though. I'm happy to change it if you'd prefer.

providers/class-two-factor-totp.php
```diff
@@ -131,14 +151,62 @@ public function user_two_factor_options( $user ) {
...
+					Loading...
+					<img src="<?php echo admin_url('images/spinner.gif'); ?>" alt="" />
```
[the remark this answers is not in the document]
iandunn: 👍🏻 fixed in 7961d39

Issue #385, opened 2020-09-30
Prevent other plugins from accidentally circumventing two factor authentication


kkmuffme:
Since two-factor relies on the default wp_login action, other plugins which use the same action could use an earlier priority, thus leading to two-factor getting circumvented - this results in an unwanted security risk for most WP site owners.

in class-two-factor-core.php 
`add_action( 'wp_login', array( __CLASS__, 'wp_login' ), 10, 2 );`

Since 10 is WP's default priority, various plugins and themes use this action with a priority between -10 to 9 to add a feature - this however means that two-factor might get circumvented by those, which is a major security issue.

I think we should change the priority 10, to -9999, thus ensuring that only devs who actively want to circumvent two-factor authentication also do so.
From a end user's (= WP website owner) perspective this makes the most sense, because: I want to make sure that two-factor is used, no matter which theme/plugins I install. Those should not reduce my site's security by accident.

Theoretically we could also change it to `PHP_INT_MIN` which would make sure that two-factor is never circumvented - I think this is not a good idea though, as there may be cases where a plugin may want to circumvent two-factor.

blogtutor:
Yes, please!  We've noticed that if we use ManageWP's automatic login feature (clicking inside ManageWP to open the site's WordPress Dashboard) it does indeed circumvent Two-Factor. 

(We're already authorized via ManageWP -- and use their 2FA method on our logins there -- but 2FA also isn't mandatory by default in ManageWP.)

r-a-y:
Although dropping the `'wp_login'` hook priority can workaround the issue in the short-term, I think #406 would be better to implement.

iandunn:
🤔 I wonder if there's a way that this plugin could automatically detect if someone is logged in through the UI w/out 2FA being triggered?

If there's a reliable way to do that, it could be more effective than bumping the `wp_login` priority. Or we could do both for extra layers of protection.

Here's a very rough idea:

1. When a user submits the username/password form, we `set_transient( '_two_factor_step_1_user_' . $user_id, true, 0 )`.
2. When their 2nd factor is authenticated, we `delete_transient( '_two_factor_step_1_user_' . $user_id )`.
3. When wp-admin loads, we check if `_two_factor_step_1_user_ . $user_id` exists. If it does, we display an `admin_notice`, or email the site admin, etc. Related: #459, #462, #476

iandunn:
I think this was fixed by #502. Is that right, @dd32 ? If not, please reopen it.

PR #490, opened 2022-11-03
Block creating authentication cookies when inappropriate to do so.

### LINKED ISSUE #406: Using auth_cookie filter instead of wp_login hook to start 2FA flow
Currently flow is started when wp_login is triggered, i.e. when the user has already been logged in, and then reversers the last part of the the default login-flow, by removing the the auth-cookie in function wp_login.

```
public static function wp_login( $user_login, $user ) {
		if ( ! self::is_user_using_two_factor( $user->ID ) ) {
			return;
		}

		// Invalidate the current login session to prevent from being re-used.
		self::destroy_current_session_for_user( $user );

		// Also clear the cookies which are no longer valid.
		wp_clear_auth_cookie();

		self::show_two_factor_login( $user );
		exit;
}

```

Why don't instead use the hook auth_cookie filter, to prevent the cookie from being set unit 2FA has been completed?

Or use wp_authenticate action hook that is triggered before the WP backend authentication process is done, removing need to destroy the session 
I think that use of wp_login hook, in addition to being somewhat backward, as already completed login is reversed,  is more likely to conflict with other hooks in sites that seek to do actions after successful login. 

Ref:
https://usersinsights.com/wordpress-user-login-hooks/
https://developer.wordpress.org/reference/hooks/auth_cookie/
https://developer.wordpress.org/reference/hooks/wp_authenticate/

dd32:
As noted in #406 and #385 it's possible for plugins to accidentally bypass the two-factor requirement by setting cookies directly. 

See https://github.com/WordPress/two-factor/issues/406#issuecomment-1301651343 for more detailed reasonings.

Fixes #406, #385

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -497,7 +497,7 @@ class Two_Factor_Core {
 	}
 
 	/**
-	 * Keep track of the last user a cookie was generated for.
+	 * Reset the last user a cookie was generated for.
 	 *
 	 * When clearing cookies, there is not user context.
 	 */
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -601,7 +601,7 @@ class Two_Factor_Core {
 	public static function login_form_show_2fa() {
 		$wp_auth_id  = filter_input( INPUT_GET, 'wp-auth-id', FILTER_SANITIZE_NUMBER_INT );
 		$nonce       = filter_input( INPUT_GET, 'wp-auth-nonce', FILTER_CALLBACK, array( 'options' => 'sanitize_key' ) );
-		$redirect_to = isset( $_REQUEST['redirect_to'] ) ? $_REQUEST['redirect_to'] : wp_get_referer();
+		$redirect_to = ! empty( $_REQUEST['redirect_to'] ) ? $_REQUEST['redirect_to'] : wp_get_referer();
 		$user        = get_user_by( 'id', $wp_auth_id );
 
 		if ( ! $wp_auth_id || ! $nonce ) {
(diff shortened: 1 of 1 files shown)


kasparsd:
This appears to be impacting the core workflows used by this plugin. Could we please add some unit tests to go with these changes?

dd32:
> Could we please add some unit tests to go with these changes?

I wasn't able to determine how best to add unit testing here, given the plugin doesn't appear to currently test it.

Do you have any suggestions @kasparsd?

iandunn:
#497 might be a better way to test it. I don't think that needs to be a blocker for this, though. We could just test this manually and then add the e2e tests once the tooling is in place (maybe as part of #468 )

iandunn:
This seems to work in my manual tests. @kasparsd , do you have any thoughts on the 2 comments above about automated tests?

class-two-factor-core.php
```diff
@@ -433,13 +444,67 @@ public static function wp_login( $user_login, $user ) {
...
+	 * Keep track of the last user a cookie was generated for.
+	 *
+	 * This is used for when context is not provided via send_auth_cookies.
+	 */
+	public static function set_auth_cookie( $auth_cookie, $expire, $expiration, $user_id ) {
+		self::$last_auth_cookie_user = $user_id;
+	}
+
+	/**
+	 * Keep track of the last user a cookie was generated for.
```
iandunn, line 500:
```suggestion
	 * Reset the last user a cookie was generated for.
```

dd32:
Closing in favour of #502 which has been merged.

PR #491, opened 2022-11-06
Migrate U2F keys to WebAuthn

mcguffin:
This is a proof of concept and a follow-up for #427.

The transformation is happening in `includes/WebAuthn/class-webauthn-key-migrator.php`. The PHP is mainly gathered together from `sjinks/wp-two-factor-provider-webauthn` and `madwizard-org/webauthn-server`.

Apart from reviewing, here's what I think should be the next steps:

 - Key migration is currently triggered by the user in the U2F keys admin table. However we could do it automatically, eg. when the user is visiting his/her profile page, like `sjinks/wp-two-factor-provider-webauthn` does. I'm feeling indecisive about it and I'm looking forward for your feedback and opinions.
 - Add an admin message after a key has been migrated.
 - Real life testing with actual U2F / YubiKeys
 - Unit tests


burnedfaceless:
@dd32 Can you please merge this pull request for me? I would really like to be able to use my U2F Keys again.

dd32:
> dd32 Can you please merge this pull request for me?

This isn't my call. I would suggest using a plugin that adds a WebAuthN provider in the meantime: https://wordpress.org/plugins/two-factor-provider-webauthn/

adampl:
are you serious? waiting 2.5+ years for review? :|

PR #494, opened 2022-11-09
Fix Backup code download with quotes in translations

### LINKED ISSUE #460: Download backup code doesn't generate working download link in french
Because the french traduction contains a single quote, encodeURI didn't encode single quote. The trick is to do the replacement of &#039; with %27, after the encoding.

I had to hack the code as following. I don't know if you have a better solution:

$text = encodeURI( txt_data );
$text = $text.replace( /&#039;/g, '%27');
$( '#two-factor-backup-codes-download-link' ).attr( 'href',  $text);

dd32:
Fixes #460

- `encodeURI()` doesn't encode enough characters for non-english languages.
- `encodeURIComponent()` doesn't encode enough characters for usage within `data:` uri's
- `i18n.title` should not be HTML escaped if it's not going to be used within a HTML context. This results in the translation string having HTML entities. An alternative would be to decode HTML entities client-side in JS and then use it that value, but encoding it in the first place here provides no security benefits.

### DIFF
--- a/providers/class-two-factor-backup-codes.php
+++ b/providers/class-two-factor-backup-codes.php
@@ -229,7 +229,12 @@ class Two_Factor_Backup_Codes extends Two_Factor_Provider {
 	 * @since 0.1-dev
 	 */
 	public function ajax_generate_json() {
-		$user = get_user_by( 'id', filter_input( INPUT_POST, 'user_id', FILTER_SANITIZE_NUMBER_INT ) );
+		$user_id = 0;
+		if ( ! empty( $_POST['user_id'] ) ) {
+			$user_id = absint( $_POST['user_id'] );
+		}
+
+		$user = get_user_by( 'id', $user_id );
 		check_ajax_referer( 'two-factor-backup-codes-generate-json-' . $user->ID, 'nonce' );
 
 		// Setup the return data.
--- a/providers/class-two-factor-backup-codes.php
+++ b/providers/class-two-factor-backup-codes.php
@@ -177,15 +177,23 @@ class Two_Factor_Backup_Codes extends Two_Factor_Provider {
 							// Update counter.
 							$( '.two-factor-backup-codes-count' ).html( response.data.i18n.count );
 
-							// Build the download link.
-							var txt_data = 'data:application/text;charset=utf-8,' + '\n';
-							txt_data += response.data.i18n.title.replace( /%s/g, document.domain ) + '\n\n';
+							// Build the downloaded file contents.
+							var txt_data = response.data.i18n.title.replace( /%s/g, document.domain ) + '\n\n';
 
 							for ( i = 0; i < response.data.codes.length; i++ ) {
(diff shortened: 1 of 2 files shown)


jeffpaul:
Open to review from @iandunn @georgestephanis or @kasparsd to help land this in the 0.8.0 release

providers/class-two-factor-backup-codes.php
```diff
@@ -177,15 +177,23 @@ public function user_options( $user ) {
 							// Update counter.
 							$( '.two-factor-backup-codes-count' ).html( response.data.i18n.count );
 
-							// Build the download link.
-							var txt_data = 'data:application/text;charset=utf-8,' + '\n';
-							txt_data += response.data.i18n.title.replace( /%s/g, document.domain ) + '\n\n';
+							// Build the downloaded file contents.
+							var txt_data = response.data.i18n.title.replace( /%s/g, document.domain ) + '\n\n';
```
kasparsd, line 181: Could we offload this to PHP and make it part of the AJAX response along with `codes`?

https://github.com/WordPress/two-factor/blob/5363ca67fd6121f1a8ae9fea6e266f97782693f7/providers/class-two-factor-backup-codes.php#L241-L261

This would allow us to make that helper testable and avoid all the browser encoding/escaping nuances.

providers/class-two-factor-backup-codes.php
```diff
@@ -177,15 +177,23 @@ public function user_options( $user ) {
...
 
 							for ( i = 0; i < response.data.codes.length; i++ ) {
 								txt_data += i + 1 + '. ' + response.data.codes[ i ] + '\n';
 							}
 
-							$( '#two-factor-backup-codes-download-link' ).attr( 'href', encodeURI( txt_data ) );
+							txt_data = encodeURIComponent( txt_data );
+							// encodeURIComponent() Does not encode -_.!~*'()
+							txt_data = txt_data.replace(/[!'()~*]/g, function(c) {
```
kasparsd, line 189: The comment mentions `-_.` in addition to `!'()~*` included in the current regex -- should those be included too?

kasparsd (approved):
Looks good!

providers/class-two-factor-backup-codes.php
dd32: Good catch!

`encodeURI()` will not encode `~!@#$&*()=:/,;?+'`.
`encodeURUComponent()` will not encode `~!*()'` (ie. everything above that's not here are encoded)
The subset that's missing are `-_.` which neither encode, but are safe left unencoded.

```suggestion
							// encodeURIComponent() does not encode -_.!~*'(), encode the problematic characters
							txt_data = txt_data.replace(/[!'()~*]/g, function(c) {
```

providers/class-two-factor-backup-codes.php
dd32: I think that's a reasonable suggestion.

I think it'd be reasonable to shift the generation of the `data:` uri to PHP as well, and just return that along with the XHR payload, removing most of the JS reliance.

PR #495, opened 2022-11-17
Register the TOTP QR generation script on the front-end.

dd32:
This allows for the renderer to enqueue the TOTP generation on the front-end - Such as within the bbPress settings panel.

Without this, the JS is never loaded and you just see a spinner.

See 70ba42e3bac72597a5b98785b6f11273f2830cb4 & #487


iandunn (approved):
LGTM 👍🏻

lkraav:
Does this help #189?

We're also looking for an alternate solution for iThemes 2FA that has zero frontend management.

iandunn:
It should, yeah. #504 will probably also help with that. There may be more that's needed, though.

PR #502, opened 2022-12-07
Block sending authentication cookies upon 2FA login

dd32:
Based on the feedback in #406 and limited interest on changing how the login flow works in #490, this PR is a minimal patch which prevents sending the authentication cookies upon hitting the 2FA interstitial page.

As there's limited changes in this PR, this should be easier for others to review and approve.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -495,6 +495,10 @@ class Two_Factor_Core {
 	 * @return WP_User|WP_Error
 	 */
 	public static function filter_authenticate_block_cookies( $user ) {
+		/*
+		 * NOTE: The `login_init` action is checked for here to ensure we're within the regular login flow,
+		 * rather than through an unsupported 3rd-party login process which this plugin doesn't support.
+		 */
 		if ( $user instanceof WP_User && self::is_user_using_two_factor( $user->ID ) && did_action( 'login_init' ) ) {
 			add_filter( 'send_auth_cookies', '__return_false', PHP_INT_MAX );
 		}
@@ -919,6 +923,11 @@ class Two_Factor_Core {
 			$rememberme = true;
 		}
 
+		/*
+		 * NOTE: This filter removal is not normally required, this is included for protection against
+		 * a plugin/two factor provider which runs the `authenticate` filter during it's validation.
+		 * Such a plugin would cause self::filter_authenticate_block_cookies() to run and add this filter.
+		 */
 		remove_filter( 'send_auth_cookies', '__return_false', PHP_INT_MAX );
 		wp_set_auth_cookie( $user->ID, $rememberme );
 
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -90,6 +90,9 @@ class Two_Factor_Core {
 		// Run only after the core wp_authenticate_username_password() check.
 		add_filter( 'authenticate', array( __CLASS__, 'filter_authenticate' ), 50 );
 
+		// Run as late as possible to disable sending authentication cookies.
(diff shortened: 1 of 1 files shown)


class-two-factor-core.php
```diff
@@ -897,6 +917,7 @@ public static function login_form_validate_2fa() {
 			$rememberme = true;
 		}
 
+		remove_filter( 'send_auth_cookies', '__return_false', PHP_INT_MAX );
```
dd32, line 920: It's worth noting that this is a just-in-case filter removal.
The filter should not normally be set as far as I can tell, but I figured it was better to remove it than potentially break an edge-case in an untested provider which ran the `authenticate` filter during it's validation process.

class-two-factor-core.php
```diff
@@ -480,6 +483,23 @@ public static function filter_authenticate( $user ) {
...
+	 * Prevent login cookies being set on login for Two Factor users.
+	 *
+	 * @param  WP_User|WP_Error $user Valid WP_User only if the previous filters
+	 *                                have verified and confirmed the
+	 *                                authentication credentials.
+	 *
+	 * @return WP_User|WP_Error
+	 */
+	public static function filter_authenticate_block_cookies( $user ) {
+		if ( $user instanceof WP_User && self::is_user_using_two_factor( $user->ID ) && did_action( 'login_init' ) ) {
```
dd32, line 496: Additionally, the `did_action( 'login_init' )` here is also for edge-case protection.

This allows for 3rd-party code authenticating the user through a non-wp-login.php page, yet using the `authenticate` filter to still call `wp_set_auth_cookie()`.

This is at odds with #385 but as long as this is noted for when/if that issue is resolved.

class-two-factor-core.php
iandunn: That sounds wise 👍🏻 

IMO it'd be good to copy/paste that into an inline comment, so folks who are curious in the future won't have to go looking through `git blame` for this PR.

class-two-factor-core.php
iandunn: Same opinion here about adding a comment.

class-two-factor-core.php
```diff
@@ -90,6 +90,9 @@ public static function add_hooks( $compat ) {
 		// Run only after the core wp_authenticate_username_password() check.
 		add_filter( 'authenticate', array( __CLASS__, 'filter_authenticate' ), 50 );
 
+		// Run as late as possible to disable sending authentication cookies.
```
iandunn, line 93:
```suggestion
		// Run as late as possible to prevent other plugins from unintentionally bypassing.
```

The "what" is obvious from the callback name and its docblock. I think the "why" would be more useful here.

class-two-factor-core.php
```diff
@@ -480,6 +483,23 @@ public static function filter_authenticate( $user ) {
 		return $user;
 	}
 
+	/**
+	 * Prevent login cookies being set on login for Two Factor users.
+	 *
```
iandunn, line 488:
```suggestion
	 *
	 * This makes it so that Core never sends the auth cookies. `login_form_validate_2fa()` will send them manually once the 2nd factor has been verified.
	 *
```

iandunn (approved):
Tested w/ these steps:

1. Checkout `master`
1. Add this code to an mu-plugin:
	```php
	add_filter( 'wp_login', function() {
		trigger_error( 'test', E_USER_ERROR );
	}, 9 );
	```
1. Log out, and then back in again with a user who has 2FA enabled. You should see the fatal error.
3. inspect cookies in browser, verify they were sent
4. Checkout this branch
5. Repeat steps and verify they weren't sent

LGTM 👍🏻

iandunn:
@dd32 is there anything left that's needed to merge this?

PR #514, opened 2023-02-09
Backup Codes: Always generate 10 codes via REST

iandunn:
The internal function accepts a param for flexibility, but currently there's no use case for letting the user choose how many they want.

This should be included in the same release as #504, since it was introduced there.

Props @xknown 
See https://github.com/WordPress/wporg-two-factor/issues/3


iandunn:
I'll fix the tests.

We should probably remove the `append` ability as well, I'll look into that.

dd32 (approved):

pkevan (approved):

PR #517, opened 2023-02-10
TOTP: Enforce single-use of TOTP one-time passwords.

dd32:
Per the [TOTP spec, section 5.2](https://www.rfc-editor.org/rfc/rfc6238#section-5.2):
>    Note that a prover may send the same OTP inside a given time-step
   window multiple times to a verifier.  The verifier MUST NOT accept
   the second attempt of the OTP after the successful validation has
   been issued for the first OTP, which ensures one-time only use of an
   OTP.

Once a TOTP token is entered, it should not be able to be re-used, it's one-time.
This is a hardening ticket, making the TOTP provider comply with the spec.

For example:
 - have two unique browser sessions open
 - generate a TOTP code, CODE1
 - Login with user/password/CODE1 on window 1
 - Login with user/password/CODE1 on window 2

Per spec, both windows logging in successfully shouldn't be allowed, window 2 should fail the TOTP code and be forced to wait for the next generated one.

For a an example of why this protection exists, If someone were to be looking over your shoulder as you type in your password and authcode, they should not be able to repeat the same on their own device and succeed at logging in, because you've already used your one-time code.

This PR takes it a step further, mostly out of convenience, by making any TOTP codes for previous-timestamp-ticks be invalid once you use the future code.

In this next scenario, both windows are logging in with unique TOTP codes, but with this change window 2 will be rejected as window 1 has already logged in with a newer one-time password. 
 - have two unique browser sessions open
 - generate a TOTP code, note it down as CODE1
 - generate the next TOTP code, note it down as CODE2
 - In window 1: Login with user/password/CODE2
 - In window 2: Login with user/password/CODE1

That means if someone were standing beside you, saw the first TOTP code (with 5s remaining on it), and you waited 
 and entered the next code generated, they should still not be able to login during the grace period with that now "old" code.

In this next scenario, both logins will proceed and login:
 - have two unique browser sessions open
 - generate a TOTP code, note it down as CODE1
 - generate the next TOTP code, note it down as CODE2
 - In window 1: Login with user/pass/CODE1
 - In window 2: Login with user/pass/CODE2
 
There is no minimum wait between successful logins, only that the next login must be using a later-generated code.

There are edge-cases which this does break, but I don't believe is a concern. For example, if there is a shared user account, with multiple TOTP generators, and their clocks are out of sync, and multiple people try to login at the same time, some will fail. 

Note: The login nonce prevents you having two windows open on the 2FA screen, you must start each test of the flow from the user/pass login screen.

See https://github.com/WordPress/wporg-two-factor/issues/3

### DIFF
--- a/providers/class-two-factor-totp.php
+++ b/providers/class-two-factor-totp.php
@@ -11,11 +11,18 @@
 class Two_Factor_Totp extends Two_Factor_Provider {
 
 	/**
-	 * The user meta token key.
+	 * The user meta key for the TOTP Secret key.
 	 *
 	 * @var string
 	 */
 	const SECRET_META_KEY = '_two_factor_totp_key';
+
+	/**
+	 * The user meta key for the last successful TOTP token timestamp logged in with.
+	 *
+	 * @var string
+	 */
+	const LAST_SUCCESSFUL_LOGIN_META_KEY = '_two_factor_totp_last_successful_login';
 
 	const DEFAULT_KEY_BIT_SIZE        = 160;
 	const DEFAULT_CRYPTO              = 'sha1';
@@ -408,6 +415,7 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 	 * @return boolean If the key was deleted successfully.
 	 */
 	public function delete_user_totp_key( $user_id ) {
+		delete_user_meta( $user_id, self::LAST_SUCCESSFUL_LOGIN_META_KEY );
 		return delete_user_meta( $user_id, self::SECRET_META_KEY );
 	}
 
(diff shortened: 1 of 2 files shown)


providers/class-two-factor-totp.php
```diff
@@ -434,29 +442,70 @@ public function is_valid_key( $key ) {
...
 	 *
 	 * @param string $key      The share secret key to use.
 	 * @param string $authcode The code to test.
 	 *
-	 * @return bool Whether the code is valid within the time frame
+	 * @return bool Whether the code is valid within the time frame.
 	 */
 	public static function is_valid_authcode( $key, $authcode ) {
+		return (bool) self::get_authcode_valid_ticktime( $key, $authcode );
+	}
```
dd32, line 498: Just noting that I chose to leave this method as `(bool)` and added a new method, badly named as `get_authcode_valid_ticktime()` instead which returns the timestamp of the validated code.

providers/class-two-factor-totp.php
```diff
@@ -434,29 +442,70 @@ public function is_valid_key( $key ) {
...
+		}
+
+		$last_totp_login = (int) get_user_meta( $user->ID, self::LAST_SUCCESSFUL_LOGIN_META_KEY, true );
+
+		// The TOTP authentication is not valid, if we've seen the same or newer code.
+		if ( $last_totp_login && $last_totp_login >= $valid_timestamp ) {
+			return false;
+		}
+
+		update_user_meta( $user->ID, self::LAST_SUCCESSFUL_LOGIN_META_KEY, $valid_timestamp );
```
dd32, line 482: It's been pointed out that this offers a race condition, but one that I can't see any real way of avoiding, as we don't have an atomic API in WordPress to verify that we're the ones that stored a specific value into the database.

iandunn (approved):
Code looks good and all of the scenarios tested well for me 👍🏻

PR #518, opened 2023-02-14
Be more tolerant of user input for auth codes

dd32:
Currently the TOTP (and backup codes / email codes, but that's less relevant) user input is tollerant of leading/trailing spaces, but doesn't accept TOTP codes that are entered how most apps display them: `123 456`.

This PR does several things:
 - Adds a placeholder value to the fields, so as to convey the format expected.
 - Makes it tolerant of whitespace in the middle of the number (`123 456`)
 - Removes `sanitize_text_field()` and uses `REQUEST` & `wp_unslash()` consistently between the providers. `sanitize_text_field()` has been removed as it strips characters from the provided code.. which shouldn't really be needed.. as invalid input is an invalid code.
 - Adds a unit test to cover the invalid characters / whitespace changes.

Examples using TOTP, the others are similar just with `78` affixed.
| Before | After | Invalid Input |
| --- | --- | --- |
| Screenshot 2023-02-14 at 1 36 24 pm <br> Screenshot 2023-02-14 at 1 36 31 pm | Screenshot 2023-02-14 at 1 35 59 pm | Screenshot 2023-02-14 at 1 36 06 pm |

After Feedback from Ian:
| Before | After |
| --- | --- |
| Screenshot 2023-02-14 at 1 36 24 pm | Screenshot 2023-02-15 at 11 40 24 am <br> Screenshot 2023-02-15 at 11 40 47 am<br> (Spaces in input highlighted) |
| Screenshot 2023-02-14 at 1 36 31 pm | Entering a character other than numbers and spaces (`[0-9 ]*`)<br> are stripped and do not "enter" into the field,<br>avoiding the "does not match expected format" error. |

Notably, it doesn't handle invalid input (non-numeric/space) well. But at least now the "Match the format requested" is more obvious based on the placeholder value presented.


iandunn (approved):
LGTM 👍🏻

providers/class-two-factor-backup-codes.php
```diff
@@ -326,11 +326,14 @@ public static function codes_remaining_for_user( $user ) {
 	 */
 	public function authentication_page( $user ) {
 		require_once ABSPATH . '/wp-admin/includes/template.php';
+
+		/* translators: Example auth code. */
+		$placeholder = sprintf( __( 'eg. %s', 'two-factor' ), '12345678' );
```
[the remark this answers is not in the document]
dd32: Standard American usage. Not in other Englishes :)  Although this is American english.. soo.. in my defence I copied directly from WordPress.com

Using the full `e.g.,` looks so ugly IMHO that no, just no :)

Instead, I'm going to merge this and the `123 456` example text info a different direction, removing the english-component of the placeholder, see the followup comments.

providers/class-two-factor-backup-codes.php
```diff
@@ -347,7 +350,12 @@ public function authentication_page( $user ) {
 	 * @return boolean
 	 */
 	public function validate_authentication( $user ) {
-		$backup_code = isset( $_POST['two-factor-backup-code'] ) ? sanitize_text_field( wp_unslash( $_POST['two-factor-backup-code'] ) ) : '';
+		if ( empty( $_REQUEST['two-factor-backup-code'] ) ) {
+			return false;
+		}
+
+		$backup_code = trim( str_replace( ' ', '', wp_unslash( $_REQUEST['two-factor-backup-code'] ) ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended, handled by the core method already.
```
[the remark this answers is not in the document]
dd32: I also don't feel strongly, but `trim()` already strips leading/trailing non-space whitespace so might as well just simplify to any whitespace.

providers/class-two-factor-backup-codes.php
```diff
@@ -326,11 +326,14 @@ public static function codes_remaining_for_user( $user ) {
...
 		require_once ABSPATH . '/wp-admin/includes/template.php';
+
+		/* translators: Example auth code. */
+		$placeholder = sprintf( __( 'eg. %s', 'two-factor' ), '12345678' );
 		?>
 		<p><?php esc_html_e( 'Enter a backup verification code.', 'two-factor' ); ?></p><br/>
 		<p>
 			<label for="authcode"><?php esc_html_e( 'Verification Code:', 'two-factor' ); ?></label>
-			<input type="tel" name="two-factor-backup-code" id="authcode" class="input" value="" size="20" pattern="[0-9]*" />
+			<input type="tel" name="two-factor-backup-code" id="authcode" class="input" value="" size="20" pattern="[0-9 ]*" placeholder="<?php echo esc_attr( $placeholder ); ?>" />
```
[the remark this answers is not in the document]
dd32: I've done that in 3ed9089, but added a JS handler to strip non-numeric/space characters, which removes the HTML5 validation error conditions.

See edits to PR description for images.

providers/class-two-factor-totp.php
```diff
@@ -314,7 +314,11 @@ public function user_two_factor_options( $user ) {
 				<input type="hidden" id="two-factor-totp-key" name="two-factor-totp-key" value="<?php echo esc_attr( $key ); ?>" />
 				<label for="two-factor-totp-authcode">
 					<?php esc_html_e( 'Authentication Code:', 'two-factor' ); ?>
-					<input type="tel" name="two-factor-totp-authcode" id="two-factor-totp-authcode" class="input" value="" size="20" pattern="[0-9]*" />
+					<?php
+						/* translators: Example auth code. */
+						$placeholder = sprintf( __( 'eg. %s', 'two-factor' ), '123456' );
```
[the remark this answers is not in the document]
dd32: I've taken a slightly different approach, which I'd like your review on @iandunn.

Instead of treating it like a string number, or groups of two digits, I've instead increased the width between all the characters, which both makes it easier to read, and due to the increased space, "feels like" (to me) that there's less of a requirement for spaces on the input (as it's auto-spaced).

Adding a space between the chunks like this is still possible, and ends up like this:
| Current | with extra space | user input |
| --- | --- | --- |
|  | Screenshot 2023-02-15 at 11 52 57 am | Screenshot 2023-02-15 at 11 55 46 am<br>(no user-input spaces)|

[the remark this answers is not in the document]
iandunn: 🤔 , I like this better than `e.g., 123 456`. I lean towards `123 456` being best, though. IMO blocks of numbers like in auth codes have two problems:

1) they're hard to _read_ because the numbers all run together
2) they're hard to _remember_ because the sequence is too long

Adding extra (uniform) space between each letter solves `1`, but not `2`. _Grouping_ the numbers into small blocks like `123 456` solves both problems IMO.

I don't feel strongly, though. I think `1 2 3 4 5 6` is an improvement on the current situation 👍🏻

providers/class-two-factor-backup-codes.php
```diff
@@ -347,7 +347,11 @@ public function authentication_page( $user ) {
 	 * @return boolean
 	 */
 	public function validate_authentication( $user ) {
-		$backup_code = isset( $_POST['two-factor-backup-code'] ) ? sanitize_text_field( wp_unslash( $_POST['two-factor-backup-code'] ) ) : '';
+		$backup_code = $this->sanitize_code_from_request( 'two-factor-backup-code' );
```
iandunn, line 350: Should this and email pass the `$length` param too?

dd32: Ideally yes, however, there isn't a defined class constant or property for those right now so I simply skipped over them in the initial code.

I was considering adding a `get_length()` method to the classes, which defaulted to 8 with a filter, to allow others to adjust the length of codes, but seemed out of the scope of this. I also have questions around how to handle things like Backup Codes if Two Factor (or another plugin) were to extend the length of those later on - what would happen to existing backup codes?

dd32:
> GitHub itself uses XXXXXX as the placeholder text

I've always been partial to the Stripe UI (and similar ones) that use dedicated boxes. Those that use this format however are less flexible of singular input boxes that just "accept a two factor challenge response" (ie. TOTP/Backup/Email/SMS/etc code)

Screenshot 2023-02-17 at 1 00 06 pm

providers/class-two-factor-backup-codes.php
iandunn: Yeah, those are good points. I think it's fine to leave as-is 👍🏻

class-two-factor-core.php
```diff
@@ -775,8 +775,48 @@ public static function login_html( $user, $login_nonce, $redirect_to, $error_msg
...
+							if ( ! spaceInserted && expectedLength && value.length === Math.floor( expectedLength / 2 ) ) {
+								value += ' ';
+								spaceInserted = true;
+							} else if ( spaceInserted && ! this.value ) {
+								spaceInserted = false;
+							}
+
+							this.value = value;
+
+							// Auto-submit if it's the expected length.
```
iandunn, line 808: I like the convenience of this when it works, but I also sometimes run into a problem where I make a typo on the last digit. When that happens the page refreshes and I have to start over.

I think ideally it'd use an XHR, so that the user can pick up where they left off if there's an error. That's probably more trouble than it's worth, though.

I'm not opposed to merging this, it's probably an improvement on average. Just noting the potential problems in case anyone has a good solution.

class-two-factor-core.php
```diff
@@ -785,12 +785,25 @@ public static function login_html( $user, $login_nonce, $redirect_to, $error_msg
...
 				if ( inputEl ) {
+					let spaceInserted = false;
 					inputEl.addEventListener(
 						'input',
 						function() {
-							this.value = this.value.replace( /[^0-9 ]/g, '' );
+							let value = this.value.replace( /[^0-9 ]/g, '' ).trimStart();
+
+							if ( ! spaceInserted && expectedLength && value.length === Math.floor( expectedLength / 2 ) ) {
+								value += ' ';
```
iandunn, line 800: This feels more complicated than it should be, but I don't see a good way to do it in CSS. Hopefully in the future we'll have [`nth-letter`](https://github.com/w3c/csswg-drafts/issues/3208) -- and  [`nth`-everything](https://css-tricks.com/a-call-for-nth-everything/), really 🤞🏻 

I did find [a workaround using `columns` and `word-wrap`](https://stackoverflow.com/a/24616034/450127), but I'm guessing it won't work inside an `<input>`. Maybe in a `<p contenteditable="true">`? That's also pretty hacky, though, and could cause other problems 🤷🏻‍♂️ 

The JS-based approach may be the most practical approach for now 👍🏻

dd32: I was also hoping for a `nth-letter`, and saw a bunch of suggestions for using `<p contenteditable>` but was hoping not to go there.

class-two-factor-core.php
dd32: That's fair, I'm happy to remove this and put it into a second PR, which I think I'll do.

class-two-factor-core.php
iandunn: Yeah, I think the current approach is fine 👍🏻

class-two-factor-core.php
iandunn: Either way 👍🏻

PR #522, opened 2023-02-15
Add unit tests for Two_Factor_Provider::get_code().

dd32:
This was split out of #518 as it's technically unrelated, and is failing for some reason in GitHub actions, but passing locally.

I can't really determine by looking at the code why it's failing though.

### DIFF
--- a/tests/providers/class-two-factor-provider.php
+++ b/tests/providers/class-two-factor-provider.php
@@ -1,0 +1,31 @@
+<?php
+/**
+ * Test Two Factor Provider
+ *
+ * @package Two_Factor
+ */
+
+/**
+ * Class Tests_Two_Factor_Provider
+ *
+ * @package Two_Factor
+ * @group providers
+ */
+class Tests_Two_Factor_Provider extends WP_UnitTestCase {
+	/**
+	 * @covers Two_Factor_Provider::get_code
+	 */
+	function test_get_code() {
+		$code = Two_Factor_Provider::get_code( 3, '1' );
+		$this->assertEquals( '111', $code );
+
+		$code = Two_Factor_Provider::get_code( 8, '1' );
+		$this->assertEquals( '11111111', $code );
+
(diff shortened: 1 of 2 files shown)


dd32:
This fails due to https://core.trac.wordpress.org/ticket/55194

Opened #523 to run unit tests against trunk as well.

dd32:
In making this testable, since `Two_Factor_Provider` is an abstract class, I chose to simply make it a static method due to it's lack of `$this->` usage. `$this->get_code()` is still acceptable, but now `Two_Factor_Provider::get_code()` also is.

tests/providers/class-two-factor-provider.php
```diff
@@ -0,0 +1,31 @@
...
+	 */
+	function test_get_code() {
+		$code = Two_Factor_Provider::get_code( 3, '1' );
+		$this->assertEquals( '111', $code );
+
+		$code = Two_Factor_Provider::get_code( 8, '1' );
+		$this->assertEquals( '11111111', $code );
+
+		$code = Two_Factor_Provider::get_code( 8, 'A' );
+		$this->assertEquals( 'AAAAAAAA', $code );
```
iandunn, line 26:
```suggestion
		$this->assertEquals( 'AAAAAAAA', $code );
		
		$code = Two_Factor_Provider::get_code( 30, array( 'A', 'B', 'C' ) );
		$this->assertSame( 1, preg_match( '/^[ABC]*$/', $code ) );
```

This tests the `$chars = implode( '', $chars );` statement.

iandunn (approved):
Looks like this is working after #524 👍🏻

tests/providers/class-two-factor-provider.php
dd32: Good call, I extended that to also validate that the generated code is 30char as requested, and an additional test that validates non-1-char-length strings.

iandunn: 👍🏻

PR #525, opened 2023-02-15
Tests: Generate an HTML coverage report when running tests locally.

iandunn:
This makes it convenient to check coverage locally, instead of having to go to Coveralls.

This does part of the documentation for #467 , but only the local part. Docs will still need to be added for Coveralls when #364 is done.


iandunn:
I'm having trouble getting the tests to run in Docker (even on `master`). If anyone has it working and can test that it still works with this PR, that'd be helpful. The tests passed in the GH action, though, so it should be fine.

iandunn:
I'm gonna go ahead and merge this, but LMK if anyone sees any problems in the future.

Issue #476, opened 2022-10-14
Notify user/admin when sensitive events occur


iandunn:
Most services w/ strong security standards will notify you when security-sensitive events happen. That (hopefully) makes you aware when an attacker has compromised something.

Here are some rough thoughts on events that should trigger a notification:

### User only
* when 2fa settings changed in any way

### Admin _and_ User
* when backup code used (should be filterable off for large sites)
* when user disables all 2fa providers _(maybe not necessarily if we add a mechanism to force 2fa on for certain roles - #185, #239, #255, #307, #451 )_

Should anything be added to that?

PR #526, opened 2023-02-17
Broadcast events for logging, email notifications, etc

iandunn:
This is a rough sketch of a potential way to provide details on 2FA events, so that they can be logged (#459), emailed (#476), added to Stream (https://github.com/xwp/stream/issues/1386), etc. 

This takes a more generic approach than #462, and doesn't do any logging/etc itself, it just publishes the events so that other code can subscribe to them and do whatever is desired. It also allows for more granular events, rather than only failures in the providers.

I'd like to get some early feedback before adding anything else. What do y'all think of this approach?

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -886,6 +886,7 @@ class Two_Factor_Core {
 		$login_nonce = get_user_meta( $user_id, self::USER_META_NONCE_KEY, true );
 
 		if ( ! $login_nonce || empty( $login_nonce['key'] ) || empty( $login_nonce['expiration'] ) ) {
+			self::broadcast( 'login_nonce_missing', compact( 'user_id' ) );
 			return false;
 		}
 
@@ -904,6 +905,8 @@ class Two_Factor_Core {
 
 		// Require a fresh nonce if verification fails.
 		self::delete_login_nonce( $user_id );
+
+		self::broadcast( 'login_nonce_mismatch', compact( 'user_id' ) );
 
 		return false;
 	}
@@ -974,7 +977,13 @@ class Two_Factor_Core {
 		 * @param bool     $rate_limited Whether the user login is rate limited.
 		 * @param WP_User $user          The user attempting to login.
 		 */
-		return apply_filters( 'two_factor_is_user_rate_limited', $rate_limited, $user );
+		$rate_limited = apply_filters( 'two_factor_is_user_rate_limited', $rate_limited, $user );
+
+		if ( $rate_limited ) {
+			self::broadcast( 'user_rate_limited', compact( 'user', 'rate_limit', 'last_failed' ) );
+		}
+
+		return $rate_limited;
 	}
 
 	/**
@@ -1006,6 +1015,7 @@ class Two_Factor_Core {
 			if ( isset( $providers[ $provider ] ) ) {
 				$provider = $providers[ $provider ];
 			} else {
+				self::broadcast( 'unavailable_provider', compact( 'user', 'provider' ) );
 				wp_die( esc_html__( 'Cheatin&#8217; uh?', 'two-factor' ), 403 );
 			}
 		} else {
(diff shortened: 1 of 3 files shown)


class-two-factor-core.php
```diff
@@ -1311,4 +1331,42 @@ public static function rememberme() {
...
+	 *
+	 * This doesn't handle the event, it simply broadcasts it so that plugins can hook in and log it, send email notifications, etc.
+	 *
+	 * @param string $event_code A unique name for the event.
+	 * @param array  $event_data Contains any additional context about the event.
+	 *                           ⚠️ DO NOT add any sensitive info, like tokens, passwords, etc. It could end up in
+	 *                           error logs, emails, etc, and the plugin handling the event might not have enough
+	 *                           context to be aware of it or redact it.
+	 *                           The only exception to that is a `WP_User` object, which will automatically be
+	 *                           reduced to the `user_login`.
```
dd32, line 1347:
```suggestion
	 *                           The only exception to that is a `WP_User` object in the 'user' key, which will
	 *                           automatically be reduced to the `user_login`.
```

class-two-factor-core.php
```diff
@@ -886,6 +886,7 @@ public static function verify_login_nonce( $user_id, $nonce ) {
 		$login_nonce = get_user_meta( $user_id, self::USER_META_NONCE_KEY, true );
 
 		if ( ! $login_nonce || empty( $login_nonce['key'] ) || empty( $login_nonce['expiration'] ) ) {
+			self::broadcast( 'login_nonce_missing', compact( 'user_id' ) );
```
dd32, line 889: I'm concerned that adding events for low-level validation is a bit over the top, yes, they could be used to pick up someone fuzzing the login forms, but they're not scenario's where a logger needs/should be aware of IMHO.

class-two-factor-core.php
```diff
@@ -905,6 +906,8 @@ public static function verify_login_nonce( $user_id, $nonce ) {
 		// Require a fresh nonce if verification fails.
 		self::delete_login_nonce( $user_id );
 
+		self::broadcast( 'login_nonce_mismatch', compact( 'user_id' ) );
```
dd32, line 909: same for this

class-two-factor-core.php
```diff
@@ -1240,6 +1250,7 @@ public static function enable_provider_for_user( $user_id, $new_provider ) {
 		$available_providers = self::get_providers();
 
 		if ( ! array_key_exists( $new_provider, $available_providers ) ) {
+			self::broadcast( 'invalid_provider', compact( 'user_id', 'new_provider' ) );
```
dd32, line 1253: and this

providers/class-two-factor-totp.php
```diff
@@ -445,6 +445,7 @@ public function is_valid_key( $key ) {
 	 */
 	public function validate_authentication( $user ) {
 		if ( empty( $_REQUEST['authcode'] ) ) {
+			Two_Factor_Core::broadcast( 'totp_missing_authcode', compact( 'user' ) );
```
dd32, line 448: and this

dd32 (commented):
I don't mind the approach myself, but as noted, I feel that exposing low-level protections of invalid input isn't useful, it only causes more logs of irrelevant details that hides actual important failures.

class-two-factor-core.php
iandunn: I was thinking it'd be useful to have a detailed audit trail when investigating a security incident, but at the same time I agree that we also don't wanna clutter everyday logs with standard pentesting.

_Broadcasting_ low-level events doesn't mean they'll necessarily be _used_, though. #459 might only log high-level stuff by default, but log everything if a user changes a filter. I imagine #476 would hardcode a very short list of events that it chooses to send emails about.

The specifics will vary based on the caller, and the site, though. So IMO it's best to leave that choice up to them. What do you think?

I don't feel strongly, though. Broadcasting high-level stuff would be better than nothing, and we can always add more in the future if we change our minds.

Issue #484, opened 2022-10-19
Reauth 2nd factor to change 2FA settings 


iandunn:
Most sites w/ strong 2FA require re-authorizing the 2nd factor in order to make any changes to 2FA settings. Without that, certain types of attacks could disable 2FA, add unauthorized keys, etc.

For convenience, there could be a ~5 minute time window when re-auth isn't required, similar to `sudo` in Unix-based systems.

Related #476

dd32:
In order to facilitate this, having a cookie set that contains the last-time that 2FA was processed would be beneficial.

The redirect code in #490 may also be beneficial here, as with some validation of said cookie, could require re-auth to continue.

PR #527, opened 2023-02-20
Switch to using `$_REQUEST` directly instead of `filter_input()` and remove `backup_2fa` 

dd32:
Currently there's two different routes into the 2FA credentials window, `action=validate_2fa` and `action=backup_2fa`, these are exactly the same, except that the former uses POST and the latter uses GET.

This is partially caused by the usage of `filter_input()` which has other problems for us, such as being unable to be unit tested.

This PR simplifies it by combining `action=validate_2fa` and `action=backup_2fa` and simplifying the resulting code.

This is part of an initial work towards #484.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -267,7 +267,11 @@ class Two_Factor_Core {
 	 * @return boolean
 	 */
 	public static function is_valid_user_action( $user_id, $action ) {
-		$request_nonce = isset( $_REQUEST[ self::USER_SETTINGS_ACTION_NONCE_QUERY_ARG ] ) ? sanitize_key( $_REQUEST[ self::USER_SETTINGS_ACTION_NONCE_QUERY_ARG ] ) : '';
+		$request_nonce = isset( $_REQUEST[ self::USER_SETTINGS_ACTION_NONCE_QUERY_ARG ] ) ? wp_unslash( $_REQUEST[ self::USER_SETTINGS_ACTION_NONCE_QUERY_ARG ] ) : '';
+
+		if ( ! $user_id || ! $action || ! $request_nonce ) {
+			return false;
+		}
 
 		return wp_verify_nonce(
 			$request_nonce,
@@ -300,10 +304,10 @@ class Two_Factor_Core {
 	 * @return void
 	 */
 	public static function trigger_user_settings_action() {
-		$action  = isset( $_REQUEST[ self::USER_SETTINGS_ACTION_QUERY_VAR ] ) ? sanitize_key( $_REQUEST[ self::USER_SETTINGS_ACTION_QUERY_VAR ] ) : '';
+		$action  = isset( $_REQUEST[ self::USER_SETTINGS_ACTION_QUERY_VAR ] ) ? wp_unslash( $_REQUEST[ self::USER_SETTINGS_ACTION_QUERY_VAR ] ) : '';
 		$user_id = self::current_user_being_edited();
 
-		if ( ! empty( $action ) && self::is_valid_user_action( $user_id, $action ) ) {
+		if ( self::is_valid_user_action( $user_id, $action ) ) {
 			/**
 			 * This action is triggered when a valid Two Factor settings
 			 * action is detected and it passes the nonce validation.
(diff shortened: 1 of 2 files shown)


lkraav:
> This is partially caused by the usage of `filter_intput()` which has other problems for us, such as being unable to be unit tested.

Interesting, I never knew `filter_input()` could have drawbacks (other than not able to do $_REQUEST).

It this ChatGPT reply about all there is to know?

> > Why is it difficult to unit test PHP filter_input()?

> Unit testing PHP's filter_input() function can be challenging because it is a function that relies on external input, and its behavior can be affected by the environment in which it is running. Here are some reasons why it can be difficult to unit test:

> External input dependency: filter_input() relies on external input from the user or from other sources, which can be difficult to simulate in a unit test environment. For example, if you want to test the behavior of filter_input() when it receives a specific input value, you need to find a way to simulate that input in your test code.

> Side effects: filter_input() can have side effects, such as changing the values of superglobal arrays such as $_GET, $_POST, and $_COOKIE. This can make it difficult to write isolated unit tests that don't interfere with the rest of the system.

> Environment dependency: The behavior of filter_input() can be affected by the environment in which it is running, such as the PHP version, the configuration settings, and the underlying operating system. This can make it difficult to write tests that are consistent across different environments.

> To address these challenges, it is often helpful to use a combination of unit tests and integration tests. Unit tests can focus on the behavior of filter_input() itself, while integration tests can test the interaction between filter_input() and other parts of the system, such as the database or other external services. It is also important to use test data that covers a range of input values and edge cases, to ensure that the function behaves correctly in a variety of scenarios.

dd32:
> Interesting, I never knew `filter_input()` could have drawbacks (other than not able to do $_REQUEST).

That response is mostly accurate, although there are some things that I don't believe are truthful (such as `filter_input` changing the superglobals).

The best description is IMHO that: filter_input() only works on the raw input to PHP, it does not use the php user-space super globals, it only uses the data passed to the php sapi process at the start of the request. That has some benefits, in that it enforces code not alter the source-of-truth for the request, but has the downside of making unit testing harder. `filter_var()` can replace `filter_input` except you loose the array-key-exists checks and have to do it yourself, which looses most of the advantages of it, for example:

```diff
-		$request_nonce = filter_input( INPUT_GET, self::USER_SETTINGS_ACTION_NONCE_QUERY_ARG, FILTER_CALLBACK, array( 'options' => 'sanitize_key' ) );
+		$request_nonce = isset( $_REQUEST[ self::USER_SETTINGS_ACTION_NONCE_QUERY_ARG ] ) ? filter_var( $_REQUEST[ self::USER_SETTINGS_ACTION_NONCE_QUERY_ARG ], FILTER_CALLBACK, array( 'options' => 'sanitize_key' ) ) : '';
```

Using the PHP7+ null-coalescing operator can make it similar though:
 ```diff
-		$request_nonce = filter_input( INPUT_GET, self::USER_SETTINGS_ACTION_NONCE_QUERY_ARG, FILTER_CALLBACK, array( 'options' => 'sanitize_key' ) );
+		$request_nonce = filter_var( $_REQUEST[ self::USER_SETTINGS_ACTION_NONCE_QUERY_ARG ] ?? '', FILTER_CALLBACK, array( 'options' => 'sanitize_key' ) );
```

In other words, while the `filter_*` API offers some benefits, in the space of a WordPress plugin it offers very little benefit where custom callbacks exist for validation/coercing and you need to mock requests for unit testing.

My experience of the filter extension is a bit biased, as back in the early PHP 5.x era where it was a new feature, it was very hard to rely upon due to it being disabled sometimes, and security vulnerabilities / bugs existed in older versions which although you may have called `FILTER_SANITIZE_URL` it may return different values depending upon which PHP version it was.

There's another reason for the change here though too - running `sanitize_key()` over the input just means that clients can send junk and the application interprets it as something valid and expected, which in something that's a security plugin is kind of not expected behaviour. User input sure, you should coerce user-input into the format they intended on supplying, but something such as a hard-coded `action` form value should be taken as what's sent, not a 'sanitized' version of it.

File: class-two-factor-core.php [hunk 1]
```diff
@@ -577,43 +580,6 @@ public static function show_two_factor_login( $user ) {
 		self::login_html( $user, $login_nonce['key'], $redirect_to );
 	}
 
-	/**
-	 * Display the Backup code 2fa screen.
-	 *
-	 * @since 0.1-dev
-	 */
-	public static function backup_2fa() {
```
iandunn, line 585: The secondary provider selection doesn't work for me with this gone. I tested with Backup Codes and Dummy available, and selecting the secondary logged me in without showing the provider's UI or validating any 2FA token.

iandunn (changes requested):
I think the `backup_2fa` stuff is necessary.

The `$_REQUEST` stuff looks good, though, and I agree that `filter_input()` adds minimal value, and makes testing harder in an environment like WP.

class-two-factor-core.php
dd32: Ah.. this is a oddity in the way the `Dummy` handler works, and something that I'd worked around in #529 via f6f5dfd0c524df5d7fc4230dfc1012cd309ab923 but failed to sync to this branch.

Basically, if you select the Dummy Auth as the provider, the auth is checked as part of the form being displayed, and well, dummy auth always returns truthful, it doesn't check to see if it's prompted and been confirmed (which IMHO, is a fault of the provider).

The reason why setting the Dummy provider as the primary provider doesn't have this problem, is that unlike the `validate_2fa` handler, there's _another_ handler that displays the form (but doesn't trigger the validation methods).

dd32: I've merged that into this branch with 87985df9b26936a4907a0f2e60ca33ff325025ed which I think is better than keeping the existing multiple entry points.

dd32: Okay, I was wrong.

The reason it was split was so that the handler for that action handler didn't attempt to process the auth. It duplicated most of the code, except for the code that actually performed validation.

Removing the `backup_2fa` method meant that it was processing auth on every request, which sounds fine, except that the Dummy method would always succeeded (and so then logged in), and it would cause other handlers to then instantly throw a `invalid code` error message.

In b1f6adb1e83490db4fcf7baafb4aee6f44734ea3 I've taken the alternative approach of making the singular action handle this scenario, by not processing the request if it's not an actual form submission. That removes the need for https://github.com/WordPress/two-factor/commit/f6f5dfd0c524df5d7fc4230dfc1012cd309ab923.

I'm not against restoring `backup_2fa` if you still think it's the best option.

iandunn: Nope, this sounds good, and it works well in my tests 👍🏻

iandunn (approved):

PR #528, opened 2023-02-20
Store the two-factor details in the user session at login time

dd32:
This PR is part of some larger work for #484.

This PR causes the two-factor timestamp and provider to be stored in the user session, which can be later used to validate that the user session is two-factor authenticated.

|Before|After|
|---|---|
| Screenshot 2023-02-20 at 2 45 59 pm | Screenshot 2023-02-20 at 2 45 46 pm |

Code Notes:
 - The 'attach_session_information' hook is poorly designed for this case, requiring the usage of an anonymous function so as to retrieve the details from the current-scope of the class.
 - It's possible not to use the callback (See the commit history where I've gone back and forth) by creating the session token yourself and passing that into `wp_set_auth_cookie()`, but that requires the caller to duplicate significantly more code than this option.
 - Note: The screenshots were taken with 0749b4adc0f483bd2c44e05d63f783786281b59c which places the extra array fields at the end, with dbee8261096062aa08cf49299eabb642f8ffdb7c the fields are placed at the start of the array. This has no meaningful effect.
 - Unit testing this should be possible, but has not been included at time of PR creation.

To test, you can use something like `$session = WP_Session_Tokens::get_instance( $user->ID )->get( wp_get_session_token() );` on an admin page to inspect the current sessions session, this is what's used in the above screenshots.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1089,9 +1089,9 @@ class Two_Factor_Core {
 	 * @return int|false The last time the two-factor was validated on success, false if not currently using a 2FA session.
 	 */
 	public static function is_current_user_session_two_factor() {
-		$user    = wp_get_current_user();
-		$manager = WP_Session_Tokens::get_instance( $user->ID );
+		$user_id = get_current_user_id();
 		$token   = wp_get_session_token();
+		$manager = WP_Session_Tokens::get_instance( $user_id );
 		$session = $manager->get( $token );
 
 		if ( empty( $session['two-factor-login'] ) ) {
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1234,7 +1234,7 @@ class Two_Factor_Core {
 		$session_information_callback = function( $session, $user_id ) use( $provider, $user ) {
 			if ( $user->ID === $user_id ) {
 				$session['two-factor-login']    = time();
-				$session['two-factor-provider'] = get_class( $provider );
+				$session['two-factor-provider'] = $provider->get_key();
 			}
 
 			return $session;
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1082,6 +1082,26 @@ class Two_Factor_Core {
 	}
 
 	/**
+	 * Determine if the current user session is logged in with 2FA.
+	 *
+	 * @since 0.8.0
+	 *
+	 * @return int|false The last time the two-factor was validated on success, false if not currently using a 2FA session.
+	 */
+	public static function is_current_user_session_two_factor() {
+		$user    = wp_get_current_user();
+		$manager = WP_Session_Tokens::get_instance( $user->ID );
+		$token   = wp_get_session_token();
+		$session = $manager->get( $token );
+
+		if ( empty( $session['two-factor-login'] ) ) {
+			return false;
+		}
+
+		return (int) $session['two-factor-login'];
+	}
+
+	/**
 	 * Login form validation.
 	 *
 	 * @since 0.1-dev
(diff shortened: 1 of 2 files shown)


dd32:
> Unit testing this should be possible, but has not been included at time of PR creation.

Unfortunately full unit testing isn't possible unless #527 lands.

I've added a unit test to verify that this new getter returns false for a user session set outside of the two-factor auth flow though.

dd32:
> > Unit testing this should be possible, but has not been included at time of PR creation.
> 
> Unfortunately full unit testing isn't possible unless #527 lands.

I've rebased this PR after I merged #527, and tests are now included that covers the login flow - validating that a cookie set outside of the login flow doesn't get flagged as 2fa, and that a login via the 2fa flow sets the appropriate session flags.

This unfortunately did require changing how the `login_form_validate_2fa` handler works, splitting it in two - the part that does the login, and the part that does the `exit`. Not ideal, but does now allow for `validate_2fa` to be unit testable,

class-two-factor-core.php
```diff
@@ -1007,11 +1007,44 @@ public static function is_user_rate_limited( $user ) {
...
+	}
+
+	/**
+	 * Login form validation.
+	 *
+	 * This function exists for unit testing, as `exit` prevents testing.
+	 *
+	 * @since x.x
+	 */
+	public static function _login_form_validate_2fa() {
```
iandunn, line 1047: The early `return [void]` statements below might need to be `return false` now.

tests/class-two-factor-core.php
```diff
@@ -734,4 +766,85 @@ public function test_show_password_reset_error() {
...
+		$this->assertArrayNotHasKey( 'two-factor-provider', $session );
+	}
+
+	/**
+	 * Validate that a simulated 2fa login sets the session two-factor data.
+	 *
+	 * @covers Two_Factor_Core::is_current_user_session_two_factor()
+	 * @covers Two_Factor_Core::login_form_validate_2fa()
+	 */
+	public function test_is_current_user_session_two_factor_with_two_factor() {
```
iandunn, line 809: This only covers about 40% of the function, but a lot of it is "controller" code that we wouldn't test w/ PHPUnit anyway. The suggestion above to break it into a controller/model would make it easier to see what "model" code isn't being tested. That may be more trouble than it's worth, though.

iandunn (commented):
I ran out of time today, but wanted to send what I have so far. I'll pick it back up tomorrow.

class-two-factor-core.php
dd32: Doesn't really matter since we only really care about the `true` return values, but `return bool` is more explicit than `return null|true`

class-two-factor-core.php
```diff
@@ -1028,7 +1061,7 @@ public static function login_form_validate_2fa() {
 
 		if ( true !== self::verify_login_nonce( $user->ID, $nonce ) ) {
 			wp_safe_redirect( home_url() );
```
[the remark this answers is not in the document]
dd32: That's a fair statement, but the existence of this function as it is is really a hack, and not something I really want to change in _this_ PR.

This function exists purely for unit testing purposes, and due to how the plugin is architectured, there's very little room for splitting Model and Controller without breaking compatibility.

I'd be happy to explore splitting it up in a new PR, where the only functional changes are splitting it up and making it work, but the pure fact that we don't have any unit tests for this flow at all currently means the risk of breakage is already pretty high.

Alternatively, Reverting 042fa12062bb7a8a4aacef4f82c32a03ade61115 and all unit tests is probably a better route forward here.

tests/class-two-factor-core.php
dd32: This plugin doesn't follow any such design, so kind of have to make do with what we have, and this is that.

iandunn: I didn't mean to imply some kind of formal thing, just basically going farther with what you've already done. But I agree it's not worth trying to do a big refactor like that in this PR 👍🏻

class-two-factor-core.php
```diff
@@ -1198,15 +1231,29 @@ public static function login_form_validate_2fa() {
 			$rememberme = true;
 		}
 
+		$session_information_callback = function( $session, $user_id ) use( $provider, $user ) {
+			if ( $user->ID === $user_id ) {
+				$session['two-factor-login']    = time();
+				$session['two-factor-provider'] = get_class( $provider );
```
dd32, line 1237: Per #546 

```suggestion
				$session['two-factor-provider'] = $provider->get_key();
```

PR #529, opened 2023-02-20
Revalidate two factor settings prior to allowing any two-factor changes to an account.

### LINKED ISSUE #484: Reauth 2nd factor to change 2FA settings

dd32:
## What?

This PR requires that the user "revalidate" their two-factor details prior to changing their two-factor settings.
The grace-time from last-2fa-validation to changes is 10 minutes (Twice this for the save, so you can start editing at minute 9, and save at 18 minutes).

## Why?

It's generally accepted that a user should not be able to change their 2FA settings without validating that the user currently has their 2FA token.
For example, consider an account left logged in on a public computer (or left unattended in an office), a malicious user should not be able to add a new key or create new backup codes.

## How?

This works by building off #528 to track when the user last validated their 2FA tokens, and preventing updates unless the session is 2fa'd recently.

## Testing Instructions

1. Login
2. Validate you can change 2FA settings
3. Wait 15 minutes (Or change the grace time via `two_factor_revalidate_time` to a lower time)
4. Validate you can't change 2FA settings after the grace-time is expired
5. Revalidate your 2FA, check you can change 2FA settings again.

## Changelog Entry

Added Security - Validate the user has access to their 2FA keys prior to changing 2FA details.

## Tickets

Fixes #484.

#527 and #528 have been split out of this, but are included in the git commit history of this branch.


class-two-factor-core.php
```diff
@@ -1433,11 +1612,63 @@ public static function user_two_factor_options( $user ) {
...
+					'<br><a class="button" href="%s">' . __( 'Revalidate now', 'two-factor' ) . '</a>',
+					esc_url( $url )
+				)
+			);
+		}
+
+		printf(
+			'<fieldset id="two-factor-options" %s>',
+			$show_2fa_options ? '' : 'disabled="disabled"',
+		);
```
dd32, line 1685: I'm not too fond of this UI, but without recreating the entire UI this seems to be the simplest way forward.

I had planned on using the "interim login" feature to display in a thickbox-style overlay, but the core re-authentiacte-your-session implementation isn't extendable and would need to be duplicated into the plugin. Likewise, the interim-login code only has support for the Customizer, not for other JS clients.

class-two-factor-core.php
```diff
@@ -1230,6 +1289,144 @@ public static function login_form_validate_2fa() {
...
+			?>
+			</div>
+			<?php
+			/** This action is documented in wp-login.php */
+			do_action( 'login_footer' );
+			?>
+			</body></html>
+			<?php
+			exit;
+		}
```
dd32, line 1364: I'm not sure if this interim-login functionality is useful to keep in here or not, as while it provides a better in-line UI for revalidating, it doesn't provide any way of reloading the parent page / changing the state of the parent pages form.

class-two-factor-core.php
```diff
@@ -1433,11 +1631,63 @@ public static function user_two_factor_options( $user ) {
...
+					'<div class="notice notice-warning inline"><p>%s</p></div>',
+					sprintf(
+						'You are currently logged in with a session from %s. <br>' .
+							'<strong>You are NOT using a Two Factor session.</strong>',
+						'<code>' . date( 'r', $session['login'] ) . '</code>'
+					)
+				);
+			}
+		}
+		// DEBUG END
```
dd32, line 1663: This is for testing only, to make it clearer of what the current login session details are.

dd32:
Examples of this PR:

| Example | |
| --- | --- |
| Just logged in | Screenshot 2023-03-27 at 1 09 31 pm |
| Past 2FA grace period | Screenshot 2023-03-27 at 1 46 15 pm |
| revalidate flow | <video src="https://user-images.githubusercontent.com/767313/227836343-38705527-1762-49d8-b2a8-f2c0add63953.mov"> |

The UI is "ugly" in that it's just an alert box that sends you off to a wp-login page, it's not a pretty inline thickbox iframe or anything.
This fulfils the basic requirements of the issue, but obviously needs UX, design, and other design work done on it. Before that can start though, I think it needs some technical agreement on the direction.

StevenDufresne:
Based on my testing, provided the page is not refreshed, actions can still be carried out, is that correct?

dd32:
@StevenDufresne I'm not sure I follow, you'll need to explain it better.

StevenDufresne:
Here's what I did.

1. Sign up, visit `wp-admin/profile.php`
2. Wait (15 mins although i updated this locally to 1min)
3. Click "Generate Validation Codes"
4. Codes are generated
5. Reload the page
6. See "revalidate" notice.

So if a user leaves their profile page open, a third party could modify data provided they don't reload the page?

dd32:
> So if a user leaves their profile page open, a third party could modify data provided they don't reload the page?

Ahh, gotcha.

I suspect this is that you were still within the grace-period, The notice shows up at `$time` but you can save forms for `$time * 2`.

`* 2` is possibly too long, and maybe it should be simply `+ MINUTE_IN_SECONDS`, or `+ 2 * MINUTE_IN_SECONDS`.
https://github.com/WordPress/two-factor/pull/529/files#diff-ee04f1d323104504c6bfa38dd11ef43e78d1dbac2883293af0b5c310d16e9519R1133-R1143

The reasoning for that is that you should be able to view the form at `$expiry - 1 second` and save the form. `$time * 2` is something that WordPress has in a few places as the grace between generation and save.

iandunn:
> I suspect this is that you were still within the grace-period

I tested this a bit and that's what it seems like to me too.

class-two-factor-core.php
```diff
@@ -1144,86 +1222,29 @@ public static function _login_form_validate_2fa( $user, $nonce = '', $provider =
...
-				self::show_password_reset_error();
-				return;
-			}
-
-			$login_nonce = self::create_login_nonce( $user->ID );
-			if ( ! $login_nonce ) {
-				wp_die( esc_html__( 'Failed to create a login nonce.', 'two-factor' ) );
-			}
-
-			self::login_html( $user, $login_nonce['key'], $redirect_to, esc_html__( 'ERROR: Invalid verification code.', 'two-factor' ), $provider );
```
iandunn, line 1226: It feels good to modularize all this into a pure function, even discounting the ability to reuse it 👍🏻

iandunn (approved):
This looks great, kudos!

Increasing the test coverage would be nice, but not necessarily a blocker.

providers/class-two-factor-backup-codes.php
```diff
@@ -56,7 +56,7 @@ public function register_rest_routes() {
 				'methods'             => WP_REST_Server::CREATABLE,
 				'callback'            => array( $this, 'rest_generate_codes' ),
 				'permission_callback' => function( $request ) {
-					return current_user_can( 'edit_user', $request['user_id'] );
+					return Two_Factor_Core::rest_api_can_edit_user_and_update_two_factor_options( $request['user_id'] );
```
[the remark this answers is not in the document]
dd32: That's a very good point, and something I was considering here.

For example, should 2FA confirmation be needed to change your own password? Seems like that would be a reasonable suggestion! but right now with how this is implemented that's not really possible.

But on the other hand, we can't really override the capability checks, because they include very little context of "what" the capability is being checked for. Is it being checked to see if the user edit screen is being displayed? or is it being checked to see if the user can be edited?

I was thinking about this, and one option would be to capture all `edit_user` capability checks on the REST API between `rest_request_before_callbacks` and `rest_dispatch_request`, but that's *only* rest-api checks, it doesn't cover any non-api-based methods - which is why I ended up with this callback.

class-two-factor-core.php
```diff
@@ -99,6 +99,8 @@ public static function add_hooks( $compat ) {
 		add_filter( 'wp_login_errors', array( __CLASS__, 'maybe_show_reset_password_notice' ) );
 		add_action( 'after_password_reset', array( __CLASS__, 'clear_password_reset_notice' ) );
 		add_action( 'login_form_validate_2fa', array( __CLASS__, 'login_form_validate_2fa' ) );
+		add_action( 'login_form_revalidate_2fa', array( __CLASS__, 'login_form_revalidate_2fa' ) );
```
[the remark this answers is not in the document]
dd32: Perhaps the feature flag for that could simply be the revalidate time, if it's falsey disable the feature.

Part of my hesitation in merging this as-is is the danger that no followup UI work happens to make this scenario any better.

I think it would be well agreed that the current Two-Factor UI is not optimal and not at all the most user friendly, but as we know, creating a new UI and supporting existing provider is not going to be the easiest task.

dd32:
Changes in the commits above that are notable since reviews:
 - Disabling revalidation by setting a false/0 time: [6502d24](https://github.com/WordPress/two-factor/pull/529/commits/6502d2458a4cf0d85b254c113abc5311803f0560)
 - When initially setting 2FA, mark the session as 2FA so that the users current session can continue to edit their 2FA settings. [6721a41](https://github.com/WordPress/two-factor/pull/529/commits/6721a414c500305c4a7c9e3df9893f4b68fe0dce) Without this, if you login with 2FA, setup 2FA, and save, the next page load prompted for 2FA revalidation, annoying if you hadn't properly setup your 2FA settings yet.
 - Don't return true for logged out requests to `Two_Factor_Core::current_user_can_update_two_factor_options()`. [d3e024c](https://github.com/WordPress/two-factor/pull/529/commits/d3e024cd1b2e48ac2994f3e99c798c29f90b3587) This should **really** be used in addition to a `current_user_can()` check, but better to include a check here.

ravinderk (approved):

class-two-factor-core.php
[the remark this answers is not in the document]
iandunn: Yeah, I agree 👍🏻 I don't feel strongly about a flag.

PR #530, opened 2023-02-24
Document the TOTP Filters, add Issuer filter

dd32:
In https://github.com/WordPress/wporg-two-factor/issues/69 it's been realised that the TOTP plugin has a `two_factor_totp_title` filter for the label, but doesn't have one for the `issuer`.

The label is supposed to be in the format of `Issuer: $user` but others may incorrectly only change the prefix of the label without also changing the issuer details.

This PR does two things:
 - Documents the existing filters
 - Adds a `two_factor_totp_issuer` filter to change the Site name that's then used as part of the label.


iandunn (approved):
Looks good 👍🏻

PR #481, opened 2022-10-17
Add Security Policy

iandunn:
Fixes #328

PR #541, opened 2023-03-16
Add information about where to submit security bugs

felixarntz:
## What?
This adds an FAQ entry for where to file security bugs.

## Why?
Per request from the WordPress security team.

## How?
The copy used is similar to the one that is already present here in the Gutenberg `SECURITY.md` file, which was used as a reference.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Reporting Security Issues
+
+The plugin contributors and WordPress community take security bugs seriously. We appreciate your efforts to responsibly disclose your findings, and will make every effort to acknowledge your contributions.
+
+To report a security issue, please visit the [WordPress HackerOne](https://hackerone.com/wordpress) program.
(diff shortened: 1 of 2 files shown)


peterwilsoncc (approved):
Thanks Felix, much appreciated.

iandunn:
Thanks! #481 also added a `SECURITY.md`, but in `.github/` instead of `/`:

https://github.com/WordPress/two-factor/blob/c0eae28ce5c83d28bdf810dc940f33febe38a0b7/SECURITY.md

https://github.com/WordPress/two-factor/blob/c0eae28ce5c83d28bdf810dc940f33febe38a0b7/.github/SECURITY.md

It sounds like [GH supports both locations](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository). I personally prefer `.github` to reduce clutter, but if all our other repos are using `/` then the consistency is probably more important.

Any other opinions on which one we should delete?

felixarntz:
@iandunn Apologies, I wasn't ware of that.

I think having `SECURITY.md` in the root helps visibility. I have to admit, I didn't even know that GitHub actually _does_ something with these files, and I have never paid attention to see that there's a Security tab on those repos that uses the security policy from that file. 🤦

That may just be me, but maybe other people too. That's why I would personally argue it should be in the root, for better visibility. Furthermore it's common to have other "similar" files like `CONTRIBUTING.md` at the root level too.

Issue #540, opened 2023-03-16
Test against WordPress 6.2


jeffpaul:
### Is your enhancement related to a problem? Please describe.

With 6.2 due out later this month and already in the RC phase, it would be great to test that `master` branch against WP 6.2 RC2 to see if we're clear to bump the "tested up to" value to 6.2.

### Proposed Solution

n/a

### Designs

n/a

### Describe alternatives you've considered

n/a

### Please confirm that you have searched existing issues in this repository.

Yes

Issue #533, opened 2023-03-07
Release version 0.8.0


jeffpaul:
## Release instructions

- [x] Branch: Starting from `master`, create a branch named `release/0.8.0` for the release-related changes.
- [x] Version bump: Bump the version number in `readme.txt` and `two-factor.php` if it does not already reflect the version being released.  Update both the plugin "Version:" header value and the plugin `TWO_FACTOR_VERSION` constant in `two-factor.php`.
- [ ] Changelog: Add/update the changelog in `readme.txt`.
- [x] New files: Check to be sure any new files/paths that are unnecessary in the production version are included in [.distignore](https://github.com/WordPress/two-factor/blob/master/.distignore).
- [x] Readme updates: Make any other readme changes as necessary. `readme.md` is geared toward GitHub and `readme.txt` contains WordPress.org-specific content. The two are slightly different.
- [x] Create Release PR: Push any local changes in `release/0.8.0` to origin, create a release PR, and request review to ensure all CI checks pass and ensure master branch changes are limited to merges only.
- [ ] Merge: After review approval, merge the release pull request (or make a non-fast-forward merge from your release branch to `master`).  `master` contains the latest stable release.
- [ ] Release: Create a [new release](https://github.com/WordPress/two-factor/releases/new), naming the tag and the release with the new version number, and targeting the `master` branch.  Paste the changelog from `readme.txt` into the body of the release and include a link to the [closed items on the milestone](https://github.com/WordPress/two-factor/milestone/23?closed=1).
- [ ] SVN: Wait for the [GitHub Action: Deploy](https://github.com/WordPress/two-factor/actions/workflows/deploy.yml) to finish deploying to the WordPress.org repository.  If all goes well, users with SVN commit access for that plugin will receive an emailed diff of changes.
- [ ] Check WordPress.org: Ensure that the changes are live on https://wordpress.org/plugins/two-factor/. This may take a few minutes.
- [ ] Close the milestone: Edit the [milestone](https://github.com/10up/simple-local-avatars/milestone/23) with the release date (in the `Due date (optional)` field) and link to the GitHub release (in the `Description` field), then close the milestone.
- [ ] Punt incomplete items: If any open issues or PRs which were milestoned for `0.8.0` do not make it into the release, update their milestone to `0.9.0` or `Future Release`.

jeffpaul:
[...]

PR #543, opened 2023-03-20
Update deploy.yml

jeffpaul:
## What?
Updates the [current deploy.yml file](https://github.com/WordPress/two-factor/blob/c0eae28ce5c83d28bdf810dc940f33febe38a0b7/.github/workflows/deploy.yml) by swapping in https://github.com/10up/action-wordpress-plugin-deploy for the steps to deploy to SVN.

## Why?
Deploys to SVN are disabled as they're currently failing, see: https://github.com/WordPress/two-factor/blob/c0eae28ce5c83d28bdf810dc940f33febe38a0b7/.github/workflows/deploy.yml#L44.

## How?
Replaces the previous `Authenticate with WordPress.org SVN` and `Deploy to WordPress.org SVN` steps with `WordPress Plugin Deploy` and `Upload release asset` steps utilizing the https://github.com/10up/action-wordpress-plugin-deploy GitHub Action.

Note that this also ONLY deploys on published releases and no longer on ANY merge to `master` which means that only newly released versions will be deployed to SVN and any individual merges to `master` on GitHub will no longer be deployed to `trunk` on SVN.

## Testing Instructions
This is a tough one, pretty much need to YOLO into a published release and 🤞🏼 that the Action processes correctly (e.g., that `SVN_USERNAME`, `SVN_PASSWORD`, and `GITHUB_TOKEN` are accurate).

## Screenshots or screencast
n/a

## Changelog Entry

> Changed - Deploy workflow uses `10up/action-wordpress-plugin-deploy` instead of a manual connection to SVN.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -29,6 +29,8 @@ jobs:
       - name: Install NPM dependencies
         run: npm install
 
+      - name: Build plugin
+        run: npm run build
       - name: WordPress Plugin Deploy
         id: deploy
         uses: 10up/action-wordpress-plugin-deploy@stable
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -37,6 +37,7 @@ jobs:
         with:
           generate-zip: true
         env:
+          BUILD_DIR: dist
           SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
           SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,10 +1,7 @@
 name: Deploy
 
-# Run deploy only on tag and master builds.
+# Run deploy only on publised releases.
 on:
-  push:
-    branches:
-      - master
   release:
     types: [published]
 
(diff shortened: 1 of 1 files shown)


jeffpaul:
Open to whether we try this with 0.8.0 or figure out post-release.

iandunn (approved):
I don't have a strong opinion, but this sounds good to me 👍🏻

.github/workflows/deploy.yml
```diff
@@ -32,15 +29,21 @@ jobs:
       - name: Install NPM dependencies
         run: npm install
```
dd32, line 31:
```suggestion
      - name: Install NPM dependencies
        run: npm install

      - name: Build plugin
        run: npm run build
```

.github/workflows/deploy.yml
```diff
@@ -32,15 +29,21 @@ jobs:
...
-      - name: Authenticate with WordPress.org SVN
-        run: svn info --non-interactive --username "${{ secrets.SVN_USERNAME }}" --password "${{ secrets.SVN_PASSWORD }}" https://plugins.svn.wordpress.org/two-factor/
+      - name: WordPress Plugin Deploy
+        id: deploy
+        uses: 10up/action-wordpress-plugin-deploy@stable
+        with:
+          generate-zip: true
+        env:
+          SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
+          SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
```
dd32, line 39:
```suggestion
        env:
          BUILD_DIR: dist
          SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
          SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
```

dd32 (commented):
I have no strong opinion here. I  can confirm all the secret constants are correct.

However, I'm 99% certain this won't work as-is, see suggestions:
 - `npm run build` needs to happen after `npm install` to actually build the plugin (npm run deploy did this previously)
 -  `BUILD_DIR: dist` needs to be added to the `env`

I can't test this, as `npm run build` doesn't work on my environments,

.github/workflows/deploy.yml
kasparsd: Yes, this need to be built before it can be pushed up.

kasparsd (commented):
This looks good!

The build process is being tested as part of this step in the other CI file:

https://github.com/WordPress/two-factor/blob/a9b54920ae73fe460681c85596eb1be10fe89aa9/.github/workflows/test.yml#L47-L48

and it is running correctly:

https://github.com/WordPress/two-factor/actions/runs/4471733733/jobs/7856996785?pr=543#step:12:17

As long as the deploy action picks up the files at `dist/*` (after implementing the suggestion from @dd32), it should be all good!

kasparsd:
> I can't test this, as npm run build doesn't work on my environments

@dd32 Do you know why that is? Should we create an issue to fix that? Technically all of the [build logic](https://github.com/WordPress/two-factor/blob/3d350aebcd6b0dedad37cc9b98b59d7b6d8e75dd/Gruntfile.js) should be handled by Node with no external system dependencies. Would moving everything to `wp-scripts` solve the issue?

dd32:
> > I can't test this, as npm run build doesn't work on my environments
> 
> dd32 Do you know why that is? Should we create an issue to fix

And suddenly now when I went to copy the error, everything works. Magic. (I haven't tested this PR, but visual review looks like it'll work to me)

Sorry for the false alarm.

jeffpaul:
@kasparsd if you're 👍🏼 on this, feel free to merge and then proceed with the steps on #533 to complete the 0.8.0 release (happy to handle those myself if you're good on this PR to ensure we're not blocked on the release deploy to SVN).

Noting here that ideally we'd sneak #540 into 0.8.0 but I do not view that as a blocker.

PR #545, opened 2023-03-21
Allow TOTP keys to be filterable

pkevan:
## What?

Allow TOTP keys to be filtered on save/retrieval.

## Why?
Since the discussion in https://github.com/WordPress/two-factor/pull/389 seems to be on-going, the following should allow users of the plugin to decide how they would like to approach the encryption part.

## How?
Adds filters on save and retrieval.

## Testing Instructions
Test adding filters, and making sure they are saved.

## Changelog Entry
Added - filters to TOTP keys.

### DIFF
--- a/providers/class-two-factor-totp.php
+++ b/providers/class-two-factor-totp.php
@@ -420,7 +420,8 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 	 * @return string
 	 */
 	public function get_user_totp_key( $user_id ) {
-		return (string) get_user_meta( $user_id, self::SECRET_META_KEY, true );
+		$meta = (string) get_user_meta( $user_id, self::SECRET_META_KEY, true );
+		return apply_filters( 'two_factor_totp_get_user_totp_key', $meta, $user_id );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


dd32 (commented):
While I'm not against this change, I don't personally see the direct need for it.

I understand the intention of why, but it seems to me that this isn't something that needs to be filtered except in rare circumstances, and is working around other limitations.

A plugin can add a second provider, which overloads those methods and saves them elsewhere.  What is not be filterable is being able to tell two-factor that `For the provider "Two_Factor_Totp" use the classname "CUSTOMPLUGIN_Factor_Totp"` as the plugin is hard-coded with the notion that provider_name === php_classname, so I'd probably be saying the filter needs to be on `Two_Factor_Core:: get_providers()`'s resultset or in it's loop that creates the provider objects.

_(Created #546 for this, which I don't think necessarily supersedes this, but rather, offers a much more flexible future approach for dealing with individual provider customizations)_

These filters as-is are even a bit limiting, the `set_user_totp_key` filter can't be used to save it outside of user_meta for example.

There's also the generic metadata filters, which admittedly probably shouldn't be used to filter this value if you're using it for encryption purposes.

pkevan:
Closing in favour of the more flexible approach in #546.

PR #546, opened 2023-03-22
Add a filter to filter the classname used for a provider

dd32:
## What?

This PR allows for the classname of a provider to differ from it's "key" that's used within the application.
This was spurred by an attempt to simply extend a core provider with a custom one to override certain functions (instead of #545 ).

This has resulted in several changes:
 - Filter addition
 - Removing variables like 'class' and replacing them with 'provider_key'
 - Addition of a `Two_Factor_Provider::get_key()` which returns the providers 'key', which defaults to the class name as present.
 - removing `get_instance()` from each individual provider and moving it to the abstract `Two_Factor_Provider`, each provider had a slightly different one, and the usage of `__CLASS__` meant children of the provider had to override the `get_instance()` method as well.

## Why?

This allows for users of the plugin to overload a provider with a custom variant of the provider, without having to register an entirely new provider and migrate users over to it.

For example:
```php

add_filter( 'two_factor_provider_classname_Two_Factor_Totp', function( $provider ) {
	return 'Secured_Two_Factor_Totp';
} );

// Stores TOTP keys securely rot13'd.
class Secured_Two_Factor_Totp extends Two_Factor_Totp {
	// Pretend to be the Two_Factor_Totp class in the UI.
	public function get_key() {
		return 'Two_Factor_Totp';
	}

	// When saving the key, ensure it's rot13'd
	public function set_user_totp_key( $user_id, $key ) {
		$key = str_rot13( $key );
		return parent::set_user_totp_key( $user_id, $key );
	}

	// When fetching the key, rot13 it back.
	public function get_user_totp_key( $user_id ) {
		return str_rot13( parent::get_user_totp_key( $user_id ) );
	}
}

```

## How?

By allowing the classname to be filtered, external code can overload core providers with customisations without needing to duplicate the entire provider.

## Testing Instructions

 - Unit tests: `npm run test`
 - Add the above code into an mu-plugin, verify that you can setup TOTP and login, but once code is removed, the key is invalid.

## Changelog Entry

Added - Better compatibility to allow overloading core Two Factor providers without duplicating the entire provider.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -411,9 +411,9 @@ class Two_Factor_Core {
 		$enabled_providers    = self::get_enabled_providers_for_user( $user );
 		$configured_providers = array();
 
-		foreach ( $providers as $classname => $provider ) {
-			if ( in_array( $classname, $enabled_providers, true ) && $provider->is_available_for_user( $user ) ) {
-				$configured_providers[ $classname ] = $provider;
+		foreach ( $providers as $provider_name => $provider ) {
+			if ( in_array( $provider_name, $enabled_providers, true ) && $provider->is_available_for_user( $user ) ) {
+				$configured_providers[ $provider_name ] = $provider;
 			}
 		}
 
@@ -712,10 +712,10 @@ class Two_Factor_Core {
 			$provider = call_user_func( array( $provider, 'get_instance' ) );
 		}
 
-		$provider_class = get_class( $provider );
+		$provider_name = $provider->get_name();
 
 		$available_providers = self::get_available_providers_for_user( $user );
-		$backup_providers    = array_diff_key( $available_providers, array( $provider_class => null ) );
+		$backup_providers    = array_diff_key( $available_providers, array( $provider_name => null ) );
 		$interim_login       = isset( $_REQUEST['interim-login'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 
 		$rememberme = intval( self::rememberme() );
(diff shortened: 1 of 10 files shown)


class-two-factor-core.php
```diff
@@ -179,17 +179,25 @@ public static function get_providers() {
...
+		foreach ( $providers as $provider_key => $path ) {
 			include_once $path;
 
+			/**
+			 * Filters the classname for a provider.
+			 *
+			 * @param string $provider_key The provider name, which is the default for the Classname.
+			 * @param string $path         The provided provider path to be included.
+			 */
+			$class = apply_filters( 'two_factor_provider_classname', $provider_key, $path );
```
dd32, line 191:
```suggestion
			$class = apply_filters( "two_factor_provider_classname_{$provider}", $provider_key, $path );
```

Perhaps it would be worthwhile not having a generic filter, but one that's specific to the provider only.

pkevan: seems neater and less error-prone to have specific filters for each provider.

dd32: Yeah, that's my thought.

If someone _needs to_ filter all of them, they can fetch the list of keys from the `two_factor_providers` filter and go from there.

dd32: Added in 51cfc3a7cb5a1f2a2d5892a11be157e9ab95f98a

pkevan (approved):
Code looks good - is there an example we can test where someone has extended the providers? 

My only concern with the changes to class instantiation might conflict with someone who has previously set this up.

dd32:
> Code looks good - is there an example we can test where someone has extended the providers?

The best example is the unit test included, and the code included in the PR description.

> My only concern with the changes to class instantiation might conflict with someone who has previously set this up.

That's a valid concern, however, it's pretty awkward to extend them at present to the point that this _shouldn't_ break any existing implementations of that. They'd have to be including their own replacement class in the `two_factor_providers` array, and as it's currently keyed by the classname, couldn't "transparently replace" a core provider without modifying code, AFAIK.

I also can't find any plugins in the directory that extend the core providers: https://wpdirectory.net/search/01GX87VBAWAPCCA0RXEY3HY7G0

PR #547, opened 2023-03-27
Fatal Error Fix: Remove unnecessary comma that causes a site-wide fatal server error on PHP 7.2

KZeni:
I'm going to be a bit brief on these details as this is a hot issue needing a hot fix.

I've started a support forum topic on this at https://wordpress.org/support/topic/fatal-error-php-7-2-syntax-error-breaking-whole-site-3/ for further discussion/reference/etc. as well.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -666,7 +666,7 @@ class Two_Factor_Core {
 		if ( ! $attempted_user ) {
 			return $errors;
 		}
-		
+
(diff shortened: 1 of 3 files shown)


KZeni:
Created an issue at https://github.com/WordPress/two-factor/issues/548 just in case that's helpful for reference & as a more complete reporting of things (WP.org support forum + GH issue + GH PR.)

iandunn:
@KZeni `0.8.1`. is now available on wordpress.org. Please let me know if you run into any issues installing it or after.

PR #566, opened 2023-05-12
Unset session data when disabling all providers to prevent fatal

### LINKED ISSUE #565: Revalidation required when no providers enabled
### Describe the bug

If you disable all providers, you're still required to revalidate. You can't enable any providers until you do, but you get a fatal error when you try.

<p>
Screenshot 2023-05-12 at 3 08 07 PM
</p>

> Uncaught Error: Call to a member function `get_key()` on null in `two-factor/class-two-factor-core.php` on line `742`

Introduced in #529. I missed this during testing, but @adamwoodnz discovered it in https://github.com/WordPress/wporg-two-factor/issues/160. His PR https://github.com/WordPress/wporg-two-factor/pull/161/ would fix downstream, but it looks like the root cause is located here.

### Steps to Reproduce

1. logout
1. delete all `_two_factor...` usermeta for the user
1. add this to an to muplugin to make testing easier
	```php
	add_filter( 'two_factor_revalidate_time', function() {
		return 30;
	} );
	```
1. login, then go to `wp-admin/profile.php`
1. wait 30 seconds, and refresh. the `Revalidate` button will not appear, because you don't have a 2fa session. that's the correct behavior
1. now enable the Dummy provider, and immediately disable it (before the 30 second revalidation period expires)
1. wait 30 seconds for the period to expire, then reload `profile.php` again. The `Revalidate` button will appear, because the WP session data still indicates that it's a 2FA session. that's wrong, and prevents you from enabling any provider until you revalidate. if you try to revalidate, you get the fatal error.

iandunn:
Fixes #565

Unsets the 2FA session data when all providers are disabled, so that the user's current session is no longer considered a 2FA session. That avoids the fatal error, and allows them to enable a provider again.

## Testing Instructions

Repeat the reproduction steps from #565.

## Changelog Entry

None needed, because #529 has not been released yet.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1828,7 +1828,9 @@ class Two_Factor_Core {
 				update_user_meta( $user_id, self::PROVIDER_USER_META_KEY, $new_provider );
 			}
 
-			self::update_current_session( $enabled_providers );
+			if ( $user_id === get_current_user_id() ) {
+				self::update_current_session( $enabled_providers );
+			}
 		}
 	}
 
@@ -1854,8 +1856,7 @@ class Two_Factor_Core {
 				// does it matter that they were existing? is it enough to just say "if there are some, then..."
 				// test still pass w/ this commented out
 			$enabled_providers &&
-			! self::is_current_user_session_two_factor() &&
-			$user_id === get_current_user_id()
+			! self::is_current_user_session_two_factor()
 		) {
 			$session['two-factor-provider'] = ''; // Set the key, but not the provider, as no provider has been used yet.
 			$session['two-factor-login']    = time();
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1830,23 +1830,23 @@ class Two_Factor_Core {
 				update_user_meta( $user_id, self::PROVIDER_USER_META_KEY, $new_provider );
 			}
 
-			// Have we enabled new providers? Set this as a 2FA session, so they can continue to edit.
-			if (
-				! $existing_providers &&
-				$enabled_providers &&
-				! self::is_current_user_session_two_factor() &&
-				$user_id === get_current_user_id()
-			) {
-				$token = wp_get_session_token();
-				if ( $token ) {
-					$manager = WP_Session_Tokens::get_instance( $user_id );
-					$session = $manager->get( $token );
+			$token = wp_get_session_token();
+			if ( $token ) {
+				$manager = WP_Session_Tokens::get_instance( $user_id );
+				$session = $manager->get( $token );
 
+				// Have we enabled new providers? Set this as a 2FA session, so they can continue to edit.
+				if (
+					! $existing_providers &&
+					$enabled_providers &&
+					! self::is_current_user_session_two_factor() &&
+					$user_id === get_current_user_id()
+				) {
 					$session['two-factor-provider'] = ''; // Set the key, but not the provider, as no provider has been used yet.
 					$session['two-factor-login']    = time();
-
-					$manager->update( $token, $session );
 				}
+
+				$manager->update( $token, $session );
 			}
 		}
 	}
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1100,6 +1100,11 @@ class Two_Factor_Core {
 		$manager = WP_Session_Tokens::get_instance( $user_id );
 		$session = $manager->get( $token );
 
+		// this works, but leaves the raw session data still indicating that it's a 2fa session. that feels wrong.
+		// if ( ! self::get_available_providers_for_user( $user_id ) ) {
+		// 	return false;
+		// }
+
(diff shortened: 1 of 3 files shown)


iandunn:
This is working in my tests for wp-admin, but not yet for 3rd party clients that use the REST API endpoints.

Maybe we should provide a new endpoint to update the session, and it'd be their responsibility to call it? It seems like an internal detail that should be handled automatically, though 🤔

dd32:
What about changing `is_current_user_session_two_factor()` to be false when the user has no 2FA providers active?

class-two-factor-core.php
```diff
@@ -1825,25 +1828,47 @@ public static function user_two_factor_options_update( $user_id ) {
...
+		}
+	}
+
+	/**
+	 * Update the user session.
+	 */
+	public static function update_session( $user_id, $enabled_providers ) {
+		// maybe rename to something more specific to updating the providers? see if this is only used when enabling/disabling, or also used when logging in, revalidating, etc
+
+		$token = wp_get_session_token();
```
dd32, line 1841: Passing $user_id here but using `wp_get_session_token()` doesn't make much sense - as $user_id is either get_current_user_id() or the function should return instantly.

class-two-factor-core.php
```diff
@@ -1825,25 +1828,47 @@ public static function user_two_factor_options_update( $user_id ) {
...
+		}
+
+		$manager = WP_Session_Tokens::get_instance( $user_id );
+		$session = $manager->get( $token );
+
+		// Have we enabled new providers? Set this as a 2FA session, so they can continue to edit.
+		if (
+			// ! $existing_providers &&
+				// does it matter that they were existing? is it enough to just say "if there are some, then..."
+				// test still pass w/ this commented out
```
dd32, line 1854: The main intention would not to extend their 2fa grace time by simply constantly enabling and disabling an additional provider I think

providers/class-two-factor-totp.php
```diff
@@ -150,6 +150,11 @@ public function rest_delete_totp( $request ) {
 		$user    = get_user_by( 'id', $user_id );
 
 		$this->delete_user_totp_key( $user_id );
+		// this doesn't call user_two_factor_options_update()
+		// maybe the custom ui should do that? well i guess it cant
+		// it's probably appropriate for this func to do that
+
+		// same for other rest api endpoints that update providers
```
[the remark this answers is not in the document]
dd32: Perhaps converting the XHR endpoints to REST API methods was poorly thought out, and instead it should've been something more generic, which ran through a centralised Two_Factor_Core handler that decided what the request did, removing the providers from the equation.

```
# Details about 2FA
GET /me/two-factor
{ enabled: [ two_factor_dummy ], available: { two_factor_email: [ param1, param2 ] } }

# Setup a 2FA
POST /me/two-factor
{ provider: two_factor_totp, { param1: ..., ... } }

# Reset / Deactivate
DELETE /me/two-factor
{ provider: two_factor_dummy }
```

tests/class-two-factor-core.php
```diff
@@ -862,6 +862,44 @@ public function test_enabling_two_factor_is_factored_session() {
...
+		Two_Factor_Core::user_two_factor_options_update( $user->ID );
+
+		$this->assertFalse( Two_Factor_Core::is_current_user_session_two_factor() );
+
+		$manager = WP_Session_Tokens::get_instance( $user->ID );
+		$token   = wp_get_session_token();
+		$session = $manager->get( $token );
+
+		$this->assertArrayNotHasKey( 'two-factor-login', $session );
+		$this->assertArrayNotHasKey( 'two-factor-provider', $session );
```
dd32, line 900: I guess at the time I thought something like "The fact the user is no longer logged in with 2FA doesn't suddenly make their session not a 2fa session", but somewhere along the line with the logic changes I've missed that they couldn't then update the session after disabling it..

dd32 (commented):
I've taken this in a slightly different direction via #567 but adding my thoughts inline here.

iandunn:
> What about changing is_current_user_session_two_factor() to be `false` when the user has no 2FA providers active?

That way [my first approach](https://github.com/WordPress/two-factor/pull/566/commits/d520895e965e501f2414c230d44fb42ccb6c3003), but it felt like a smell to leave the the raw session data in an incorrect state. I worried that it could lead to confusion and maybe even bugs in the future.

> ...alter `current_user_can_update_two_factor_options()` [to remove `&& ! $is_two_factor_session`]

That seems fine to me regardless 👍🏻

class-two-factor-core.php
iandunn: That's a good point 👍🏻  Changed in 4439b92

iandunn: ...and 94bf899

iandunn:
Fixed by https://github.com/WordPress/two-factor/pull/567

PR #574, opened 2023-05-30
Sync two-factor session meta to newly created sessions

### LINKED ISSUE #573: is_current_user_session_two_factor() returns false after changing password
### Describe the bug

During the password change process, the current session is terminated, and a new session begins. While I don't agree with that, it's how WordPress 4.0~6.2 appear to work.

The result of this is that the current sessions metadata is discarded, and the current session becomes a non-2fa session. This also means that the user must revalidate to change settings (even if they just logged in).

This is partially caused by Two-Factor only using the `attach_session_information` hook during the 2fa checks:
https://github.com/WordPress/two-factor/blob/654ebb75024c21c116b4155e7254e143c9f64413/class-two-factor-core.php#L1291-L1302

Possibly, the hook should be used outside of the login process with a check to sync over metadata from the current session to the newly generated session if the user is the same.

See https://core.trac.wordpress.org/ticket/58427
See https://github.com/WordPress/wporg-two-factor/issues/191

Per the upstream core ticket, IMHO this is a core bug, but one that Two-Factor probably has to work around.

### Steps to Reproduce

1. Login with 2FA
2. Change password
3. Observe 2FA revalidation

Uploading Screen Recording 2023-05-30 at 2.29.01 pm.mov…

https://github.com/WordPress/two-factor/assets/767313/a9ac3603-2dda-4270-b8d8-6e427b61857e

### Environment information

 - WordPress 6.2.2

Yes

Yes

dd32:
## What?

Copies two-factor session information from the current session to newly created sessions.

Fixes #573

## Why?

WordPress creates a new session when the user changes their password, for no particular reason.

See https://core.trac.wordpress.org/ticket/58427
See https://github.com/WordPress/wporg-two-factor/issues/191

## Testing Instructions

See screen recording. The debug shown is from this change:
```diff
diff --git a/class-two-factor-core.php b/class-two-factor-core.php
index 53fa43b..ce5b197 100644
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1718,6 +1718,11 @@ class Two_Factor_Core {
                        $show_2fa_options ? '' : 'disabled="disabled"',
                );

+       echo '<pre>'; var_dump( [
+               'token' => wp_get_session_token(),
+               'session' => Two_Factor_Core::get_current_user_session(),
+       ] ); echo '</pre>';
+
                wp_nonce_field( 'user_two_factor_options', '_nonce_user_two_factor_options', false );
                ?>
                <input type="hidden" name="<?php echo esc_attr( self::ENABLED_PROVIDERS_USER_META_KEY ); ?>[]" value="<?php /* Dummy input so $_POST value is passed when no providers are enabled. */ ?>" />
```

## Screenshots or screencast

See the Issue for the before.

Note that in the recording the Session token changes, but the metadata remains.

https://github.com/WordPress/two-factor/assets/767313/34042555-b896-4de6-87a4-7aef7522da83

## Changelog Entry

N/A - Fix for an unreleased change: #528

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1945,9 +1945,11 @@ class Two_Factor_Core {
 		}
 
 		$current_session = self::get_current_user_session();
-		foreach ( $current_session as $key => $value ) {
-			if ( str_starts_with( $key, 'two-factor-' ) ) {
-				$session[ $key ] = $value;
+		if ( $current_session ) {
+			foreach ( $current_session as $key => $value ) {
+				if ( str_starts_with( $key, 'two-factor-' ) ) {
+					$session[ $key ] = $value;
+				}
 			}
 		}
 
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -124,6 +124,8 @@ class Two_Factor_Core {
 		// Run as late as possible to prevent other plugins from unintentionally bypassing.
 		add_filter( 'authenticate', array( __CLASS__, 'filter_authenticate_block_cookies' ), PHP_INT_MAX );
 
+		add_filter( 'attach_session_information', array( __CLASS__, 'filter_session_information' ), 10, 2 );
+		
 		add_action( 'admin_init', array( __CLASS__, 'trigger_user_settings_action' ) );
 		add_filter( 'two_factor_providers', array( __CLASS__, 'enable_dummy_method_for_debug' ) );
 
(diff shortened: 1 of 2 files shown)


class-two-factor-core.php
```diff
@@ -1925,4 +1927,30 @@ public static function rememberme() {
...
+	 *
+	 * This is required as WordPress creates a new session when the user password is changed.
+	 *
+	 * @see https://core.trac.wordpress.org/ticket/58427
+	 *
+	 * @param array $session The Session information.
+	 * @param int   $user_id The User ID for the session.
+	 * @return array
+	 */
+	public static function filter_session_information( $session, $user_id ) {
```
iandunn, line 1942: A unit test for this would be ideal if it's not gonna be a pain to setup.

iandunn (approved):
Code looks good and it works in my tests 👍🏻

class-two-factor-core.php
dd32: added in 52c1c0d

PR #587, opened 2023-08-31
Add method to disable an individual provider

### LINKED ISSUE #585: Implement Disable Provider method
Two_Factor_Core provides a [method](https://github.com/WordPress/two-factor/blob/master/class-two-factor-core.php#L1796) `enable_provider_for_user`.

To compliment this, a method should be added `disable_provider_for_user`, which takes similar arguments ( `$user_id`, `$provider` ) and also returns a boolean.

iandunn:
## What?

Adds a `Two_Factor_Core::disable_provider_for_user()`  method.

## Why?

This provides parity with the `Two_Factor_Core::enable_provider_for_user()` method. It's not currently used by Two Factor itself, but it allows plugins to disable an individual provider directly, rather than having to know the internals and update meta keys, etc.

One example use case for that is https://github.com/WordPress/wporg-two-factor/pull/223#pullrequestreview-1570860929, where a custom front-end UI was developed in React, using REST API endpoints.

Fixes #585 

## Testing Instructions

1. Enable a provider via `wp-admin/profile.php` and save
2. Add the following code in an mu-plugin. Replace the name of the provider with one you enabled.
	add_action( 'init', function() {
		Two_Factor_Core::disable_provider_for_user( get_current_user_id(), 'Two_Factor_Dummy' );
	} );
3. Refresh `profile.php` and see that the provider has been disabled. If another provider was enabled, it should now be set as the primary provider.

## Changelog Entry

> Added - New `Two_Factor_Core::disable_provider_for_user()`  method provides parity with the `Two_Factor_Core::enable_provider_for_user()` method. It allows plugins to disable an individual provider directly, rather than having to know the internals and update meta keys, etc.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -125,7 +125,7 @@ class Two_Factor_Core {
 		add_filter( 'authenticate', array( __CLASS__, 'filter_authenticate_block_cookies' ), PHP_INT_MAX );
 
 		add_filter( 'attach_session_information', array( __CLASS__, 'filter_session_information' ), 10, 2 );
-		
+
 		add_action( 'admin_init', array( __CLASS__, 'trigger_user_settings_action' ) );
 		add_filter( 'two_factor_providers', array( __CLASS__, 'enable_dummy_method_for_debug' ) );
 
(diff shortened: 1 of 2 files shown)


iandunn:
🤔 There shouldn't be any ways for a malicious user to pass arbitrary values. This function isn't called by anything in Two Factor right now, it only exists to make things easier for plugins.

[The companion function](https://github.com/WordPress/two-factor/blob/5fe5af3ef873155dfe34592a41deb3fdae471786/class-two-factor-core.php#L1796) is used in Two Factor, but in all of those cases it's called from a REST API handler, which have `permission_callback` functions defined ([example](https://github.com/WordPress/two-factor/blob/5fe5af3ef873155dfe34592a41deb3fdae471786/providers/class-two-factor-backup-codes.php#L57-L59)).

I added a comment in [e84e061](https://github.com/WordPress/two-factor/pull/587/commits/e84e061849d8feffc4441649178a1c61bd6d30a1) for clarity.

pkevan (approved):

PR #597, opened 2024-04-23
Update readme.txt

bph:
## What?
Bump test up to 

## Why?
to remove the not tested banner on the public repository 

## How?
change the number


kasparsd (approved):

PR #605, opened 2024-04-25
Ensure PHP 5.6+ support during CI to avoid breaking changes

kasparsd:
## What?

- Enable PHP compatibility linter that ensures all plugin code is compatible with PHP 5.6 or later. That's until we bump the required version of PHP.
- Bump development specific dependencies to the latest versions.

## Why?

This now would catch issues like #604:

trailing-comma

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,16 +28,16 @@
     "php": ">=5.6"
   },
   "require-dev": {
-    "automattic/vipwpcs": "^2.3",
+    "automattic/vipwpcs": "^3.0",
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7.2",
     "php-coveralls/php-coveralls": "^2.5",
     "phpcompatibility/php-compatibility": "10.x-dev as 9.99.99",
-    "phpcompatibility/phpcompatibility-wp": "dev-master",
+    "phpcompatibility/phpcompatibility-wp": "^2.1",
     "phpunit/phpunit": "^9.5",
     "roots/wordpress-core-installer": "^1.100",
     "roots/wordpress-full": "~6.0",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -28,23 +28,23 @@
...
-    "yoast/phpunit-polyfills": "^1.0"
+    "wp-coding-standards/wpcs": "^3.1",
+    "wp-phpunit/wp-phpunit": "~6.0",
+    "yoast/phpunit-polyfills": "^2.0"
   },
   "scripts": {
     "lint": "phpcs",
-    "lint-php8": "phpcs -p --standard=PHPCompatibilityWP --runtime-set testVersion 8.0- --extensions=php --ignore='vendor/,wordpress/,node_modules/' .",
-    "test": "php -d xdebug.mode=coverage vendor/bin/phpunit",
+    "lint-compat": "phpcs -p --standard=PHPCompatibilityWP --runtime-set testVersion 5.6- --extensions=php --ignore='tests/,includes/Yubico/,vendor/,wordpress/,node_modules/' .",
```
kasparsd, line 46: Skipping checks on U2F helpers since those will be removed soon and they currently report the following warnings for PHP 8.4:

```
FILE: includes/Yubico/U2F.php
--------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
--------------------------------------------------------------------------------------------------------------------------------
 504 | WARNING | Implicitly marking a parameter as nullable is deprecated since PHP 8.4. Update the type to be explicitly
     |         | nullable instead. Found implicitly nullable parameter: $previous.
--------------------------------------------------------------------------------------------------------------------------------

Time: 1.88 secs; Memory: 24MB
```

PR #608, opened 2024-04-30
Rely on just-in-time translation loading

swissspidy:
## What?

Removes the manual `load_plugin_textdomain()` call to rely on WordPress to load translations at the right time.

Bumps the requirement to WordPress 4.6 accordingly.

Also, defines the `Requires *` headers in the main plugin file so that WordPress can actually read them and prevent you from installing the plugin if requirements are not met.

The headers mustn't really be set in the readme.

## Why?

The plugin was loading translations too early on `plugins_loaded`. Plugins should wait until `init` to ensure the current user has been set up, so that the current user's locale can be respected.

That said, why manually load translations when WordPress can do it for you :-)

## Testing Instructions

Switch to a different language and see that the translations are still loaded as expected.

## Changelog Entry

> Fixed - Ensured translations are only loaded just-in-time.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -93,7 +93,6 @@ class Two_Factor_Core {
 	 * @since 0.1-dev
 	 */
 	public static function add_hooks( $compat ) {
-		add_action( 'plugins_loaded', array( __CLASS__, 'load_textdomain' ) );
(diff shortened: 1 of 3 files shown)


kasparsd (approved):

PR #612, opened 2024-06-12
Update short description

jeffpaul:
## What?
This PR shortens the short description to be within the WP.org 150 character requirement.

Prior short description: Enable Two-Factor Authentication using time-based one-time passwords (OTP, Google Authenticator), Universal 2nd Factor (FIDO U2F, YubiKey), email, and backup verification codes.

New short description: Enable Two-Factor Authentication using time-based one-time passwords, Universal 2nd Factor (FIDO U2F, YubiKey), email, and backup verification codes.

## Why?
WP.org truncates the Short Description if its longer than 150 characters, we don't want that truncated, so this PR shortens it to under 150 characters.

## How?
Changes the short description in readme.txt as well as the roughly matching description headed in two-factor.php.

## Testing Instructions
Use an [Online Character Count Tool](https://www.charactercountonline.com/) to count the characters in the new, short description to see that its <=150 characters.

## Changelog Entry

> Changed - Shortened the Short Description to be under 150 characters per WordPress.org requirements.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -6,7 +6,7 @@ Stable tag:   0.9.1
 License:      GPL-2.0-or-later
 License URI:  https://spdx.org/licenses/GPL-2.0-or-later.html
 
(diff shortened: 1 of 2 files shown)


kasparsd (approved):
LGTM

PR #615, opened 2024-07-11
Add/filter for support links

StevenDufresne:
## What?
Sometimes users have difficulty with their two-factor settings. In the "Having Problems?" section, we currently don't have a way to add links that are not providers. 

This PR adds a filter called `two_factor_login_support_links` to allow plugins to provide non-provider support links.

## Why?
We want to give users access to documentation that can help them with using 2fa or with recovering their account.

For example, it would allow us to do

```
Having Problems?
- Use your authenticator app
- Use a recovery code
- Recover your account
```

## How?
It adds a filter inside the `<ul>` so HTML can be passed in like so:

```
function my_custom_login_support_links( $content ) {
    return '<li><a href="{href}">Recover your account</a></li>';
}
add_filter( 'two_factor_login_support_links', 'my_custom_login_support_links' )

```

## Testing Instructions
1. Add the filter mentioned above
2. With 2fa configured, try logging in
3. Expect to see the new link in the list.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -835,6 +835,7 @@ class Two_Factor_Core {
 							</a>
 						</li>
 					<?php endforeach; ?>
+					<?php echo apply_filters( 'two_factor_login_support_links', '' ); ?>
 				</ul>
 			</div>
 		<?php endif; ?>
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -835,7 +835,12 @@ class Two_Factor_Core {
 							</a>
 						</li>
 					<?php endforeach; ?>
-					<?php echo apply_filters( 'two_factor_login_support_links', '' ); ?>
+					<?php
+					/*
+					* Allow plugins to add links to the two-factor login form.
+					*/
+					echo apply_filters( 'two_factor_login_support_links', '' );
+					?>
 				</ul>
 			</div>
 		<?php endif; ?>


class-two-factor-core.php
```diff
@@ -825,17 +825,29 @@ public static function login_html( $user, $login_nonce, $redirect_to, $error_msg
...
+
+					/*
+					* Allow plugins to add links to the two-factor login form.
+					*/
+					$links = apply_filters( 'two_factor_login_support_links', $links );
+
+					// Echo out the filtered links
+					foreach ( $links as $link ) {
+						echo wp_kses_post( $link );
+					}
```
dd32, line 848:
```suggestion
					/*
					* Allow plugins to filter the backup method links on the login form.
					*/
					$links = apply_filters( 'two_factor_login_backup_links', $links );

					// Echo out the filtered links
					echo implode( '', $links );
```

Given this filter only runs when there are backup providers, calling it a support links might be confusing for some, we can probably name it something like `two_factor_login_backup_links`?

There's probably an argument that this should be built prior to the `if ( $backup_providers )` conditional to output filtered data even when there exist no backup methods for the given user?

I don't think we need to explicitly sanitize the HTML here either, it's either hard-coded HTML or from a PHP function that can do anything anyway.

class-two-factor-core.php
```diff
@@ -825,17 +825,29 @@ public static function login_html( $user, $login_nonce, $redirect_to, $error_msg
...
+
+					/*
+					* Allow plugins to add links to the two-factor login form.
+					*/
+					$links = apply_filters( 'two_factor_login_support_links', $links );
+
+					// Echo out the filtered links
+					foreach ( $links as $link ) {
+						echo wp_kses_post( $link );
```
kasparsd, line 847: **suggestion:** Considering that all of these would be links, could we make the filter run on an array of link hrefs and labels?

```php
$backup_links = [
    'https://example.com' => 'Label',
    // ...
];
```

dd32: I thought about that, but I can see that some might want to customise the attributes, or not output a link at all.

A generic filter is probably better for that reason.

kasparsd (approved):
@StevenDufresne Was there anything else you wanted to include here or is this ready for merge?

If you don't get around to it, we'll extend the inline docblock for the filter to include documentation for the arguments ([like here](https://github.com/WordPress/two-factor/blob/6a95e7ffdebb2e0408b4ccb298e9df16d5b6998c/class-two-factor-core.php#L204-L210)) and also add it [to the relevant readme section](https://github.com/WordPress/two-factor/blob/6a95e7ffdebb2e0408b4ccb298e9df16d5b6998c/readme.txt#L23-L30).

class-two-factor-core.php
dd32: These points remain:

> Given this filter only runs when there are backup providers, calling it a support links might be confusing for some, we can probably name it something like two_factor_login_backup_links?

OR 

> There's probably an argument that this should be built prior to the if ( $backup_providers ) conditional to output filtered data even when there exist no backup methods for the given user?

I suspect the intention of the filter was actually to be used when the user has no backup provider, or, at least that others who would have a use for it would be in that group.

two-factor.php
```diff
@@ -11,7 +11,7 @@
  * Plugin Name:       Two Factor
  * Plugin URI:        https://wordpress.org/plugins/two-factor/
  * Description:       Enable Two-Factor Authentication using time-based one-time passwords, Universal 2nd Factor (FIDO U2F, YubiKey), email, and backup verification codes.
- * Version:           0.9.1
+ * Version:           0.9.2
```
georgestephanis: I'm hesitant on any PR doing a plugin version bump by itself, generally it'll be batched with others.


```suggestion
 * Version:           0.9.1
```

masteradhoc:
Thanks @georgestephanis for the review. 

@StevenDufresne Sorry for the long waiting time. Would you be able to check the comments and provide an updated PR? We'd be happy to merge this asap :)

class-two-factor-core.php
```diff
@@ -1065,37 +1065,58 @@ public static function login_html( $user, $login_nonce, $redirect_to, $error_msg
...
-				$backup_link_args['rememberme'] = $rememberme;
-			}
-			if ( $redirect_to ) {
-				$backup_link_args['redirect_to'] = $redirect_to;
-			}
-			if ( $interim_login ) {
-				$backup_link_args['interim-login'] = 1;
+			$links = [];
+
+			if ( $backup_providers )  {
```
masteradhoc:
```suggestion
			if ( $backup_providers ) {
```

remove unneeded doublespace

class-two-factor-core.php
```diff
@@ -1065,37 +1065,58 @@ public static function login_html( $user, $login_nonce, $redirect_to, $error_msg
...
+						'<a href="%1$s">%2$s</a>',
+						esc_url( self::login_url( $backup_link_args ) ),
+						esc_html( $backup_provider->get_alternative_provider_label() )
+					);
+				}
 			}
-			?>
+
+			/**
+			 * Filters the html links displayed on the two-factor login form.
```
masteradhoc:
```suggestion
			 * Filters the links displayed on the two-factor login form.
```
small clarification

class-two-factor-core.php
```diff
@@ -1065,37 +1065,58 @@ public static function login_html( $user, $login_nonce, $redirect_to, $error_msg
...
+			 *
+			 * Plugins can use this filter to modify or add links to the two-factor authentication
+			 * login form, allowing users to select backup methods for authentication or provide documentation links.
+			 *
+			 * @since 0.16.0
+			 *
+			 * @param array<string> $links An array of links displayed on the two-factor login form.
```
masteradhoc:
```suggestion
			 * @param array $links An array of links displayed on the two-factor login form, each with `url` and `label` keys.
```

georgestephanis:
I'll give it a look in the AM, but if anyone beats me to it feel free to ⛴️

PR #616, opened 2024-07-12
Bump tested upto version to WP 6.6

mehul0810:
## What?
This PR has been tested against new WP 6.6 and updated the version for tested upto under `readme.txt` to 6.6

## Why?
This is necessary as WP 6.6 is nearing to the release.


jeffpaul:
@kasparsd if we're not going to get the 0.10.0 release out in the near future, perhaps pushing this update to SVN will help show 6.6 as compatible for the plugin?

dd32:
I've bumped the plugins tested-up-to on w.org.

cartpauj:
This should be 6.7 now.

Issue #618, opened 2024-07-15
Please use PHPStan


szepeviktor:
### Describe the bug

There are nasty things in the code, e.g. `return parent::__construct();`

### Steps to Reproduce

The simplest way.
```
composer require --dev szepeviktor/phpstan-wordpress
vendor/bin/phpstan analyze -c vendor/szepeviktor/phpstan-wordpress/extension.neon includes/ providers/ class-*.php two-factor.php --level=0
```

dd32:
Hi!

Can you double-check the command provided?

```
$ composer require --dev szepeviktor/phpstan-wordpress

./composer.json has been updated
Running composer update szepeviktor/phpstan-wordpress
Loading composer repositories with package information
Updating dependencies
Lock file operations: 3 installs, 0 updates, 0 removals
  - Locking php-stubs/wordpress-stubs (v6.5.3)
  - Locking phpstan/phpstan (1.11.7)
  - Locking szepeviktor/phpstan-wordpress (v1.3.5)
Writing lock file


$ vendor/bin/phpstan analyze -c vendor/szepeviktor/phpstan-wordpress/extension.neon includes/ providers/ class-*.php two-factor.php --level=0
 14/14 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 [OK] No errors

$ git remote -v
origin	git@github.com:WordPress/two-factor.git (fetch)
origin	git@github.com:WordPress/two-factor.git (push)

$ git branch
* master
```

szepeviktor:
This is surreal. 14/14: all files are checked 😲 

I cannot find the reason why there are no errors.

dd32:
I fully suspect it's related to my environment somehow :)

Thank you for the PR, it should be enough to review the changes and have them merged without issue.

(we might need to get the GitHub actions working prior to merge however, to ensure no regressions)

szepeviktor:
I'm contributing to PHPStan for years and I have no clue what is going on!

kasparsd:
PHPstan reporting and internal cache is the only unreliably thing I haven't been able to solve.

szepeviktor:
> PHPstan reporting and internal cache

Please do report your findings.
https://github.com/phpstan/phpstan/issues/new/choose
I need PHPStan to get better.

kasparsd:
@szepeviktor I got the same results out of the box on my local PHP 7.4:

[...]

PR #619, opened 2024-07-15
Fix errors reported by PHPStan

szepeviktor:
## What?

From #618

## Testing Instructions

```
vendor/bin/phpstan analyze -c vendor/szepeviktor/phpstan-wordpress/extension.neon includes/ providers/ class-*.php two-factor.php --level=0
```


providers/class-two-factor-fido-u2f-admin.php
```diff
@@ -230,7 +230,7 @@ public static function show_user_profile( $user ) {
 	 * @static
 	 *
 	 * @param int $user_id User ID.
-	 * @return false
+	 * @return void|never
```
szepeviktor, line 233: This is dubious.
`catch` should do better.

dd32: This file is pending removal via #439 so not too much effort needs to happen here :)

kasparsd (approved):
This is a nice cleanup! Looks good!

szepeviktor:
Glad to contribute.

PR #620, opened 2024-07-29
Fire an action when a user revalites their 2FA session.

dd32:
## What?

This adds an action for when a user revalidates their session.

This is similar to the existing `two_factor_user_authenticated` action for when the user authenticated via 2FA in the first place.

## Why?

I'm needing to have a cookie set for when the revalidation logic is triggered, and neither Two-Factor nor WordPress (in it's session handler) have a way to detect that the current sessions 2FA has been revalidated / modified.

## Changelog Entry

Added - A New hook was added that fires when a user revalidates their 2FA session.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1404,6 +1404,8 @@ class Two_Factor_Core {
 			'two-factor-login'    => time(),
 		) );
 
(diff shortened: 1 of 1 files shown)


kasparsd (approved):
Looks good!

Issue #622, opened 2024-08-06
Replace table tags with div tags to match Application Passwords styles


christofervas:
### Is your enhancement related to a problem? Please describe.

The current design of the "Two-Factor Options" section uses HTML tables, which results in a design mismatch compared to the "Application Passwords" section, which utilizes div tags for its headings. This inconsistency is visually noticeable and affects the overall cohesiveness of the UI, as demonstrated in the attached screenshot where basic styles have been applied.

### Proposed Solution

To improve the visual consistency and maintain a uniform design language across the profile settings page, I recommend replacing the "Two-Factor Options" heading with div tags, aligning it with the styling approach used in the "Application Passwords" section.

### Designs

![Screenshot 2024-08-06 122511](https://github.com/user-attachments/assets/99b81db3-6972-4407-ba61-47b30a99626e)

### Please confirm that you have searched existing issues in this repository.

Yes

dd32:
Hi @christofervas,

The image you've posted appears to show an interface that's not entirely core WordPress, are you using WordPress.com hosting by chance?

FWIW though, I do agree in changing this to the Application Passwords table format, as it suits that display type better than it does the existing table structure used by other profile fields on the page:

<img width="1568" alt="Screenshot 2024-08-19 at 2 59 31 PM" src="https://github.com/user-attachments/assets/6a282e3c-c6c6-4622-bc93-117cdeb3910b">

christofervas:
Hey @dd32 thank you for considering this enchantment. Hope to see it in the next release.

This is not a WordPress.com hosting but we have applied some custom styles in the admin area.

Issue #536, opened 2023-03-14
Accessibility of layout table and QR image


steverep:
Some basic a11y improvements are needed:

1. The QR code image has no alt text
2. The section heading "Two-Factor Options" should be an `<h2>`
3. Please get rid of the antiquated outer layout table (or at least mark it up properly with `role="none presentation"`)

Issue #342, opened 2020-03-30
Onboarding: checking both "enabled" and "primary" is confusing


paulschreiber:
New users are having trouble completing the setup. They often check "primary" for TOP, but do not check "enabled."

When this happens, 2FA is not enabled, and they have a useless barcode in their authenticator app, which they have to remove.

It would be helpful if, upon form submission, when TOTP is checked as primary and a valid code is entered, then TOTP gets enabled.

My1:
in my opinion all 2FA set as primary shoulld be automatically enabled I mean it IS primary after all.

also maybe make the enabled checkbox on primaries grey so they cant be disabled

kasparsd:
@paulschreiber What are your thoughts on removing the primary selector completely? What if the plugin made the choice and offered the most secure method as the primary? We could also add filter for those who need to change it.

paulschreiber:
@kasparsd That sounds great. I agree 100%.

My1:
just make sure you dont need a faulure condition to switch. I recently was on a site which has webauthn as default when enabled for the login but to switch to normal login the webauthn had to fail, otherwise the button/link wouldnt appear. which basically made it impossible to login on my phone browser

paulschreiber:
@My1 Where was that? That's pretty bad.

My1:
devolutions. the makers of for example the wayk remote support tool. and yes it IS bad but at least they used webauthn which shows stuff better in the browser and handles errors nicer (like in many cases you do not even need a retry button, which this project in my opinion really could need), also they use it for passwordless (but not usernameless (resident keys), which is good, as credential management is non existent on most devices) which would be a nice additional scenario, username PIN/finger BOOM and you're in.

7php:
> @paulschreiber What are your thoughts on removing the primary selector completely? What if the plugin made the choice and offered the most secure method as the primary? We could also add filter for those who need to change it.

I would not do that. It is not up to the plugin to judge which is the best scenario for the actual user. This *taste* varies among users at different level.

You just need to either have a `primary` option or an `enable` option, not both. That's it.

paulschreiber:
Thinking about this more, both the checkbox and the radio button should be removed. They clutter up the UI and don't confer any advantage.

Facebook/Google/GitHub let you register as many methods as you want, and the system starts with the most secure and gives you a "click/tap to try another method" prompt.

blogtutor:
[...]

PR #623, opened 2024-08-16
Update two factor options layout

thrijith:
- Update and move two factor option heading out of table
- Align two factor options with title

## What?

This PR updates the layout of Two Factor options to match Application Passwords, fixes https://github.com/WordPress/two-factor/issues/622

## Why?

To make sure the UI consistent, as mentioned in https://github.com/WordPress/two-factor/issues/622, the request in it is to use div tags but we can't do it as `fieldset` tag is required to make all inputs disabled on a condition, which won't work with `div` without the some additional CSS.

## How?

- It moves the `Two-Factor Options` out of the table and adds a heading for it, adds some style to align the two factor options table with heading.

## Testing Instructions

- Start the env after checking out into the branch go  to http://localhost:8888/wp-admin/profile.php, scroll down two factor options, confirm it matches the Application Passwords.

## Screenshots or screencast

CleanShot 2024-08-16 at 17 33 31@2x


class-two-factor-core.php
```diff
@@ -1721,12 +1721,10 @@ public static function user_two_factor_options( $user ) {
...
 		wp_nonce_field( 'user_two_factor_options', '_nonce_user_two_factor_options', false );
 		?>
+		<h3><?php esc_html_e( 'Two-Factor Options', 'two-factor' ); ?></h3>
 		<input type="hidden" name="<?php echo esc_attr( self::ENABLED_PROVIDERS_USER_META_KEY ); ?>[]" value="<?php /* Dummy input so $_POST value is passed when no providers are enabled. */ ?>" />
 		<table class="form-table">
 			<tr>
-				<th>
-					<?php esc_html_e( 'Two-Factor Options', 'two-factor' ); ?>
-				</th>
 				<td>
```
dd32, line 1728: Is it possible to remove the wrapping `<table>`? Does it make any sense to do that?

Looking at application passwords, that uses a `div.application-passwords-list-table-wrapper` here.

That might remove the need for the CSS change as well, as `table.form-table` would no longer be applying.

thrijith: I've removed the `table. form-table` it was adding some CSS to make it 100% width, so have added CSS for the existing `two-factor-methods-table` table.

thrijith:
Does the asset screenshot need an update as well now?

https://github.com/WordPress/two-factor/blob/master/assets/screenshot-1.png

dd32:
> Does the asset screenshot need an update as well now?

It does, but the next release of the plugin will likely need to change other aspects of that page, so I'm not sure it's a priority for this PR.

Perhaps create a new issue to track that?

dd32:
I've pushed some commits that makes the styles closer to core, by inheriting the core css via the expected classes, and adding the `<tfoot>` footer that other code tables do.

This has had some downsides though; as it's caused some UI issues at varying display sizes.
Things that need work:
 - Mobile table layout
 - Table headers wrapping

@thrijith Apologies, this isn't all totally what you were expecting from this PR, but it seemed like to merge this, it would be better to cover all aspects of making it closer to the core UI.

| | Before | After |
| --- | --- | --- |
| Full width | Screenshot 2024-08-22 at 2 33 45 PM | Screenshot 2024-08-22 at 2 24 24 PM |
| Slightly less | Screenshot 2024-08-22 at 2 33 56 PM | Screenshot 2024-08-22 at 2 24 49 PM |
| Mobile | Screenshot 2024-08-22 at 2 34 20 PM |  Screenshot 2024-08-22 at 2 25 00 PM |

thrijith:
Hi @dd32, is there anything required from my end on this PR?

jeffpaul:
@dd32 any further work needed on this PR or is this good to merge?

kasparsd (approved):
This looks good -- safe to merge and can be included in the next point release.

I wonder if any third-party integrations relying on this table are somehow impacted by this. I think WP VIP was adding an SMS option in there, right?

dd32:
Apologies, I didn't get back to this ticket.

Personally I'd prefer to resolve the mobile usability of it as above, but perhaps that can be pushed off to a new issue / PR, as it's arguably not worse than before.  That might be better resolved via #342 

This shouldn't cause any impacts for addon plugins, since there's no major underlying changes, but any that are expecting that specific HTML structure may break if they are using highly scoped CSS/JS selectors that don't account for the removal of the nested table.. which would be a poor choice IMHO.

Also noting, this will partially fix #536

PR #624, opened 2024-08-20
Fix Action unit testing

### LINKED ISSUE #523: Tests should run against WordPress stable and WordPress trunk, and supported PHP versions
Currently the unit tests only run on PHP 7.4 against the stable version of WordPress, with a PHP8 lint.

Ideally, they should run both on the latest Stable of WordPress, WordPress trunk, and ideally all PHP versions supported (or at least the lowest and highest).

dd32:
## What?

This PRs history is a fair bit of a mess, but in short it:
 - Restores GitHub actions
 - Splits the singular action into multiple specific parts
 - Uses wp-env instead of a custom WordPress checkout
 - Tests PHP 7.2~8.3 on WordPress latest and trunk (Fixes #523)

## Why?

Because GitHub actions have been broken for far too long.

## How?

Even I'm not sure I totally understand how.

## Testing Instructions

Just don't. The commit history should show even I don't know what I'm doing.

## Changelog Entry

Changed - Improved unit testing coverage with WordPress & PHP versions.


dd32:
Pretty sure I've broken the build script, `dist/includes/qrcode-generator/` contains all the library files and `dist/includes/qrcode.js` exists, which should be at `dist/includes/qrcode-generator/qrcode.js`.

That build script should be replaced with something not-grunt I suspect.

PR #632, opened 2024-09-17
Accessibility for options page

### LINKED ISSUE #536: Accessibility of layout table and QR image

dd32:
## What?

Fixes #536 this improves the accessibility for the options page by:
 - Setting the heading to a H2 as appropriate for the page
 - Adds a title to the QR code and marks it as an image, based on https://stackoverflow.com/a/4756461

The QR library we use does not provide the needed ability to set the title/role, which is why I've used JS to set these.

I'm not sure of the best text to use as a `title` for the QR code, but I think `Authenticator App QR Code` is reasonably descriptive yet short enough.

## Why?

This makes the page more accessible to screen readers

## Changelog Entry

> Added - New feature.
> Removed - Feature.
> Fixed - Bug fix.
> Security - Vulnerability.

Fixed - Improved accessibility of the options screen.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1721,7 +1721,7 @@ class Two_Factor_Core {
 
 		wp_nonce_field( 'user_two_factor_options', '_nonce_user_two_factor_options', false );
 		?>
-		<h3><?php esc_html_e( 'Two-Factor Options', 'two-factor' ); ?></h3>
+		<h2><?php esc_html_e( 'Two-Factor Options', 'two-factor' ); ?></h2>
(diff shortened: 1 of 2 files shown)


kasparsd (approved):
This is great!

kasparsd:
Tested this on local and everything appears to be working as expected.

Will be creating another pull request with the updated plugin screenshots.

PR #636, opened 2024-09-18
Update screenshots to match the current UI

kasparsd:
## What?
Update screenshots for WP.org plugin listing to match the current UI.

## Why?
There have been updates to the user profile UI in #622.

## Changelog Entry

Documentation: updated screenshots to match the latest plugin user interface.


dd32:
Should we remove the `Security Keys` section from the screenshots?

kasparsd:
@dd32 Yeah, as soon as #439 lands, right? There are bunch of merges that could be released before we remove the U2F (and account of that being the only enabled provider).

dd32:
Oops :) I didn't realise #439 hadn't landed. Nevermind!

PR #637, opened 2024-09-18
Delete user meta on plugin uninstall

### LINKED ISSUE #630: Uninstallation Not Deleting Plugin Data
WordPress plugins [are supposed to clear out their data when uninstalled](https://developer.wordpress.org/plugins/plugin-basics/uninstall-methods/). As part of a security review of the plugin, we found that it doesn't currently do that.

It looks like there are 14 metadata keys that can be added to users' metadata that should be removed during uninstallation:

- _two_factor_nonce
- _two_factor_last_login_failure
- _two_factor_failed_login_attempts
- _two_factor_password_was_reset
- _two_factor_enabled_providers
- _two_factor_provider
- _two_factor_backup_codes
- _two_factor_email_token_timestamp
- _two_factor_email_token
- _two_factor_fido_u2f_register_request
- _two_factor_fido_u2f_login_request
- _two_factor_fido_u2f_registered_key
- _two_factor_totp_key
- _two_factor_totp_last_successful_login

There are [several methods that can be used to delete those](https://wordpress.stackexchange.com/questions/60769/best-practice-way-to-delete-user-meta-data-during-plugin-uninstall/112921).

This could be classified as a security issue because of what is stored. You [suggest reporting security issues](https://github.com/WordPress/two-factor/security) to a bounty program, but that only accepts reports on a limited range of vulnerabilities. [We raised that problem (and others)](https://github.com/WordPress/health-check/issues/465#issuecomment-2251047252) with WordPress' handling of reporting security issues through a bug bounty program with another plugin from WordPress in July, without a response or a change in your handling of the situation so far.

kasparsd:
## What?

Fixes #630.

## Why?

Delete all used user meta keys and options during plugin uninstall as expected.

## How?

1. Register an [uninstall hook](https://developer.wordpress.org/reference/functions/register_uninstall_hook/) for the plugin.

2. During uninstall, collect all meta keys and options used by two-factor providers through a known class helper. Delete them by looping through each user (a performance concern on larger sites).

Consider using a direct database query to purge the user meta by keys to avoid looping through every user on larger sites.

## Testing Instructions

1. Setup the plugin with a few users and two-factor methods enabled for several users.
2. Uninstall the plugin (in addition to deactivating).
3. Activate the plugin again and confirm that no settings for providers have been preserved.

## Changelog Entry

Added: register an uninstall hook for the plugin to delete all user meta data when the plugin is completely uninstalled.


dd32 (approved):
General direction WFM.

See comment on the related ticket about deleting meta values by `mid` instead for performance.

class-two-factor-core.php
```diff
@@ -132,13 +132,67 @@ public static function add_hooks( $compat ) {
...
+
+		// Delete options first since that is faster.
+		if ( ! empty( $option_keys ) ) {
+			foreach ( $option_keys as $option_key ) {
+				delete_option( $option_key );
+			}
+		}
+
+		foreach ( $user_meta_keys as $meta_key ) {
+			delete_metadata( 'user', null, $meta_key, null, true );
```
kasparsd, line 186: WP core uses this for deleting all post meta [`delete_post_meta_by_key()`](https://developer.wordpress.org/reference/functions/delete_post_meta_by_key/) by key so we can re-use it as well.

kasparsd:
It is fun to test this locally -- it purges the whole plugin directory along with all git history. I even tried to create another symlinked plugin instances hoping that WP would only delete the symlink but it still took everything 😅

vbondzio:
Hi Folks, 

What would be considered sufficient DB cleanup in previous releases, just:

`DELETE FROM wp_whatever_usermeta WHERE meta_key LIKE "_two_factor%"; `

?

This is also applicable for re-installations of the plugin (<0.10.0), for example in the _**completely hypothetical**_ situation in which you accidentally reset your auth device without backing up the secrets and realize that your rarely opened 2FA recovery password manager DB had a racy sync at some point in the past. So if the only way in is deleting the plugin dir manually, then removing and re-installing, it would leave you in a spot in which you couldn't re-create a new secret (or even reuse the same) as you couldn't revalidate for the existing users. Completely hypothetical of course. 

Cheers,

Valentin

class-two-factor-core.php
```diff
@@ -150,29 +204,29 @@ public static function get_providers() {
...
 		 * Filter the supplied providers.
 		 *
-		 * This lets third-parties either remove providers (such as Email), or
-		 * add their own providers (such as text message or Clef).
-		 *
 		 * @param array $providers A key-value array where the key is the class name, and
 		 *                         the value is the path to the file containing the class.
 		 */
-		$providers = apply_filters( 'two_factor_providers', $providers );
+		$additional_providers = apply_filters( 'two_factor_providers', $providers );
```
ocean90, line 210: @kasparsd Using the same existing filter seems wrong here. Shouldn't this be a new filter so you only get the additional providers?
Right now the filter is called twice in `get_providers()`, see https://github.com/WordPress/two-factor/blob/6a95e7ffdebb2e0408b4ccb298e9df16d5b6998c/class-two-factor-core.php#L264-L275.

ocean90: @kasparsd This still seems wrong to me and is now released. Why running the filter twice?

kasparsd: Here is a pull request that removes the duplicate #651.

PR #638, opened 2024-09-18
Add basic PHPStan linter

### LINKED ISSUE #618: Please use PHPStan

kasparsd:
## What?

Introduces [phpstan](https://phpstan.org) with WordPress core stubs through [szepeviktor/phpstan-wordpress](https://github.com/szepeviktor/phpstan-wordpress).

## Why?

Fixes #618.

Static analysis helps us uncover potential bugs and edge cases. Introducing the tooling didn't require a lot adjustments thanks to work already done in #619.

## How?

Add the tooling as Composer development dependency and introduce additional lint scripts in `composer.json` and `package.json`.

## Testing Instructions

1. Install composer dependencies via `composer install` after checking out this branch.
2. Run `composer lint-phpstan` and confirm that `[OK] No errors` is returned.

## Changelog Entry

Development: Introduce PHPstan tooling for development.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /dist/
 /tests/logs/
 .phpunit.result.cache
+phpstan.neon
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -93,7 +93,7 @@ class Two_Factor_Core {
 	 * @since 0.1-dev
 	 */
 	public static function add_hooks( $compat ) {
-		add_action( 'init', array( __CLASS__, 'get_providers' ) );
+		add_action( 'init', array( __CLASS__, 'get_providers' ) ); // @phpstan-ignore return.void
 		add_action( 'wp_login', array( __CLASS__, 'wp_login' ), 10, 2 );
 		add_filter( 'wp_login_errors', array( __CLASS__, 'maybe_show_reset_password_notice' ) );
 		add_action( 'after_password_reset', array( __CLASS__, 'clear_password_reset_notice' ) );
--- a/composer.json
+++ b/composer.json
@@ -31,12 +31,14 @@
     "phpcompatibility/phpcompatibility-wp": "^2.1",
     "phpunit/phpunit": "^8.5|^9.6",
     "spatie/phpunit-watcher": "^1.23",
+    "szepeviktor/phpstan-wordpress": "^1.3",
     "wp-coding-standards/wpcs": "^3.1",
     "yoast/phpunit-polyfills": "^2.0"
   },
   "scripts": {
     "lint": "phpcs",
     "lint-compat": "phpcs -p --standard=PHPCompatibilityWP --runtime-set testVersion 7.2- --extensions=php --ignore='tests/,dist/,includes/Yubico/,vendor/,node_modules/' .",
+    "lint-phpstan": "phpstan analyse --memory-limit=1G",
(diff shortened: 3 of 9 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@
 /dist/
 /tests/logs/
 .phpunit.result.cache
+phpstan.neon
```
kasparsd, line 8: Allow local overrides.

class-two-factor-core.php
```diff
@@ -93,7 +93,7 @@ class Two_Factor_Core {
 	 * @since 0.1-dev
 	 */
 	public static function add_hooks( $compat ) {
-		add_action( 'init', array( __CLASS__, 'get_providers' ) );
+		add_action( 'init', array( __CLASS__, 'get_providers' ) ); // @phpstan-ignore return.void
```
kasparsd, line 96: @szepeviktor The default phpstan table report doesn't include the actual error identifiers making it hard to ignore them. Is there an output format that includes them?

composer.json
```diff
@@ -31,12 +31,14 @@
...
     "phpunit/phpunit": "^8.5|^9.6",
     "spatie/phpunit-watcher": "^1.23",
+    "szepeviktor/phpstan-wordpress": "^1.3",
     "wp-coding-standards/wpcs": "^3.1",
     "yoast/phpunit-polyfills": "^2.0"
   },
   "scripts": {
     "lint": "phpcs",
     "lint-compat": "phpcs -p --standard=PHPCompatibilityWP --runtime-set testVersion 7.2- --extensions=php --ignore='tests/,dist/,includes/Yubico/,vendor/,node_modules/' .",
+    "lint-phpstan": "phpstan analyse --memory-limit=1G",
```
kasparsd, line 41: Running without bumping the memory limit can sometimes return passing checks when they actually didn't run.

class-two-factor-core.php
szepeviktor: Yes. Here is an example
https://github.com/TicketSwap/phpstan-error-formatter

szepeviktor: but there is https://github.com/phpstan/phpstan-src/pull/3112
https://phpstan.org/user-guide/output-format

szepeviktor: `raw` should also display it

kasparsd: Unfortunately, it doesn't:

no-error-ident

szepeviktor: try adding a `-v`
https://github.com/phpstan/phpstan-src/blob/62c81bac66f2558c2768914d3a79abad7c4c8b3d/src/Command/ErrorFormatter/RawErrorFormatter.php#L22

kasparsd: That worked also for the default output. I'm going to add that in as the default in the `composer.json` script.

verbose

szepeviktor:
Is phpstan output on the current level empty?

dd32 (approved):
LGTM

PR #640, opened 2024-09-18
Release 0.10.0

kasparsd:
Push out all the minor improvements before a major release.

## What's Changed
* Rely on just-in-time translation loading by @swissspidy in https://github.com/WordPress/two-factor/pull/608
* Update/headers by @jeffpaul in https://github.com/WordPress/two-factor/pull/610
* Update short description by @jeffpaul in https://github.com/WordPress/two-factor/pull/612
* Fix typos by @szepeviktor in https://github.com/WordPress/two-factor/pull/617
* Bump tested upto version to WP 6.6 by @mehul0810 in https://github.com/WordPress/two-factor/pull/616
* Fire an action when a user revalites their 2FA session. by @dd32 in https://github.com/WordPress/two-factor/pull/620
* Bump minimum WP to 6.3, minimum PHP to 7.2. by @dd32 in https://github.com/WordPress/two-factor/pull/625
* Remove old grunt deploy related code. See #543 by @dd32 in https://github.com/WordPress/two-factor/pull/627
* Fix Action unit testing by @dd32 in https://github.com/WordPress/two-factor/pull/624
* Bump axios from 1.6.8 to 1.7.4 by @dependabot in https://github.com/WordPress/two-factor/pull/626
* Bump braces from 3.0.2 to 3.0.3 by @dependabot in https://github.com/WordPress/two-factor/pull/613
* Bump webpack from 5.91.0 to 5.94.0 by @dependabot in https://github.com/WordPress/two-factor/pull/628
* Update two factor options layout by @thrijith in https://github.com/WordPress/two-factor/pull/623
* Bump send and express by @dependabot in https://github.com/WordPress/two-factor/pull/634
* Accessibility for options page by @dd32 in https://github.com/WordPress/two-factor/pull/632
* Fix errors reported by PHPStan by @szepeviktor in https://github.com/WordPress/two-factor/pull/619
* Fix failing unit test by @kasparsd in https://github.com/WordPress/two-factor/pull/639
* Add basic PHPStan linter by @kasparsd in https://github.com/WordPress/two-factor/pull/638
* Update screenshots to match the current UI by @kasparsd in https://github.com/WordPress/two-factor/pull/636
* Improve discoverability by @kasparsd in https://github.com/WordPress/two-factor/pull/635
* Delete user meta on plugin uninstall by @kasparsd in https://github.com/WordPress/two-factor/pull/637

## New Contributors
* @szepeviktor made their first contribution in https://github.com/WordPress/two-factor/pull/617
[...]

PR #641, opened 2024-09-19
Keep the requirements at PHP 5.6 and WP 4.6 until we have a practical reason

kasparsd:
## What?

Reverts #625.

## Why?

Although we want users to upgrade their PHP and WP, the plugin code currently supports PHP 5.6+ and WP 4.6 so we keep those requirements until we've given plenty of notice that version 1.0.0 of this plugin will no longer support anything below 7.4 and a year old WP core. 

## How?

- Update README to include an official policy on the supported versions of PHP and WP core.
- Use the upgrade notice up until version 1.0.0 to inform users about the upcoming enforcement.

## Changelog Entry

> Added - New feature.
> Removed - Feature.
> Fixed - Bug fix.
> Security - Vulnerability.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -31,7 +31,7 @@ Here is a list of action and filter hooks provided by the plugin:
 
 = PHP and WordPress Version Support =
 
-Starting from version 1.0.0 of this plugin, it will maintain compatibility with WordPress core versions that are up to one year old, as well as the minimum PHP version supported by those WordPress versions.
+Starting from version 1.0.0 of this plugin, it will maintain compatibility with WordPress core versions that are up to one year old, as well as <a href="https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/">the minimum PHP version supported by those WordPress versions</a>.
 
 == Frequently Asked Questions ==
 
--- a/readme.txt
+++ b/readme.txt
@@ -29,6 +29,10 @@ Here is a list of action and filter hooks provided by the plugin:
 - `two_factor_user_authenticated` action which receives the logged in `WP_User` object as the first argument for determining the logged in user right after the authentication workflow.
 - `two_factor_token_ttl` filter overrides the time interval in seconds that an email token is considered after generation. Accepts the time in seconds as the first argument and the ID of the `WP_User` object being authenticated.
 
+= PHP and WordPress Version Support =
+
+Starting from version 1.0.0 of this plugin, it will maintain compatibility with WordPress core versions that are up to one year old, as well as the minimum PHP version supported by those WordPress versions.
+
 == Frequently Asked Questions ==
 
 = How can I send feedback or get help with a bug? =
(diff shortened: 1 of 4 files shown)


readme.txt
```diff
@@ -29,6 +29,10 @@ Here is a list of action and filter hooks provided by the plugin:
 - `two_factor_user_authenticated` action which receives the logged in `WP_User` object as the first argument for determining the logged in user right after the authentication workflow.
 - `two_factor_token_ttl` filter overrides the time interval in seconds that an email token is considered after generation. Accepts the time in seconds as the first argument and the ID of the `WP_User` object being authenticated.
 
+= PHP and WordPress Version Support =
```
kasparsd, line 32: @dd32, @jeffpaul What do you think of this policy?

readme.txt
```diff
@@ -47,6 +51,13 @@ To report a security issue, please visit the [WordPress HackerOne](https://hacke
 2. U2F Security Keys section under User Profile.
 3. Email Code Authentication during WordPress Login.
 
+== Upgrade Notice ==
+
+<!-- Keep this in sync with the next version up until 1.0.0. -->
+= 0.10.0 =
+
+From version 1.0.0, this plugin will support WordPress versions up to one year old and the minimum PHP version they require.
```
kasparsd, line 59: We use the built-in WP plugin upgrade notice UI to inform users about the upcoming change.

dd32: The majority of users will never see such a message, whether it be in the Changelog, Description, or Upgrade Notice (which is seen only on the Dashboard -> Updates page)

dd32:
I don't personally support this, as there's no reasoning given to retain it.

Please ensure that the testing pipeline is updated inline with this, but please keep a **maintainable** test runner. That's one of the main reasons for the increase, the inability to test with ancient versions.

So if you're looking at it from a practical point of view; AFAICT no-one tests the plugin with old PHP / WPs, and I definately do not care about them. Practically so few of them will update.

If you're looking at it from a usage point of view, I think it makes sense to follow Core's "Less than 5% usage is time to move on from it" which per https://github.com/WordPress/two-factor/pull/640#issuecomment-2360006532 we're well past.

Duplicating the relevant part here:
> I did some digging for stat data for you, for 0.9.x of the plugin:
> - the PHP bump will affect 0.15% of users (1% If we include users of < 0.9)
> - the WP bump will affect 2.6% of 0.9.1 users (A singular site somewhere is using 0.9.0 + WP 6.2 according to the data, and nothing lower)

> From version 1.0.0, this plugin will support WordPress versions up to one year old and the minimum PHP version they require.

I can support this, although I'd prefer a much more aggressive support deprecation.
As of today, -1 year would mean WordPress 6.3 / PHP 7.0.
I'd prefer that each x.y version simply support the last major WordPress and it's associated requirements.
If there's a security update required to the plugin, that can be backported.

My reasoning is that WordPress.org plugins usage data suggests that the majority of plugin users run up-to-date WordPress (likely thanks to hosts and core auto-updates) and outdated plugins. It's much more rare for a recently updated plugin to be used on an older site.

jeffpaul:
>  I'd prefer that each x.y version simply support the last major WordPress and it's associated requirements.

This matches my general preference and one in which we've mostly put in place at 10up (though we tend to use WP-2 so two versions back and not just 1).

kasparsd:
Thank you all for providing perspective and input. Let's keep the dependency bumps as is.

PR #642, opened 2024-09-19
Encourage setting up a second recovery method 

### LINKED ISSUE #485: Encourage setting up a recovery factor
Ideally users should setup two factors, one as a primary and one as a backup. e.g., WebAuthn as the primary and TOTP as the backup; or TOTP as the primary and Backup Codes as the backup.

Otherwise, they could get locked out of their account. On smaller sites an admin could reset them, but that's not practical on larger sites, or sites where the admin doesn't personally know the user.

Rough idea:

Screen Shot 2022-10-20 at 9 39 20 AM

kasparsd:
## What?

Fixes #485.

## Why?

Encourage users to configure at least two two-factor methods to prevent them from being locked out of account in case of loosing access to the primary method.

## How?

Showing a notice before any of the methods are configured _and_ when only one method is configured.

## Testing Instructions

1. Setup a plugin and go to your profile to setup the methods.
2. Select only email, save the settings and confirm that a message is shown suggesting to setup another method.

## Screenshots or screencast

No methods configured:

notice-default

Only one method configured:

notice-second

Multiple notices:

multiple-notices

## Changelog Entry

Added a notice to the user profile suggesting to configure at least two two-factor methods for ensuring access.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1821,9 +1821,21 @@ class Two_Factor_Core {
 			$show_2fa_options ? '' : 'disabled="disabled"'
 		);
 
-		wp_nonce_field( 'user_two_factor_options', '_nonce_user_two_factor_options', false );
+		$notices = [];
+		if ( empty( $enabled_providers ) ) {
+			$notices[] = __( 'Configure a primary two-factor method along with a backup method, such as Recovery Codes, to avoid being locked out if you lose access to your primary method.', 'two-factor' );
+		} elseif ( 1 === count( $enabled_providers ) ) {
+			$notices['warning'] = __( 'To prevent being locked out of your account, consider enabling a backup method like Recovery Codes in case you lose access to your primary authentication method.', 'two-factor' );
+		}
+
 		?>
 		<h2><?php esc_html_e( 'Two-Factor Options', 'two-factor' ); ?></h2>
+		<?php foreach ( $notices as $notice_type => $notice ) : ?>
+		<div class="<?php echo esc_attr( $notice_type ? 'notice inline notice-' . $notice_type : '' ); ?>">
+			<p><?php echo esc_html( $notice ); ?></p>
+		</div>
+		<?php endforeach; ?>
+		<?php wp_nonce_field( 'user_two_factor_options', '_nonce_user_two_factor_options', false ); ?>
 		<input type="hidden" name="<?php echo esc_attr( self::ENABLED_PROVIDERS_USER_META_KEY ); ?>[]" value="<?php /* Dummy input so $_POST value is passed when no providers are enabled. */ ?>" />
 		<table class="wp-list-table widefat fixed striped table-view-list two-factor-methods-table">
 			<thead>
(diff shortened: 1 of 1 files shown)


class-two-factor-core.php
```diff
@@ -1803,16 +1805,16 @@ public static function user_two_factor_options( $user ) {
...
+				urlencode( self::get_user_settings_page_url( $user->ID ) . '#two-factor-options' ),
+				self::get_user_two_factor_revalidate_url()
+			);
 
-			printf(
-				'<div class="notice notice-warning inline"><p>%s</p></div>',
-				sprintf(
-					__( 'To update your Two-Factor options, you must first revalidate your session.', 'two-factor' ) .
-					'<br><a class="button" href="%s">' . __( 'Revalidate now', 'two-factor' ) . '</a>',
+			$notices['warning two-factor-warning-revalidate-session'] = sprintf(
```
kasparsd, line 1814: I don't like the hidden meaning of the first key string being the class suffix for `notice-*` but that's how we can allow for multiple notices of the same type without introducing some kind of an abstraction.

class-two-factor-core.php
```diff
@@ -1821,9 +1823,20 @@ public static function user_two_factor_options( $user ) {
...
+			$notices['warning two-factor-warning-suggest-backup'] = esc_html__( 'To prevent being locked out of your account, consider enabling a backup method like Recovery Codes in case you lose access to your primary authentication method.', 'two-factor' );
+		}
 		?>
 		<h2><?php esc_html_e( 'Two-Factor Options', 'two-factor' ); ?></h2>
+		<?php foreach ( $notices as $notice_type => $notice ) : ?>
+		<div class="<?php echo esc_attr( $notice_type ? 'notice inline notice-' . $notice_type : '' ); ?>">
+			<p><?php echo $notice; ?></p>
+		</div>
+		<?php endforeach; ?>
+		<p>
```
kasparsd, line 1836: These are now permanent instructions suggest the best-practice setup.

class-two-factor-core.php
```diff
@@ -1821,9 +1823,20 @@ public static function user_two_factor_options( $user ) {
...
 
-		wp_nonce_field( 'user_two_factor_options', '_nonce_user_two_factor_options', false );
+		if ( 1 === count( $enabled_providers ) ) {
+			$notices['warning two-factor-warning-suggest-backup'] = esc_html__( 'To prevent being locked out of your account, consider enabling a backup method like Recovery Codes in case you lose access to your primary authentication method.', 'two-factor' );
+		}
 		?>
 		<h2><?php esc_html_e( 'Two-Factor Options', 'two-factor' ); ?></h2>
+		<?php foreach ( $notices as $notice_type => $notice ) : ?>
+		<div class="<?php echo esc_attr( $notice_type ? 'notice inline notice-' . $notice_type : '' ); ?>">
+			<p><?php echo $notice; ?></p>
```
thrijith, line 1833: I see the values are already escaped early should there be a `wp_kses` here to be safe / should we mention the values are already escaped above?

jeffpaul: @kasparsd FYI on the question above for you

kasparsd: This is now done in 2284d6a37c7f106a1b275900d6d53db675129bbb. We're using `wp_kses_post()` similar to how WP core does it here:

https://github.com/WordPress/WordPress/blob/369a6754887a948830f2107602648e46cb370dfb/wp-admin/includes/class-wp-site-health.php#L2419

PR #643, opened 2024-09-19
Enable TOTP method when method is configured

### LINKED ISSUE #556: TOTP: After successful set up, JS should check the "Enabled" checkbox for the TOTP provider

In two-factor v0.8.0, TOTP was switched over to use the REST API and uses JS to re-render the TOTP options portion of the table on successful TOTP set up. However, after successful TOTP set up, the TOTP provider's "Enabled" checkbox is not checked. This is a usability issue.

### Proposed Solution

After successful TOTP set up, the JS should check the "Enabled" checkbox for the TOTP provider in this block: https://github.com/WordPress/two-factor/blob/836ef620796d6c678e1bf92f853f0300e4f2d97d/providers/class-two-factor-totp.php#L363-L365.

Also, I would also like to see a custom jQuery `trigger()` call so 3rd-party developers can do something after successful TOTP set up or if there is an error with TOTP set up.

### Designs

_No response_

Currently, I'm hooking into jQuery's `ajaxComplete` event handler to do what I want to do. See https://github.com/r-a-y/bp-two-factor/commit/e7f0da9e6be91d97218d42fdbead8d01b79798ac .

Yes

kasparsd:
## What?

Users expect a method to be enabled if they chose to configure it.

## Why?

Fixes #556.

## Testing Instructions

Setup workflow:

1. Configure the TOTP method by submitting the generated code.
2. Confirm the TOTP method is now enabled in the table.

Reset workflow: 

1. Click on RESET TOTP method button.
2. Confirm the TOTP method is now disabled in the table.

## Screenshots or screencast

https://github.com/user-attachments/assets/bd3a9143-86b5-4fa9-b31d-4e27f395f391

## Changelog Entry

> Added - New feature.
> Removed - Feature.
> Fixed - Bug fix.
> Security - Vulnerability.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1896,10 +1896,8 @@ class Two_Factor_Core {
 	 * @return bool True if the provider was enabled, false otherwise.
 	 */
 	public static function enable_provider_for_user( $user_id, $new_provider ) {
-		$available_providers = self::get_providers();
-
 		// Ensure the provider is even available.
-		if ( isset( $available_providers[ $new_provider ] ) ) {
+		if ( ! array_key_exists( $new_provider, self::get_providers() ) ) {
 			return false;
 		}
 
@@ -1930,7 +1928,7 @@ class Two_Factor_Core {
 	 */
 	public static function disable_provider_for_user( $user_id, $provider_to_delete ) {
 		// Check if the provider is even enabled.
-		if ( ! in_array( $provider_to_delete, self::get_providers_classes(), true ) ) {
+		if ( ! array_key_exists( $provider_to_delete, self::get_providers() ) ) {
 			return false;
 		}
 
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1893,11 +1893,18 @@ class Two_Factor_Core {
 	 * @return bool True if the provider was enabled, false otherwise.
 	 */
 	public static function enable_provider_for_user( $user_id, $new_provider ) {
+		if ( ! in_array( $new_provider, self::get_providers_classes(), true ) ) {
+			return false;
+		}
+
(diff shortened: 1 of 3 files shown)


class-two-factor-core.php
```diff
@@ -1902,30 +1901,20 @@ public static function user_two_factor_options( $user ) {
 	 * @return bool True if the provider was enabled, false otherwise.
 	 */
 	public static function enable_provider_for_user( $user_id, $new_provider ) {
-		$available_providers = self::get_providers();
```
kasparsd, line 1905: This method is very heavy as it will instantiate new set of provider classes on every call. That's why we introduced `get_provider_classes()` earlier which doesn't do any of that.

providers/class-two-factor-totp.php
```diff
@@ -386,8 +391,10 @@ public function user_two_factor_options( $user ) {
 
 							$error.find('p').text( errorMessage );
 
+							$( '#enabled-Two_Factor_Totp' ).prop( 'checked', false );
```
kasparsd, line 394: We now enable the TOTP method when the key has been registered by default.

class-two-factor-core.php
kasparsd: Had to switch back to `get_providers()` because the classes helper still needs a list of available providers.

providers/class-two-factor-totp.php
```diff
@@ -375,6 +379,7 @@ public function user_two_factor_options( $user ) {
 								user_id: <?php echo wp_json_encode( $user->ID ); ?>,
 								key: key,
 								code: code,
+								enable_provider: true,
```
kasparsd, line 382: Instruct the validator to also enable the provider, if everything is a pass.

PR #651, opened 2024-12-02
Remove duplicate `two_factor_providers` filter calls to allow disabling core providers

kasparsd:
See https://github.com/WordPress/two-factor/pull/637#discussion_r1831364092

## What?

In #637 we introduced a dedicated helper to retrieve all possible two-factor providers (even if disabled) so that we could purge all their added user meta and options during uninstall.

Accidentally it added another instance of the `two_factor_providers` filter which was already defined in `get_providers()`.

## How?

- Introduce a method for getting the default built-in providers. This is used as the starting point for enabled providers and when collecting all providers to clean during plugin uninstall.
- Pass that default list of providers to the `two_factor_providers` filter when getting "enabled" providers from `get_providers()`.
- Refactor `get_providers_classes()` to accept a list of provider names and their class files instead of attempting to retrieve some provider state. During uninstall, we need all providers while during runtime only the enabled ones.

## Testing Instructions

- Use the plugin without any `two_factor_providers` filters. All default providers are available to all users.
- Attach a filter to `two_factor_providers` that limits just one provider. Attempt to uninstall the plugin and confirm that it removes also the meta for the disabled providers (see the phpunit test case cover this).

## Changelog Entry

Changed - removed duplicate two-factor provider filter.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -151,12 +151,7 @@ class Two_Factor_Core {
 
 		$providers = self::get_default_providers();
 
-		/**
-		 * Include additional providers to be uninstalled.
-		 *
-		 * @param array $providers A key-value array where the key is the class name, and
-		 *                         the value is the path to the file containing the class.
-		 */
+		/** This filter is documented in the get_providers() method */
 		$additional_providers = apply_filters( 'two_factor_providers', $providers );
 
 		// Merge them with the default providers.
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -289,6 +289,7 @@ class Two_Factor_Core {
 		// Map provider keys to classes so that we can instantiate them.
 		$providers = self::get_providers_classes( $providers );
 
+		// TODO: Refactor this to avoid instantiating the provider instances every time this method is called.
 		foreach ( $providers as $provider_key => $provider_class ) {
 			try {
 				$providers[ $provider_key ] = call_user_func( array( $provider_class, 'get_instance' ) );
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -241,9 +241,9 @@ class Two_Factor_Core {
 			$class = apply_filters( "two_factor_provider_classname_{$provider_key}", $class, $path );
 
 			/**
-			 * Confirm that it's been successfully included before instantiating.
+			 * Confirm that it's been successfully included.
 			 */
-			if ( method_exists( $class, 'get_instance' ) ) {
+			if ( class_exists( $class ) ) {
 				$providers[ $provider_key ] = $class;
 			} else {
 				unset( $providers[ $provider_key ] );
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -161,7 +161,7 @@ class Two_Factor_Core {
 
 		// Merge them with the default providers.
 		if ( ! empty( $additional_providers ) ) {
-			return array_merge( $providers, $additional_providers );
+			$providers = array_merge( $providers, $additional_providers );
 		}
 
 		foreach ( self::get_providers_classes( $providers ) as $provider_class ) {
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -218,17 +218,17 @@ class Two_Factor_Core {
 	}
 
 	/**
-	 * Get the classnames for all registered providers.
+	 * Get the classnames for specific providers.
 	 *
-	 * Note some of these providers might not be enabled.
+	 * @param array $providers List of paths to provider class files indexed by class names.
 	 *
 	 * @return array List of provider keys and classnames.
 	 */
-	private static function get_providers_classes() {
-		$providers = self::get_providers_registered();
-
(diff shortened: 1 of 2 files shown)


class-two-factor-core.php
```diff
@@ -204,6 +204,9 @@ public static function get_providers_registered() {
 		/**
 		 * Filter the supplied providers.
 		 *
+		 * This lets third-parties either remove providers (such as Email), or
```
ocean90, line 207: You can't remove providers here since the default ones will always be merged in line 217, right? This will likely break implementations like https://github.com/WordPress/wporg-two-factor/blob/fe0a981432ecd52f6b74a31cd2c7d26dd761e3df/wporg-two-factor.php#L81-L93

kasparsd: Indeed.

This `get_providers_registered()` is intended mostly for the uninstall logic so it is attempting to collect every possible provider that could have been registered at any point, or any overrides.

I'll think about this some more.

class-two-factor-core.php
```diff
@@ -149,7 +149,22 @@ public static function uninstall() {
...
+		 *                         the value is the path to the file containing the class.
+		 */
+		$additional_providers = apply_filters( 'two_factor_providers', $providers );
+
+		// Merge them with the default providers.
+		if ( ! empty( $additional_providers ) ) {
+			$providers = array_merge( $providers, $additional_providers );
+		}
+
+		foreach ( self::get_providers_classes( $providers ) as $provider_class ) {
```
kasparsd, line 167: Here we still pass the provider classes through the filters so any provider adjustments are accounted for.

class-two-factor-core.php
```diff
@@ -287,15 +287,13 @@ public static function get_providers() {
...
 
 		foreach ( $providers as $provider_key => $provider_class ) {
-			if ( method_exists( $provider_class, 'get_instance' ) ) {
-				try {
-					$providers[ $provider_key ] = call_user_func( array( $provider_class, 'get_instance' ) );
-				} catch ( Exception $e ) {
-					unset( $providers[ $provider_key ] );
-				}
+			try {
+				$providers[ $provider_key ] = call_user_func( array( $provider_class, 'get_instance' ) );
```
kasparsd, line 294: This will fail both when (1) `get_instance` is not defined for the class and (2) if that methods triggers some other error.

class-two-factor-core.php
kasparsd: This is being fixed in #651.

class-two-factor-core.php
```diff
@@ -192,43 +207,28 @@ public static function uninstall() {
 	 *
 	 * @return array List of provider keys and paths to class files.
 	 */
-	public static function get_providers_registered() {
```
kasparsd, line 195: TODO: consider that an integration could have been using this already. Might need to keep the name or add a helper method.

kasparsd: This was added only in the previous version so we don't have to worry about that.

tests/class-two-factor-core.php
```diff
@@ -1581,4 +1581,56 @@ public function test_uninstall_removes_user_meta() {
...
+		$this->assertContains( 'Two_Factor_Email', array_keys( Two_Factor_Core::get_providers() ), 'Email provider is enabled by default' );
+
+		add_filter(
+			'two_factor_providers',
+			function( $providers ) {
+				return array_diff_key( $providers, array( 'Two_Factor_Email' => null ) );
+			}
+		);
+
+		$this->assertNotContains( 'Two_Factor_Email', array_keys( Two_Factor_Core::get_providers() ), 'Default provider can be disabled via a filter' );
```
kasparsd, line 1595: @ocean90 @r-a-y Here is a unit test to confirm that a core provider can be disabled via the filter.

kasparsd:
This has been revised and is ready for code review. I've also added unit tests to confirm that it fixes the issue and allows removing core providers.

Would appreciate a code review.

class-two-factor-core.php
```diff
@@ -149,7 +149,22 @@ public static function uninstall() {
...
-		foreach ( self::get_providers_classes() as $provider_class ) {
+		$providers = self::get_default_providers();
+
+		/**
+		 * Include additional providers to be uninstalled.
+		 *
+		 * @param array $providers A key-value array where the key is the class name, and
+		 *                         the value is the path to the file containing the class.
+		 */
+		$additional_providers = apply_filters( 'two_factor_providers', $providers );
```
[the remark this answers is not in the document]
r-a-y: I would remove the PHPDoc and do what WordPress does when the same filter is used more than twice:

```
/** This filter is documented in the get_providers() method */
```

r-a-y:
@kasparsd - Thanks for your work on this. I can confirm that it is once again possible to diable a core provider with the `'two_factor_providers'` filter.

class-two-factor-core.php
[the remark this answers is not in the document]
kasparsd: Thanks for the suggestion! This has been implemented in 604ff9f034957dfcf571d200d1dc4e7cb602d6d4.

PR #653, opened 2024-12-02
Add filters for email token and backup code length

### LINKED ISSUE #374: Backup Codes (Improve 8 to 12 numbers)
It would be nice to improve the "Backup Codes" option from 8 numbers to 12 numbers.

Example: 97997695 > 520184887438

These should probably mean than in a online attack scenario (assuming one thousand guesses per second) it should jump from 1.29 days for 8 numbers, to 35.33 years for the 12 numbers.

I think everyone would feel a little less concern with random online guessing with these small improvement.

For human usability maybe use dashes between numbers: 520184887438 would look: 5201-8488-7438

kasparsd:
Fixes #374.

Replaces #419.

## What?

Implements two new filters:

- `two_factor_email_token_ttl` filter overrides the default 8 character count for email tokens. This has been renamed from `two_factor_token_ttl` which is now marked as deprecated.

- `two_factor_backup_code_length` filter overrides the default 8 character count for backup codes. Providers the `WP_User` of the associated user as the second argument.

## Why?

Users might want to increase this.

## How?

- Introduce the filters and rename existing ones for consistency. Add tests to confirm both old and new filters are working.

- Updated the input field placeholders to match the expected character length. Note that this might give away the expected character length for a brute force attacker but we're rate limiting these attempts so that is not an issue.

## Screenshots or screencast

Note the updated `XXXX` characters for the placeholders and the correct length. Previously the placeholder was also including spaces but that could have confused the users about the format not matching what they get in the email or app. The input automatically removes invalid characters so the processing happens in either way.

email-codes

recover-codes

Inspiration:

input-format

## Changelog Entry

Added - Introduce `two_factor_email_token_length` and `two_factor_backup_code_length` filters for adjusting the length of the respective tokens.

### DIFF
--- a/providers/class-two-factor-backup-codes.php
+++ b/providers/class-two-factor-backup-codes.php
@@ -239,8 +239,16 @@ class Two_Factor_Backup_Codes extends Two_Factor_Provider {
 			$codes_hashed = (array) get_user_meta( $user->ID, self::BACKUP_CODES_META_KEY, true );
 		}
 
+		/**
+		 * Customize the character count of the backup codes.
+		 *
+		 * @var int $code_length Length of the backup code.
+		 * @var WP_User $user User object.
+		 */
+		$code_length = (int) apply_filters( 'two_factor_backup_code_length', 8, $user );
+
 		for ( $i = 0; $i < $num_codes; $i++ ) {
-			$code           = $this->get_code();
+			$code           = $this->get_code( $code_length );
 			$codes_hashed[] = wp_hash_password( $code );
 			$codes[]        = $code;
 			unset( $code );
--- a/providers/class-two-factor-backup-codes.php
+++ b/providers/class-two-factor-backup-codes.php
@@ -214,6 +214,13 @@ class Two_Factor_Backup_Codes extends Two_Factor_Provider {
 		<?php
 	}
 
+	/**
+	 * Get the backup code length for a user.
+	 *
+	 * @param WP_User $user User object.
+	 *
+	 * @return int Number of characters.
+	 */
 	private function get_backup_code_length( $user ) {
 		/**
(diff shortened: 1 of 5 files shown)


kasparsd:
This is ready for code review.

providers/class-two-factor-email.php
```diff
@@ -142,14 +144,25 @@ public function user_token_lifetime( $user_id ) {
...
+		/**
+		 * Number of seconds the token is considered valid
+		 * after the generation.
+		 *
+		 * @deprecated 0.11.0 Use {@see 'two_factor_email_token_ttl'} instead.
+		 *
+		 * @param integer $token_ttl Token time-to-live in seconds.
+		 * @param integer $user_id User ID.
+		 */
+		$token_ttl = (int) apply_filters_deprecated( 'two_factor_token_ttl', array( $token_ttl, $user_id ), '0.7.0', 'two_factor_email_token_ttl' );
```
kasparsd, line 156: Decided to rename this to `two_factor_email_token_ttl` for clarity and consistency.

providers/class-two-factor-backup-codes.php
```diff
@@ -326,11 +347,14 @@ public static function codes_remaining_for_user( $user ) {
 	 */
 	public function authentication_page( $user ) {
 		require_once ABSPATH . '/wp-admin/includes/template.php';
+
+		$code_placeholder = str_repeat( 'X', $this->get_backup_code_length( $user ) );
+
 		?>
-		<p class="two-factor-prompt"><?php esc_html_e( 'Enter a recovery code.', 'two-factor' ); ?></p><br/>
```
kasparsd, line 330: This was causing too much blank space between the message and the input:

recover-codes

PR #656, opened 2024-12-03
Full RFC6238 Compatibility

ericmann:
## What?

Adds support for SHA256 and SHA512 hashes for the underlying TOTP algorithm to fully support the RFC6238 spec.

Replaces #207 for a cleaner GH expereince.

## Why?

Up til now, the TOTP provider in this plugin (and most other PHP implementations) claims to support a specified hash type, but doesn't actually work with anything other than SHA1.

This is due to key lengths and hash lengths being different for the three hash variants.

## How?

This change introcudes support for both SHA256 and SHA512, porting the implementation directly from https://github.com/ericmann/totp.

See https://tools.ietf.org/html/rfc6238#section-1.2 for more information on the MAY USE notation for SHA256 and SHA512.

See https://tools.ietf.org/html/rfc6238#appendix-A for a fully compliant reference implementation in Java.

## Testing Instructions

See https://tools.ietf.org/html/rfc6238#appendix-B for test vectors showing TOTPs generated for specific time values and the three hash variants.

See https://github.com/ericmann/totp/blob/master/test/phpunit/ReferenceTest.php for example unit tests verifying this particular implementation before it was ported to the plugin.

### DIFF
--- a/providers/class-two-factor-totp.php
+++ b/providers/class-two-factor-totp.php
@@ -72,7 +72,7 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 	 *
 	 * @param int $now Timestamp to use when overriding time().
 	 */
-	public static function __set_time( $now ) {
+	public static function set_time( $now ) {
 		self::$now = $now;
 	}
 
--- a/providers/class-two-factor-totp.php
+++ b/providers/class-two-factor-totp.php
@@ -529,11 +529,13 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 	 *
 	 * @param string $key      The share secret key to use.
 	 * @param string $authcode The code to test.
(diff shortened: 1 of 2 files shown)


providers/class-two-factor-totp.php
```diff
@@ -572,11 +601,13 @@ public static function get_authcode_valid_ticktime( $key, $authcode ) {
 		$ticks = range( - $max_ticks, $max_ticks );
 		usort( $ticks, array( __CLASS__, 'abssort' ) );
 
-		$time = floor( time() / self::DEFAULT_TIME_STEP_SEC );
+		$time = floor( self::time() / self::DEFAULT_TIME_STEP_SEC );
```
ericmann:
```suggestion
		$time = floor( self::time() / $time_step );
```

PR #657, opened 2025-01-06
Add autocomplete "one-time-code" attribute

stefanmomm:
## What?
This PR disables autocompletion for authcode input fields. 

## Why?
Since authcode are always unique, autocompletion is not useful here.

## Changelog Entry
Added - Disabled autocompletion for authcode input fields.
Fixed - Corrected input type.

### DIFF
--- a/providers/class-two-factor-totp.php
+++ b/providers/class-two-factor-totp.php
@@ -682,7 +682,7 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 		</p>
 		<p>
 			<label for="authcode"><?php esc_html_e( 'Authentication Code:', 'two-factor' ); ?></label>
-			<input type="text" inputmode="numeric" autocomplete="one-time-code" name="authcode" id="authcode" class="input authcode" value="" size="20" pattern="[0-9 ]*" placeholder="123 456" autocomplete="off" data-digits="<?php echo esc_attr( self::DEFAULT_DIGIT_COUNT ); ?>" />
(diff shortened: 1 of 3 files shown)


providers/class-two-factor-totp.php
```diff
@@ -349,7 +349,7 @@ public function user_two_factor_options( $user ) {
 						/* translators: Example auth code. */
 						$placeholder = sprintf( __( 'eg. %s', 'two-factor' ), '123456' );
 					?>
-					<input type="tel" name="two-factor-totp-authcode" id="two-factor-totp-authcode" class="input" value="" size="20" pattern="[0-9 ]*" placeholder="<?php echo esc_attr( $placeholder ); ?>" />
+					<input type="text" inputmode="numeric" name="two-factor-totp-authcode" id="two-factor-totp-authcode" class="input" value="" size="20" pattern="[0-9 ]*" placeholder="<?php echo esc_attr( $placeholder ); ?>" autocomplete="off" />
```
kasparsd, line 352: I just confirmed that GitHub is also using `text` input type with `numeric` inputmode:

totp-input-github

But they also have `autocomplete="one-time-code"` ([per spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#one-time-code)), so I wonder if we actually need to change that to match.

kasparsd (commented):
I reviewed a few of the industry examples like GitHub and it appears that there is an update spec for the `autocomplete` values so we could use those instead:

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#one-time-code

Any concerns with this approach?

stefanmomm:
@kasparsd Looks good to me, I just changed `autocomplete` to `one-time-code`.

kasparsd (approved):
This is great, thank you!

PR #660, opened 2025-01-10
Trigger two-factor flow only when expected

### LINKED ISSUE #659: Clearing auth cookies when visiting wp-login.php while being already logged-in
### Describe the bug

When a user who is logged-in (and has two-factor configured) visits the wp-login.php page, their authentication state is cleared. The clearing should happen only during the two-factor login workflow.

By default, WP core redirects users to the dashboard if they visit wp-login.php while being already logged in. This is not happening and prevents other plugins that rely on the logged-in state on the wp-login.php page from functioning.

This happens because the function attached to `wp_login` is running the logic even outside of the login flow:

https://github.com/WordPress/two-factor/blob/80e76efc69a3fafb937f6b050772be3522398109/class-two-factor-core.php#L641-L653

### Steps to Reproduce

1. Login at a site.
2. Configure and enable any two-factor method.
3. Visit /wp-login.php and notice that a login form is shown or just the two-factor prompt.

Image

### Environment information

- Two-factor 0.11.0
- Default theme 2025
- No other plugins

Yes

Yes

kasparsd:
## What?

Avoid triggering the two-factor logic on every `wp_login` hook even when no login attempt is expected.

## Why?

Fixes #659.

## How?

WP core uses the following filters during the `wp-login.php` request in the order of priority:

    // Priority 20.
    add_filter( 'authenticate', 'wp_authenticate_username_password', 20, 3 );
    add_filter( 'authenticate', 'wp_authenticate_email_password', 20, 3 );
    add_filter( 'authenticate', 'wp_authenticate_application_password', 20, 3 );

    // Priority 30.
    add_filter( 'authenticate', 'wp_authenticate_cookie', 30, 3 );

    // Priority 99.
    add_filter( 'authenticate', 'wp_authenticate_spam_check', 99 );

We want to trigger the two-factor workflow *only for requests that*:

1. supply username and/or password,
2. and don't have existing and valid user cookies.

So we hook right after `wp_authenticate_cookie` priority 30.

## Testing Instructions

1. Setup a user with one of two-factor methods enabled.
2. Attempt to login in a new private browser session and confirm that the second factor is requested.
3. After a successful login, visit `wp-login.php` and confirm that you're redirected to the admin dashboard.

## Changelog Entry

Fixed -- ensure that two-factor workflow is triggered only during login attempts.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -682,17 +682,11 @@ class Two_Factor_Core {
 	 * @return WP_User|WP_Error
 	 */
 	public static function filter_authenticate( $user ) {
-		if ( $user instanceof WP_User && self::is_user_using_two_factor( $user->ID ) ) {
-			// Trigger the second-factor flow if the password was correct.
-			add_action( 'wp_login', array( __CLASS__, 'wp_login' ), 10, 2 );
-
-			// Disable the XML-RPC and REST API for users with two-factor enabled.
-			if ( self::is_api_request() && ! self::is_user_api_login_enabled( $user->ID ) ) {
-				return new WP_Error(
-					'invalid_application_credentials',
-					__( 'Error: API login for user disabled.', 'two-factor' )
-				);
-			}
+		if ( $user instanceof WP_User && self::is_api_request() && self::is_user_using_two_factor( $user->ID ) && ! self::is_user_api_login_enabled( $user->ID ) ) {
+			return new WP_Error(
+				'invalid_application_credentials',
+				__( 'Error: API login for user disabled.', 'two-factor' )
+			);
 		}
 
 		return $user;
@@ -715,6 +709,7 @@ class Two_Factor_Core {
 		 * rather than through an unsupported 3rd-party login process which this plugin doesn't support.
 		 */
 		if ( $user instanceof WP_User && self::is_user_using_two_factor( $user->ID ) && did_action( 'login_init' ) ) {
+			add_action( 'wp_login', array( __CLASS__, 'wp_login' ), 10, 2 );
 			add_filter( 'send_auth_cookies', '__return_false', PHP_INT_MAX );
 		}
 
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -699,7 +699,7 @@ class Two_Factor_Core {
 			}
 
 			// Trigger the second-factor flow for valid users.
-			add_action( 'wp_login', array( __CLASS__, 'wp_login' ), 10, 2 );
+			add_action( 'wp_login', array( __CLASS__, 'wp_login' ), PHP_INT_MAX, 2 );
 		}
 
 		return $user;
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -116,8 +116,13 @@ class Two_Factor_Core {
 		add_action( 'set_auth_cookie', array( __CLASS__, 'collect_auth_cookie_tokens' ) );
 		add_action( 'set_logged_in_cookie', array( __CLASS__, 'collect_auth_cookie_tokens' ) );
 
-		// Run only after the core wp_authenticate_username_password() check.
-		add_filter( 'authenticate', array( __CLASS__, 'filter_authenticate' ), 50 );
+		/**
+		 * Enable the two-factor workflow only for login requests with username
+		 * and without an existing user cookie.
+		 *
+		 * Run after core username/password and cookie checks.
+		 */
+		add_filter( 'authenticate', array( __CLASS__, 'filter_authenticate' ), 31, 3 );
 
 		// Run as late as possible to prevent other plugins from unintentionally bypassing.
 		add_filter( 'authenticate', array( __CLASS__, 'filter_authenticate_block_cookies' ), PHP_INT_MAX );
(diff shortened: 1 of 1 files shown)


class-two-factor-core.php
```diff
@@ -117,8 +116,13 @@ public static function add_hooks( $compat ) {
...
 
-		// Run only after the core wp_authenticate_username_password() check.
-		add_filter( 'authenticate', array( __CLASS__, 'filter_authenticate' ), 50 );
+		/**
+		 * Enable the two-factor workflow only for login requests with username
+		 * and without an existing user cookie.
+		 *
+		 * Run after core username/password and cookie checks.
+		 */
+		add_filter( 'authenticate', array( __CLASS__, 'filter_authenticate' ), 31, 3 );
```
kasparsd, line 125: Priority 31 is still after `wp_authenticate_username_password()` so the resolved `$user` will be set correctly.

class-two-factor-core.php
```diff
@@ -673,21 +677,30 @@ public static function destroy_current_session_for_user( $user ) {
...
 	 * @return WP_User|WP_Error
 	 */
-	public static function filter_authenticate( $user ) {
-		if ( $user instanceof WP_User && self::is_api_request() && self::is_user_using_two_factor( $user->ID ) && ! self::is_user_api_login_enabled( $user->ID ) ) {
-			return new WP_Error(
-				'invalid_application_credentials',
-				__( 'Error: API login for user disabled.', 'two-factor' )
-			);
+	public static function filter_authenticate( $user, $username, $password ) {
+		if ( strlen( $username ) && $user instanceof WP_User && self::is_user_using_two_factor( $user->ID ) ) {
```
kasparsd, line 693: The main difference here is that we're now triggering any of this logic only for requests with the `$username` supplied which indicates a login attempt instead of a simple authenticated request to `wp-login.php`.

class-two-factor-core.php
```diff
@@ -673,21 +677,30 @@ public static function destroy_current_session_for_user( $user ) {
...
+			// Disable authentication requests for API requests for users with two-factor enabled.
+			if ( self::is_api_request() && ! self::is_user_api_login_enabled( $user->ID ) ) {
+				return new WP_Error(
+					'invalid_application_credentials',
+					__( 'Error: API login for user disabled.', 'two-factor' )
+				);
+			}
+
+			// Trigger the two-factor flow only for login attempts.
+			add_action( 'wp_login', array( __CLASS__, 'wp_login' ), PHP_INT_MAX, 2 );
```
kasparsd, line 703: Another major difference here is that we're now triggering the `wp_login` much later to avoid any other integration plugins to do their job before we completely hijack the remaining login flow.

The only risk here is that some other plugin hijacks the login flow before us and issues a redirect, for example.

I add this sample hook:

```php
add_action(
	'wp_login',
	function ( $user ) {
		if ( $user ) {
			wp_redirect( admin_url( '?redirect=test' ) );
			exit;
		}
	}
);
```

which returns a redirect before our `wp_login`, and the login session correctly fails because we're blocking the login cookies through `filter_authenticate_block_cookies()` attached to the `authenticate` hook which always fires before `wp_login`.

no-cookies-on-redirect

kasparsd (commented):
This is ready for code review. Please see the comments inline the pull request for additional details.

kasparsd:
@dd32 Could this impact the WP.org implementation in any way?

dd32:
> @dd32 Could this impact the WP.org implementation in any way?

Definately could, changing when cookies are sent could affect other things we have hooked in to prevent cookies under specific situations (ie. Other login interstitials). If it could affect us, it could affect other plugins that are hooked into the authentication processes.

I won't be able to test this, or properly review it, for awhile though.

jeffpaul:
> Definately could, changing when cookies are sent could affect other things we have hooked in to prevent cookies under specific situations (ie. Other login interstitials). If it could affect us, it could affect other plugins that are hooked into the authentication processes.

Bah, that's not great.  The approach in this PR does seem more like the "right" way of doing things, though impacting how others might be extending the work of the plugin is not great.  Perhaps we hold this for `0.13.0` and message in the `0.12.0` release that this change is coming and to look into this PR specifically for anyone concerned with the change planned for `0.13.0`?

dd32:
I want to suggest it's a safe change, but without testing it I can't really confirm, it's just that the entire login flow is rather weird and ill-designed for this use-case while it preserves back-compat.. I've run into weirdness a multitude of times both where I've said `Why is two-factor working like this?!` and `What? Why is Core.. wheres the filter.. ugh fine..`

I'm not a fan of putting any form of warning in release notes for future changes, because the people who would need to see it never do. I'd just make the change.

kasparsd:
@dd32 Would it be possible for you to test this on any kind of environment?

The only regression we're trying to avoid is the complete lack of second factor workflow. Honestly, I'm not even sure what kind of setup could trigger that because as shown in the example above -- any integration doing something strange during `wp_login` will simply disable ALL login because we're silencing the login cookies during `authenticate`.

WP-org and WPVIP are probably our two largest users so any testing on either of implementations would be helpful.

@sjinks Could you help with getting this tested on a baseline VIP setup?

Issue #663, opened 2025-02-13
Avoid deprecated marking a parameter as nullable in PHP 8.4


BrookeDot:
### Describe the bug

[PHP 8.4 deprecated marking a parameter as nullable](https://www.php.net/manual/en/migration84.deprecated.php). For sites running Two Factor a PHP error shows up in the logs:

> u2flib_server\Error::__construct(): Implicitly marking parameter $previous as nullable is deprecated, the explicit nullable type must be used instead

**Stacktrace**

```
    wp-content/plugins/two-factor/includes/Yubico/U2F.php:504
    Two_Factor_FIDO_U2F->__construct()
    wp-content/plugins/two-factor/providers/class-two-factor-provider.php:28
    Two_Factor_Provider::get_instance()
    wp-content/plugins/two-factor/class-two-factor-core.php:291
    Two_Factor_Core::get_providers()
    wp-includes/class-wp-hook.php:324
    do_action('init')
    wp-settings.php:704
```

### Steps to Reproduce

1. Install PHP 8.4, Two Factor and optionally Query Monitor but ensure that logging is enabled
2. Notice the warning in the logs

### Environment information

PHP 8.4.2 
WordPress 6.7.2

### Please confirm that you have searched existing issues in this repository.

Yes

### Please confirm that you have tested with all plugins deactivated except Two-Factor.

Yes

BrookeDot:
I can open a PR but think we need to update line 504 to:

```
    public function __construct($message, $code, $previous = null) {
```

Otherwise if we can bump the PHP version:

```
    public function __construct($message, $code, \Exception|null $previous = null) {
```

Both fix the problem with PHP 8.4 so looking for input on the best way forward here. 

Line with warning:https://github.com/WordPress/two-factor/blob/master/includes/Yubico/U2F.php#L504

kasparsd:
Fixed in #664.

BrookeDot:
@felipeelia pointed out when working on the same problem for ElasticPress (https://github.com/10up/ElasticPress/issues/4086) that we can use `?` which is [supported in PHP 7.1](https://www.php.net/manual/en/migration71.new-features.php) so I went with that here. Thanks for the quick merge @kasparsd

PR #664, opened 2025-02-13
Fix PHP 8.4 Implicitly marking parameter $previous as nullable is deprecated

BrookeDot:
## What?
PHP 8.4 requires `null` values to be explicitly declared: https://www.php.net/manual/en/migration84.deprecated.php

This PR does that for the U2F.php class fixing the warning:
> u2flib_server\Error::__construct(): Implicitly marking parameter $previous as nullable is deprecated, the explicit nullable type must be used instead

## Why?
To avoid noise in the PHP logs and it is likely this may be a fatal in a future version of PHP.

## How?
This adds a `?` to the class to allow it to be explicitly null.

## Testing Instructions
Install PHP 8.4, make the change in the PR and watch the error go away.

## Changelog Entry
Fixed - PHP 8.4: Implicitly marking parameter $previous as nullable is deprecated (#663)

### DIFF
--- a/includes/Yubico/U2F.php
+++ b/includes/Yubico/U2F.php
@@ -501,7 +501,7 @@ class Error extends \Exception
      * @param int $code
      * @param \Exception|null $previous
      */
-    public function __construct($message, $code, \Exception $previous = null) {
+    public function __construct($message, $code, ?\Exception $previous = null) {
         parent::__construct($message, $code, $previous);
     }
 }


kasparsd:
The plugin is currently supporting PHP 7.2 and above:

https://github.com/WordPress/two-factor/blob/953b7c98a51d3ab2f3d2c6c25f007f626be4b7de/two-factor.php#L16C23-L16C26

It appears that support for nullable arguments [was added to 7.1](https://www.php.net/manual/en/migration71.new-features.php) so this is safe to merge. 

Thanks for the pull request!

Noting that U2F support will be removed soon #439 but having this fixed is definitely worth it.

Issue #662, opened 2025-01-15
Profile settings section rendered regardless of whether any providers are enabled


Player701:
### Describe the bug

If there are no two-factor providers enabled, the corresponding profile section is still rendered on the user settings page. This creates a visual inconsistency in scenarios where two-factor authentication needs to be disabled for a specific user or group of users.

**Expected behavior:** Users should not see the two-factor section in their profile settings if there are no providers currently available.

**Actual behavior:** Users always see the two-factor section in their profile settings even if there are no providers currently available. (See screenshot)

### Steps to Reproduce

This filter will disable all providers:

```
add_filter(
    'two_factor_providers',
    function ( $providers ) {
        return [];
    },
    9999
);
```

Of course this code is for demonstration purposes only. Actual code would likely use `two_factor_enabled_providers_for_user` and perform some check on the user ID first. The result, however, is the same.

### Screenshots, screen recording, code snippet

![Image](https://github.com/user-attachments/assets/f8ad78f8-c2b6-4910-afda-19280cc65d8a)

### Environment information

WordPress 6.7.1, plugin version 0.11.0

### Please confirm that you have searched existing issues in this repository.

Yes

### Please confirm that you have tested with all plugins deactivated except Two-Factor.

Yes

jeffpaul:
@Player701 in what scenario would no providers be enabled and the plugin activated?

Player701:
> [@Player701](https://github.com/Player701) in what scenario would no providers be enabled and the plugin activated?

An example scenario would be to restrict the usage of two-factor authentication to admins, or another specific group of users.

If you want to know about my particular use case:

I have only a few users on my website, but they are not power-users, and the two-factor settings are simply too confusing to them. Most of them don't even know what "two-factor" is... Additionally, all of those users browse on mobile devices, but the layout of the two-factor settings is currently partially broken on mobile and also takes up a lot of screen space.

So, believe it or not - this has actually resulted in one of the users complaining to me about their profile not updating! It turned out they somehow couldn't find the submit button at the very bottom because of all that clutter added by the plugin. Don't ask me how, it just happened...

Therefore, I've been considering to disable these settings for normal users entirely, especially taking into account that my site does not handle any security-sensitive data (e.g. payment cards and such). However, as an admin who has full management access, I still want to keep using the plugin myself.

kasparsd:
The approach in #669 is leaving the section present but hiding the table. Please let me know if this approach works.

[...]

PR #669, opened 2025-02-14
Add a filter to limit two-factor providers available to each user

### LINKED ISSUE #647: Allow to limit providers per user

I'd like to limit the available providers per user. Example code:

```php
/**
 * Limit the possible providers to email and time-based one-time password.
 */
add_filter(
	'two_factor_providers',
	static fn( array $providers ): array =>
		array_intersect_key(
			$providers,
			[
				'Two_Factor_Email' => '',
				'Two_Factor_Totp'  => '',
			]
		)
);

/**
 * Require two factor authentication via email for all users without the manage_extended_two_factor capability.
 */
add_filter(
	'two_factor_enabled_providers_for_user',
	static fn( $providers, $user_id ) =>
		user_can( $user_id, 'manage_extended_two_factor' ) ? $providers : [ 'Two_Factor_Email' ],
	10,
	2
);
add_filter(
	'two_factor_primary_provider_for_user',
	static fn( $provider, $user_id ) =>
		user_can( $user_id, 'manage_extended_two_factor' ) ? $provider : 'Two_Factor_Email',
	10,
	2
);
```

While this does prevent users without the `manage_extended_two_factor` capability to use the TOTP provider, the UI still renders the TOTP option.

Image

That's because the table uses the `Two_Factor_Core::get_providers()` method to render the list of providers which isn't user-specific.

https://github.com/WordPress/two-factor/blob/6a95e7ffdebb2e0408b4ccb298e9df16d5b6998c/class-two-factor-core.php#L1837

### Proposed Solution

I'm not sure if `Two_Factor_Core::get_providers()` should get an argument for a user or if there should be a wrapper for `Two_Factor_Core::get_providers()` like `Two_Factor_Core::get_providers_for_user( $user )`.

Looking at the current usage of `Two_Factor_Core::get_providers()`, in all cases we'd have a user ID available.

### Designs

_No response_

_No response_

Yes

kasparsd:
## What?

1. Add a filter to allow limiting the available providers per user.
2. Skip rendering the provider table if no providers are available for a user.

## Why?

Fixes #647, #662.

## How?

1. Introduce a new method `get_supported_providers_for_user( $user )` which wraps the output of `get_providers()` in a filter `two_factor_providers_for_user`.
2. Use the new helper whenever we're looking for two-factor methods that should be available to a user.

Note that we already similar methods:

- `get_enabled_providers_for_user( $user = null )` for the provider _keys_ that are enabled in the user profile **but might not be configured, yet**.
- `get_available_providers_for_user( $user = null )` for the provider _keys_ that **are both enabled and configured**.

so I chose the name `supported` to designate if a provider is available to a specific user.

## Testing Instructions

Unit tests have been added to cover the filtering scenario. There are no user settings to toggle this.

## Screenshots or screencast

If all providers are disabled for a user we now display a notice instead of an empty table:

two-factor-options-no-providers

## Changelog Entry

Add a filter `two_factor_providers_for_user` to disable providers for specific users.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1879,11 +1879,15 @@ class Two_Factor_Core {
 		</div>
 		<?php endforeach; ?>
 
+		<fieldset id="two-factor-options" <?php echo $show_2fa_options ? '' : 'disabled="disabled"'; ?>>
 		<?php
 		if ( $providers ) {
 			self::render_user_providers_form( $user, $providers );
 		}
+		?>
+		</fieldset>
 
+		<?php
 		/**
 		 * Fires after the Two Factor methods table.
 		 *
@@ -1906,7 +1910,6 @@ class Two_Factor_Core {
 			<?php esc_html_e( 'Configure a primary two-factor method along with a backup method, such as Recovery Codes, to avoid being locked out if you lose access to your primary method.', 'two-factor' ); ?>
 		</p>
 
-		<fieldset id="two-factor-options" <?php echo $show_2fa_options ? '' : 'disabled="disabled"'; ?>>
 		<?php wp_nonce_field( 'user_two_factor_options', '_nonce_user_two_factor_options', false ); ?>
 		<input type="hidden" name="<?php echo esc_attr( self::ENABLED_PROVIDERS_USER_META_KEY ); ?>[]" value="<?php /* Dummy input so $_POST value is passed when no providers are enabled. */ ?>" />
 
@@ -1957,7 +1960,6 @@ class Two_Factor_Core {
 				</tr>
 			</tbody>
 		</table>
-		</fieldset>
 		<?php
 	}
 
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -298,6 +298,28 @@ class Two_Factor_Core {
 	}
 
 	/**
+	 * Get providers available for user which may not be enabled or configured.
+	 *
+	 * @see Two_Factor_Core::get_enabled_providers_for_user()
+	 * @see Two_Factor_Core::get_available_providers_for_user()
+	 *
+	 * @param  WP_User|int|null $user User ID.
+	 * @return array List of provider instances indexed by provider key.
+	 */
+	public function get_supported_providers_for_user( $user = null ) {
+		$user = self::fetch_user( $user );
+		$providers = self::get_providers();
+
+		/**
+		 * List of providers available to user which may not be enabled or configured.
+		 *
+		 * @param array       $providers List of available provider instances indexed by provider key.
+		 * @param int|WP_User $user User ID.
+		 */
+		return apply_filters( 'two_factor_providers_for_user', $providers, $user );
+	}
+
+	/**
 	 * Enable the dummy method only during debugging.
 	 *
 	 * @param array $methods List of enabled methods.
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -309,7 +309,7 @@ class Two_Factor_Core {
 	 * @param  WP_User|int|null $user User ID.
 	 * @return array List of provider instances indexed by provider key.
 	 */
-	public function get_supported_providers_for_user( $user = null ) {
+	public static function get_supported_providers_for_user( $user = null ) {
 		$user = self::fetch_user( $user );
 		$providers = self::get_providers();
 
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1841,12 +1841,11 @@ class Two_Factor_Core {
 		wp_enqueue_style( 'user-edit-2fa', plugins_url( 'user-edit.css', __FILE__ ), array(), TWO_FACTOR_VERSION );
 
 		$enabled_providers = array_keys( self::get_available_providers_for_user( $user ) );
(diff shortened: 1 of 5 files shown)


class-two-factor-core.php
```diff
@@ -249,9 +249,12 @@ private static function get_providers_classes( $providers ) {
 	}
 
 	/**
-	 * Get all enabled two-factor providers with keys as the original
+	 * Get all registered two-factor providers with keys as the original
 	 * provider class names and the values as the provider class instances.
 	 *
+	 * @see Two_Factor_Core::get_enabled_providers_for_user()
```
kasparsd, line 255: Cross-linked the related methods for easier discovery and comparison.

class-two-factor-core.php
```diff
@@ -297,6 +300,28 @@ public static function get_providers() {
...
+	/**
+	 * Get providers available for user which may not be enabled or configured.
+	 *
+	 * @see Two_Factor_Core::get_enabled_providers_for_user()
+	 * @see Two_Factor_Core::get_available_providers_for_user()
+	 *
+	 * @param  WP_User|int|null $user User ID.
+	 * @return array List of provider instances indexed by provider key.
+	 */
+	public static function get_supported_providers_for_user( $user = null ) {
```
kasparsd, line 312: This is the new method and the associated filter.

class-two-factor-core.php
```diff
@@ -1804,34 +1839,35 @@ public static function manage_users_custom_column( $output, $column_name, $user_
...
+				esc_url( $url )
 			);
 		}
 
-		printf(
-			'<fieldset id="two-factor-options" %s>',
-			$show_2fa_options ? '' : 'disabled="disabled"'
-		);
+		if ( empty( $providers ) ) {
+			$notices['notice two-factor-notice-no-providers-supported'] = esc_html__( 'No providers are available for your account.', 'two-factor' );
```
kasparsd, line 1866: Display a notice if no providers are available for user. 

I'm assuming that with the plugin enabled it would be more confusing to hide the section completely. With the section being present users can at least know the option should be available and is simply disabled.

class-two-factor-core.php
```diff
@@ -1804,34 +1839,35 @@ public static function manage_users_custom_column( $output, $column_name, $user_
...
+		}
 
-		if ( 1 === count( $enabled_providers ) ) {
+		// Suggest enabling a backup method if only method is enabled and there are more available.
+		if ( count( $providers ) > 1 && 1 === count( $enabled_providers ) ) {
```
kasparsd, line 1870: Display this suggestion only if at least two methods are available.

class-two-factor-core.php
```diff
@@ -1842,15 +1878,44 @@ public static function user_two_factor_options( $user ) {
 			<p><?php echo wp_kses_post( $notice ); ?></p>
 		</div>
 		<?php endforeach; ?>
+
+		<fieldset id="two-factor-options" <?php echo $show_2fa_options ? '' : 'disabled="disabled"'; ?>>
+		<?php
+		if ( $providers ) {
+			self::render_user_providers_form( $user, $providers );
```
kasparsd, line 1885: To avoid wrapping a large block into this conditional, I've moved the form to a helper method.

providers/class-two-factor-fido-u2f-admin.php
```diff
@@ -164,6 +164,10 @@ protected static function asset_version() {
 	 * @param WP_User $user WP_User object of the logged-in user.
 	 */
 	public static function show_user_profile( $user ) {
+		if ( ! Two_Factor_FIDO_U2F::is_supported_for_user( $user ) ) {
```
kasparsd, line 167: Avoid rendering the FIDO U2F settings if the provider is not "supported".

kasparsd (commented):
@ocean90 Would this approach provide you with the required filtering as described in #647?

class-two-factor-core.php
jeffpaul: I agree, perhaps iterating on the copy to provide a user with some insight on what to do next?  Eg. "No providers are available for your account.  Please contact the site administrator for more details." or whatever verbage is used elsewhere in core to point folks to the admin (either site or network)?

kasparsd:
@ocean90 Can you please review the implementation and confirm if it matches your requirements?

class-two-factor-core.php
```diff
@@ -1820,18 +1856,18 @@ public static function user_two_factor_options( $user ) {
...
-		printf(
-			'<fieldset id="two-factor-options" %s>',
-			$show_2fa_options ? '' : 'disabled="disabled"'
-		);
+		if ( empty( $providers ) ) {
+			$notices['notice two-factor-notice-no-providers-supported'] = esc_html__( 'No providers are available for your account.', 'two-factor' );
+		}
 
-		if ( 1 === count( $enabled_providers ) ) {
+		// Suggest enabling a backup method if only method is enabled and there are more available.
```
ocean90, line 1869:
```suggestion
		// Suggest enabling a backup method if only one method is enabled and there are more available.
```

ocean90 (approved):
Thanks looks good to me! Just one minor typo.

shaunek-hero:
I don't have an immediate need for this, but props for doing this change, it could be useful in the future!

PR #676, opened 2025-04-02
Label recommended methods to simplify the configuration

### LINKED ISSUE #483: Help folks choose which provider is right for them
Many folks aren't familiar w/ 2FA and it's not intuitive to them, so they might struggle to pick a provider that fits their threat model. it could help to add some kind of messaging to the Providers table that tells people how secure & convenient each method is, so they can make an informed decision about tradeoffs.

Rough idea:
Screen Shot 2022-10-19 at 9 31 32 AM

Alternatively, that could be displayed as list, like:
```
Strength of security: 5/5
Ease of setup: 2/5
Ease of use: 4/5
```

... or even as a "help me choose" type of wizard. Or maybe just linking off to an article or video that describes each provider and explains the pros/cons in depth.

kasparsd:
## What?

Marks methods that are relatively secure and easy to setup as "Recommended".

## Why?

Fixes #483.

## How?

- Introduce a "Recommended" label that applied to specific methods.
- Add a `two_factor_recommended_providers` filter that allows adding/removing methods from being "Recommended".
- Update the help text explaining why the "Recommended" methods are recommended.

## Screenshots or screencast

recommended-methods

## Changelog Entry

Added: mark TOTP and backup codes as recommended methods to simplify the decision making.


kasparsd:
@jeffpaul @iandunn What are your thoughts on this implementation as the first pass at helping users to pick the two-factor method?

iandunn:
That looks good to me 👍🏻

Issue #544, opened 2023-03-20
Repo/process updates


jeffpaul:
### Is your enhancement related to a problem? Please describe.

In working through the 0.8.0 release, the following items came up that may be worth considering:
1. Considering updating from the current `deploy.yml` workflow to one utilizing https://github.com/10up/action-wordpress-plugin-deploy <-- already prepared in #543
2. The changelog in `readme.txt` links to https://github.com/wordpress/two-factor/releases.  Should we have a full changeling in `readme.txt`?
3. Is `assets` folder within SVN `trunk` necessary given that duplicates the root `assets` folder?
4. License not fully correct per expected formats. [composer.json](https://github.com/WordPress/two-factor/blob/c0eae28ce5c83d28bdf810dc940f33febe38a0b7/composer.json#L5) properly lists `GPL-2.0-or-later`, [`package.json`](https://github.com/WordPress/two-factor/blob/c0eae28ce5c83d28bdf810dc940f33febe38a0b7/package.json#L32) shows `GPL-2.0+` instead of `GPL-2.0-or-later`, [`two-factor.php`](https://github.com/WordPress/two-factor/blob/c0eae28ce5c83d28bdf810dc940f33febe38a0b7/two-factor.php#L11-L18) leaves out the `License` and `License URI` [header fields](https://developer.wordpress.org/plugins/plugin-basics/header-requirements/#header-fields), and [`readme.txt`](https://github.com/WordPress/two-factor/blob/c0eae28ce5c83d28bdf810dc940f33febe38a0b7/readme.txt#L2-L7) also leaves out the `License` and `License URI` [header fields](https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/#readme-header-information).  I'd recommending ensuring all these note `GPL-2.0-or-later` as the license and where links are required that we link to https://spdx.org/licenses/GPL-2.0-or-later.html
5. Similarly ensure that the other header fields in `two-factor.php` and `readme.txt` match the formats/information as listed by the WP.org Plugin team in [`Header Requirements`](https://developer.wordpress.org/plugins/plugin-basics/header-requirements/#header-fields) and [`Plugin Readmes`](https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/#readme-header-information) respectively.
[...]

PR #683, opened 2025-04-30
Fix/assets

jeffpaul:
## What?
Related to #544:
> Is assets folder within SVN trunk necessary given that duplicates the root assets folder?

This PR renames the `assets` folder to `.wordpress-org` and ignores it for distribution/deploy.

## Why?
This should remove the duplicate `assets` folder from our `trunk` branch on SVN (https://plugins.trac.wordpress.org/browser/two-factor/#trunk/assets) and leave the top level `assets` folder with these files (https://plugins.trac.wordpress.org/browser/two-factor/assets).

## Changelog Entry

> Added - New feature.
> Removed - Feature.
> Fixed - Bug fix.
> Security - Vulnerability.


.distignore
```diff
@@ -1,4 +1,5 @@
 # Directories
+/.wordpress-org/
```
kasparsd, line 2: Looks like action-wordpress-plugin-deploy defaults to `.wordpress-org` being the default name for the `assets` directory so there is no additional config need for the GitHub Action.

.gitignore
```diff
@@ -1,8 +1,14 @@
-.DS_Store
-/node_modules/
-/npm-debug.log
-/vendor/
+# Directories
+/.git
+/.github
+/.wordpress-org
```
kasparsd, line 4: We still want to track it in this GitHub repo, right?

kasparsd (commented):
Had question about the directory being added `.gitignore`. I believe we want to keep tracking it in the repo, just not push it to SVN (covered by `.distignore` update).

.gitignore
jeffpaul: Changed in [8b132b4](https://github.com/WordPress/two-factor/pull/683/commits/8b132b456562109eb78044e936b402f24c555b10).

PR #688, opened 2025-05-07
Trigger checkbox js change event

gedeminas:
## What?
Triggers change event on checkbox state change.

## Why?
Enables listening to events more easily.


kasparsd (commented):
Looks like jQuery is not sending change events for the input elements when the props are updated per https://stackoverflow.com/a/24410621 so this makes sense.

Can you please check if there are other scripts where this would need to be applied?

PR #697, opened 2025-07-03
Fix #300: support for application passwords

### LINKED ISSUE #300: Link to Application Passwords
Users might want to disable the two factor authentication for certain API requests so we should document the `two_factor_user_api_login_enable` filter and how to use it together with [Application Passwords](https://wordpress.org/plugins/application-passwords/) plugin.

joostdekeijzer:
Suggested fix for issue #300

## What?
Proposed fix for #300: when a user is authenticated using an application password no 2factor is needed.

## Why?

Re #300

## How?

Depending of if the `application_password_did_authenticate` action has run the plugin now passes `TRUE` (did run) or `FALSE` (did not run) through the `two_factor_user_api_login_enable` filter in the `is_user_api_login_enabled` function.

## Testing Instructions

1. Install the WordPress app on your device
2. Create an application password and login with your WordPress app

If you succeed my PR works, if you can't login my PR does not work.

## Changelog Entry

> Added - New feature.
> Removed - Feature.
> Fixed - Bug fix.
> Security - Vulnerability.

Fixed #300
Added application passwords support.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -776,7 +776,7 @@ class Two_Factor_Core {
 	 * @return boolean
 	 */
 	public static function is_user_api_login_enabled( $user_id ) {
-		return (bool) apply_filters( 'two_factor_user_api_login_enable', false, $user_id );
+		return (bool) apply_filters( 'two_factor_user_api_login_enable', (bool) did_action( 'application_password_did_authenticate' ), $user_id );
 	}
 
 	/**


kasparsd:
Thanks for the pull request!

The goal of #300 is to make the two-factor plugin more vocal about Application Password and XML-RPC logins being blocked by default. I believe we don't want to enable API logins by default. Primarily because XML-RPC logins allow standard user passwords for auth.

So the solution could be adding a notice to the user profiles near the Application Passwords section. Something like:

> The Two Factor plugin currently prevents logins with application passwords. See this documentation for how to adjust that.

In addition, we could also add a checkbox to the two-factor settings that allows enabling application passwords for that particular user. The option would be hidden if the `two_factor_user_api_login_enable` was already allowing API logins.

class-two-factor-core.php
```diff
@@ -776,7 +776,7 @@ public static function filter_authenticate_block_cookies( $user ) {
 	 * @return boolean
 	 */
 	public static function is_user_api_login_enabled( $user_id ) {
-		return (bool) apply_filters( 'two_factor_user_api_login_enable', false, $user_id );
+		return (bool) apply_filters( 'two_factor_user_api_login_enable', (bool) did_action( 'application_password_did_authenticate' ), $user_id );
```
kasparsd, line 779: This is a good action to use, I believe. It fires right after a _successful_ application passwords login.

kasparsd (approved):
I've been thinking about this some more and I believe this plugin should match the behaviour of most other application (including Gmail and GitHub) where application passwords are allowed to bypass the second factor.

This will also enable XML-RPC logins as long as an application password was used.

Issue #737, opened 2025-12-11
Avoid pushing unreleased `Stable tag` to SVN trunk


kasparsd:
### Is your enhancement related to a problem? Please describe.

During the release process (described in RELEASING.md) we're currently:

1. creating a pull request with the release changes (version bump, readme updates, etc.),
2. merging in the pull request to `master,
3. creating a new release in GitHub which creates a new tag for that latest revision of `master`.

With the introduction of `master` deploys to WP-org `trunk` in #734, it now:

1. pushed the upcoming release to WP-org SVN `trunk` (even though it hasn't been tagged and deployed yet).
2. pushes the SVN `tag/x.x.x` when the release is tagged and the CI runs.

This leads to the following warning on WP.org during step 1 and 2:

<img width="2464" height="484" alt="Image" src="https://github.com/user-attachments/assets/ce421f2c-d625-42b5-8c37-5a2c9f7b6766" />

### Proposed Solution

Introduce a `trunk` branch which is updated only _after_ releases (possibly by the release workflow)?

Use it for isolated updates to readme and WP-org assets. Automatically merge it back to `master` if there are differences.

### Please confirm that you have searched existing issues in this repository.

Yes

PR #734, opened 2025-12-11
Fix blueprint and wporg asset deploys

kasparsd:
## What?

- Ensure the `blueprint.json` is deployed to WP.org `assets` directory.
- Use a dedicated deploy action for asset and readme changes to avoid tagging version releases for minor updates.

## Why?

Fixes #731.

## How?

Note:

- This will deploy the `readme.txt` of the latest release tag version _before_ the tag has been created since that happens after merging the release pull request into `master`. I'm not 100% sure what WP-org does when the specified stable tag is actually not part of SVN. 
 
## Changelog Entry

> Added - New feature.
> Changed - Existing functionality.
> Deprecated - Soon-to-be removed feature.
> Removed - Feature.
> Fixed - Bug fix.
> Security - Vulnerability.

PR #738, opened 2025-12-11
Pause deploys to SVN trunk for merges to `master`

kasparsd:
## What?

Disable SVN trunk deploys (introduced in #734) since they push out versions before they're actually released.

## Why?

Temporary solution to #737.

## Changelog Entry

> Added - New feature.
> Removed - Feature.
> Fixed - Bug fix.
> Security - Vulnerability.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,7 +1,5 @@
 # Release Process
 
-Note that all merges to the `master` branch have the asset directory `.wordpress-org` and `readme.txt` deployed to WordPress.org to avoid creating a new version release when updating only `Tested up to` header or screenshots.
-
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -37,7 +37,7 @@ jobs:
 
       - name: WordPress.org asset and readme update
         uses: 10up/action-wordpress-plugin-asset-update@stable
-        if: github.ref_name == 'master'
+        if: github.ref_name == 'trunk'
```
kasparsd: This branch doesn't exist so it will never run this action.

PR #760, opened 2026-01-30
Fix missing_direct_file_access_protection

### LINKED ISSUE #759: Plugin Check: missing_direct_file_access_protection
### Describe the bug

Currently we get a lot "missing_direct_file_access_protection " errors reported by Plugin Check.
These affects the following files:
- class-two-factor-compat.php
- class-two-factor-core.php
- providers/class-two-factor-backup-codes.php
- providers/class-two-factor-dummy.php
- providers/class-two-factor-email.php
- providers/class-two-factor-fido-u2f.php
- providers/class-two-factor-fido-u2f-admin.php
- providers/class-two-factor-totp.php
- two-factor.php

It would also affect those but we'll skip those files as those come from WP directly, correct @kasparsd ?
- FILE: includes/function.login-footer.php
- FILE: includes/function.login-header.php

Here is the full message for one example file:
## `class-two-factor-compat.php`

| Line | Column | Type | Code | Message | Docs |
| --- | --- | --- | --- | --- | --- |
| 0 | 0 | ERROR | missing_direct_file_access_protection | PHP file should prevent direct access. Add a check like: if ( ! defined( 'ABSPATH' ) ) exit; | [Docs](https://developer.wordpress.org/plugins/wordpress-org/common-issues/#direct-file-access) |

### Steps to Reproduce

1. Install Plugin Check Plugin
2. Choose Two Factor, Categories = "Plugin Repo" and Types = "Error" & "Warning"
3. see results

_No response_

### Environment information

_No response_

Yes

Yes

masteradhoc:
Fixes #759

## What?

Adds direct file access protection to plugin PHP files to prevent them from being executed outside of the WordPress runtime.

## Why?

WordPress.org coding standards require plugin PHP files to block direct access when WordPress is not loaded. [Learn more here](https://developer.wordpress.org/plugins/wordpress-org/common-issues/#direct-file-access)

## How?

The PR adds a standard abspath guard at the top of affected PHP files. This ensures the files exit early when accessed directly, while leaving normal WordPress execution completely unchanged. No functional or behavioral logic was modified.

## Testing Instructions

1. Install Plugin Check Plugin
2. Choose Two Factor, Categories = "Plugin Repo" and Types = "Error" & "Warning"
3. see results
4. apply fix
5. see results

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -5,10 +5,6 @@
  * @package Two_Factor
  */
 
-if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Exit if accessed directly.
-}
-
 /**
  * Class for creating two factor authorization.
  *


kasparsd:
As mentioned above, WP core also don't have them in classes and [even in](https://github.com/WordPress/wordpress-develop/blob/trunk/wp-config-sample.php) `wp-config.php` (which loads WP core) so I would suggest we don't add them. I'm OK with having it in the main plugin bootstrap file, though.

PR #764, opened 2026-01-31
Add Settings Page

### LINKED ISSUE #249: Add plugin settings
Allow site admins to restrict the available two factor options, see #247.

masteradhoc:
## What?

- Adds Plugin Settings to the plugin row action links on the Plugins screen.
- Adjusts the current "Settings" link in the plugin row action links to "User Settings"
- Adds a dedicated Two-Factor Settings admin page under Settings → Two-Factor.
- Introduces a new settings UI (class-two-factor-settings.php) to disable specific Two-Factor providers site-wide via a saved option.

Fixes #249

## Why?

Right now, site admins don’t have a UI to manage which Two-Factor providers are available on the site. Disabling providers requires custom code via filters. This PR provides a simple admin-facing settings page to enable/disable providers without requiring code changes.

## How?

two-factor.php
- Loads the settings UI class.
- Registers an admin menu entry (Settings → Two-Factor) and renders the settings page.

class-two-factor-settings.php
- Renders a settings page listing providers (via Two_Factor_Core::get_providers()).
- Saves disabled provider keys to the two_factor_disabled_providers option (with nonce + sanitization).

class-two-factor-core.php
- Updates plugin action links to include:
- Plugin Settings → options-general.php?page=two-factor-settings
- User Settings → profile.php#application-passwords-section (existing destination preserved)- 

Adds enforcement filters:
-  Filters two_factor_providers to remove disabled providers from the available set.
- Filters user/provider related hooks so disabled providers can’t be enabled/used when disallowed.

## Testing Instructions

1. Install/activate the plugin.
2. Go to Settings → Two-Factor.
3. Check/uncheck a few providers and click Save Settings.
4. Verify that:
     Disabled providers no longer appear as available providers for users.
     Existing users can’t enable newly-disabled providers.
5. Go to Plugins → Two-Factor and confirm the action links show:
     Plugin Settings
     User Settings
6. Confirm the settings page itself still lists all providers so an admin can re-enable them.

This PR also works with existing filter to disable providers. Those will not be shown on the settings page:
```
add_filter( 'two_factor_providers', function( $providers ) {
    // Remove providers you want to hide for testing.
    unset( $providers['Two_Factor_Email'] );
    unset( $providers['Two_Factor_Backup_Codes'] );
    // unset( $providers['Two_Factor_Totp'] );
    // unset( $providers['Two_Factor_FIDO_U2F'] );
    // unset( $providers['Two_Factor_Dummy'] );
    return $providers;
}, 99 );
```

## Screenshots or screencast

1. Settings UI
image

2. Profile user UI after adjusting the settings
image

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -395,7 +395,10 @@ class Two_Factor_Core {
 
 		// Show plugin settings first, then user settings.
 		array_unshift( $links, $user_settings_link );
-		array_unshift( $links, $plugin_settings_link );
+
+		if ( current_user_can( 'manage_options' ) ) {
+			array_unshift( $links, $plugin_settings_link );
+		}
 
 		return $links;
 	}


georgestephanis:
So my uncertainty right now is how necessary this is.

We've got U2F pared down, so the only providers we'd ship with are Backup Codes, TOTP, and Emailed Passwords?

I don't think we want to enable disabling emailed passwords trivially as it's an important fallback.  Backup codes and TOTP I'm also not sure we want to make easy to disable -- I'd personally prefer to have any disabling ui limited to a filter, with any additional providers easily disabled by disabling their plugin?

Maybe I'm missing a use case here, @masteradhoc?

TimothyBJacobs:
IMO something like this would be important for a Core merge proposal. Some site owners are not going to want to allow TOTP because of the risk their users lose their devices. 

I do think the UI should be switched though. Having a Disabled list, that you then enable with checkboxes get's confusing. I'd make it a list of Enabled Providers, and default them all to on.

PR #765, opened 2026-02-01
Update required WordPress Version

masteradhoc:
## What?

[Based on our readme we support the latest two WP releases.](https://github.com/WordPress/two-factor/blob/14d448ffaed4fe87e99ff154f3b9d778af3355e7/readme.txt#L42-L44)
With the release of 6.9 (12.2025) this means we support 6.8 and 6.9.
This PR updates the version number in the plugin header.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -55,13 +55,12 @@ jobs:
           SVN_USERNAME: ${{ secrets.SVN_USERNAME }}
           SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
 
+
       - name: Upload release asset
-        uses: actions/upload-release-asset@v1
+        uses: softprops/action-gh-release@v2
         if: startsWith( github.ref, 'refs/tags/' )
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          upload_url: ${{ github.event.release.upload_url }}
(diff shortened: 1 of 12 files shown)


kasparsd:
Yeah, that's a tricky thing. I personally see the upgrade/dependency policy mostly as a guide of what to expect from the plugin. So if at some point we _need_ to bump the WP/PHP requirement because of some feature or practical requirement, we could do it in a way that is predictable and understandable to everyone.

Also agree that it is very likely that sites using an older WP core would also run outdated plugins. Even the [usage graph here](https://wordpress.org/plugins/two-factor/advanced/) somewhat confirms that only 62% of plugin users run the latest version of the plugin.

usage

masteradhoc:
@kasparsd yeah probably if the WP version is not getting updated also the plugin is left behind. Totally understand your point here. If you want feel free to merge, if you want to wait another plugin release with this one its also not an issue for me at all.

kasparsd:
@jeffpaul What are your thoughts on this? Should we bump the WP/PHP requirements as early as possible or delay them, if possible?

jeffpaul:
My stance on plugins, is to generally keep to a limited number of major versions to support (e.g. current major and 1-2 previous) in hopes that helps nudge folks to upgrade their WP version.  In that case, if we're putting into place an L-1 policy then with 6.9 released I'd be fine bumping the minimum to 6.8.

PR #781, opened 2026-02-07
Since docs

masteradhoc:
## What?
Update docblocks to include @since tags inline with [WordPress documentation standards](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#since-section-changelogs).

## Why?
Accurate @since tags help maintainers and users track when APIs were introduced, and they are required for WordPress coding standards compliance.

## How?
Checked git blame step by step and added since tags based on when the functions initially

## Testing Instructions
- Run phpcs to confirm docblock standards compliance.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -8,7 +8,7 @@
 /**
  * Class for creating two factor authorization.
  *
- * @since 0.1-dev
+ * @since 0.2.0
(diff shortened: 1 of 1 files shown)


kasparsd (approved):
This is great!

Would love to know how you extracted these.

Issue #783, opened 2026-02-09
TOTP URL Around QR Code is not Rendered


sjinks:
### Describe the bug

The TOTP URL inside the `#two-factor-qr-code` paragraph in the User Profile is not rendered:

<img width="521" height="269" alt="Image" src="https://github.com/user-attachments/assets/b0c6971d-efb1-4c81-b3ea-3cefb68d34fe" />

This happens because `esc_url()` returns an empty string if it cannot recognize the protocol.

### Steps to Reproduce

1. Log in.
2. Got to /wp-admin/profile.php
3. Check the link around the QR code under the "Please scan the QR code or manually copy the shared secret key from below to your Authenticator app:" label.

### Screenshots, screen recording, code snippet

See above.

### Environment information

The default installation of WP 6.9.1 with the Two Factor plugin installed.

### Please confirm that you have searched existing issues in this repository.

Yes

### Please confirm that you have tested with all plugins deactivated except Two-Factor.

Yes

PR #784, opened 2026-02-09
fix(totp): otpauth link in QR code URL

sjinks:
## What?

The TOTP URL inside the `#two-factor-qr-code` paragraph in the User Profile is not rendered because `esc_url()` returns an empty string if it encounters an unknown/not allowed protocol.

Fixes: #783

## Why?

`esc_url()` returns an empty string if it encounters an unknown/not allowed protocol. `otpauth:` is not among the allowed ones.

## How?

Pass `array( 'otpauth' )` to `esc_url()`.

## Testing Instructions

1. Log in.
2. Got to /wp-admin/profile.php
3. Check the link around the QR code under the "Please scan the QR code or manually copy the shared secret key from below to your Authenticator app:" label.

## Screenshots or screencast

Screenshot_20260209_192446

## Changelog Entry

> Fixed - OTP Authentication URL is rendered correctly.


kasparsd (approved):
Thank you! Not sure how this got missed. I see that other instances have the protocol added correctly.

PR #798, opened 2026-02-17
fix: Ensure two-factor methods are fully configured before enabling them

### LINKED ISSUE #157: Activating multiple methods is confusing and can fail silently
What I did:

1. Activated Two Factor on Engie on my sandbox only, logged in and navigated to my user profile…
2. Checked “SMS”, “Backup Verification Codes”, and “Time Based One-time Password” methods
3. Clicked “Update Profile”
4. The edit profile page came back with “Two-Factor: You are out of backup codes and need to regenerate!”… no methods were checked
5. It seems like if any method which you have checked fails it’s activation checks, then any new methods you have checked also fails activation.

Suggestions:

* Consider moving Two Factor configuration to it’s own page; the user profile page is already crowded, and breaking this out might help make things clearer. Perhaps leave a link to the “configure two factor” page.
* An explanatory admin notice to say that activating methods has failed
* An explanatory admin notice for each failed method, explaining what needs to be done to get past this

### LINKED ISSUE #796: Silent fallback to recovery codes when TOTP key is missing
## Summary

When `_two_factor_enabled_providers` lists `Two_Factor_Totp` but `_two_factor_totp_key` is missing from user meta, `get_primary_provider_for_user()` silently falls back to the next available provider — typically `Two_Factor_Backup_Codes`. The user sees a prompt for a recovery code when they expect to enter a TOTP code from their authenticator app.

There is no warning or indication that the primary method was skipped.

## Steps to Reproduce

1. Enable TOTP for a user via the profile page.
2. Delete `_two_factor_totp_key` from user meta (simulating a failed TOTP setup — see #157 and the related issue I'm filing about the two-step setup UX).
3. Leave `_two_factor_enabled_providers` intact with `Two_Factor_Totp` listed.
4. Log out and log back in.
5. The 2FA prompt shows "Enter a recovery code" instead of "Enter the code from your authenticator app."

This inconsistent state occurs naturally when a user checks the TOTP "Enabled" checkbox and clicks "Update Profile" without first clicking the "Submit" verification button that saves the TOTP key via REST API.

## Expected Behavior

When `get_primary_provider_for_user()` falls back from the user's configured primary provider to a different provider, the plugin should display a visible warning — either on the login 2FA screen or on the user's profile page. A message like "Your configured TOTP method is unavailable. Falling back to recovery codes." would prevent confusion.

Alternatively (or additionally), the profile save handler could validate that the TOTP key exists before allowing `Two_Factor_Totp` to be saved in `_two_factor_enabled_providers`.

## Impact

This affects any downstream consumer of `get_primary_provider_for_user()`. In our case, the [WP Sudo](https://github.com/dknauss/wp-sudo) reauthentication plugin calls this method to render a 2FA challenge. When Two Factor ~~silently~~ "quietly" falls back to Backup Codes (or any other alternative), the 2FA challenge page gives correct but very muted feedback in the text response. A user who has become blind to this feedback will enter a valid TOTP code (the customary default), but it is validated (correctly) as a backup code, and validation fails with no explanation. In this type of scenario a **louder** bit of UX copy or other feedback is needed.

## Environment

- Two Factor 0.14.2
- WordPress 6.7.x
- Tested on both MySQL and SQLite backends

## Related Issues

- #157 — Covers the general UX confusion around activating multiple methods
- #586 — PR addressing "fail open" when provider classes disappear
- #643 — Auto-enables TOTP checkbox after REST verification, but doesn't prevent the inconsistent state

### LINKED ISSUE #797: TOTP setup: Update Profile saves enabled-providers without validating TOTP key exists
## Summary

The TOTP setup on the profile page uses a two-step save flow:

1. **"Submit" button** (inline, next to the code input) — fires a REST API call (`POST /two-factor/1.0/totp`) that validates the authenticator code and saves `_two_factor_totp_key`.
2. **"Update Profile" button** (bottom of page) — submits the main profile form, which saves `_two_factor_enabled_providers` with whatever checkboxes are checked.

If a user checks the TOTP "Enabled" checkbox, skips the "Submit" verification button, and clicks "Update Profile" directly, `_two_factor_enabled_providers` is saved with `Two_Factor_Totp` listed but `_two_factor_totp_key` was never written. This creates an inconsistent state.

## Steps to Reproduce

1. Go to your profile page with Two Factor active.
2. Check the "Enabled" checkbox next to Time Based One-Time Password (TOTP).
3. **Do not** scan the QR code or click the "Submit" verification button.
4. Scroll down and click "Update Profile."
5. Check user meta: `_two_factor_enabled_providers` now includes `Two_Factor_Totp`, but `_two_factor_totp_key` is empty.

## Expected Behavior

The server-side profile save handler (`personal_options_update` / `edit_user_profile_update`) should validate that the TOTP key exists in user meta before allowing `Two_Factor_Totp` to be included in `_two_factor_enabled_providers`. If the key is missing, the provider should be silently removed from the enabled list and/or an admin notice should warn the user that TOTP setup is incomplete.

The PR #643 improved the JS side (auto-checking the Enabled box after successful REST verification), but the server side still allows saving an invalid provider configuration.

## Impact

The inconsistent state causes `Two_Factor_Totp::is_available_for_user()` to return `false` (no key), which causes `get_primary_provider_for_user()` to silently fall back to another provider (typically Backup Codes). See #796 for the downstream impact of the silent fallback.

This is easy to trigger accidentally — the "Update Profile" button is the large, prominent button at the bottom of the page, while the TOTP "Submit" verification button is a small inline button that users naturally overlook.

## Suggested Fix

In the `personal_options_update` handler where `_two_factor_enabled_providers` is saved, add a check:

```php
if ( in_array( 'Two_Factor_Totp', $providers, true ) ) {
    $key = get_user_meta( $user_id, self::SECRET_META_KEY, true );
    if ( empty( $key ) ) {
        // Remove TOTP from enabled providers — setup was never completed.
        $providers = array_diff( $providers, array( 'Two_Factor_Totp' ) );
        // Optionally: set an admin notice warning the user.
    }
}
```

## Environment

- Two Factor 0.14.2
- WordPress 6.7.x
- Tested on both MySQL and SQLite backends

## Related Issues

- #796 — Silent fallback to recovery codes when TOTP key is missing (the downstream consequence of this bug)
- #157 — General UX confusion around activating multiple methods
- #643 — Auto-enables TOTP checkbox after REST verification (JS-side improvement)
- #507 — Analogous issue for Backup Codes (saved before user intends)

kasparsd:
## What?

Check that a method is fully configured (TOTP secret stored, backup codes stored) before enabling it for the user.

## Why?

Fixes #797, fixes #796, fixes #157.

## How?

- Introduce an error store for the plugin.
- Forward any errors to the profile edit logic.
- Render provider-specific errors inline the fields.

## Testing Instructions

1. Go to Two Factor settings on a fresh user profile.
2. Enable "Authentication Code" method and save the profile.
3. Confirm that an error was shown at the top of the page and next to the provider.
4. Confirm that the method _was not_ stored as enabled.

~Note: the error handling on user profile pages don't allow us to easily render a notice when the method is not actually enabled. This needs to be implemented in a follow-up request.~ (added this here since there was no decent way to show the errors both at the top of the page and in the context of the actual provider config).

## Screenshots or screencast

All errors added to the top of the profile page if validation fails:

profile-errors

And provider-specific errors rendered inline the provider config:

provider-errors

## Changelog Entry

> Added - New feature.
> Removed - Feature.
> Fixed - Bug fix.
> Security - Vulnerability.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -2484,7 +2484,7 @@ class Two_Factor_Core {
 				// No providers, enabling one (or more)
 				( ! $existing_providers && $enabled_providers ) ||
 				// Has providers, and is disabling one (or more), but remaining with 2FA.
-				( $existing_providers && $enabled_providers && array_diff( $existing_providers, $enabled_providers ) )
+				( $existing_providers && $enabled_providers && array_diff( $existing_providers, array_keys( $enabled_providers ) ) )
 			) {
 				if ( $user_id === get_current_user_id() ) {
 					// Keep the current session, destroy others sessions for this user.


class-two-factor-core.php
```diff
@@ -2042,7 +2043,7 @@ private static function get_recommended_providers( $user ) {
 	 */
 	private static function render_user_providers_form( $user, $providers ) {
 		$primary_provider_key      = self::get_primary_provider_key_selected_for_user( $user );
-		$enabled_providers         = self::get_enabled_providers_for_user( $user );
+		$available_providers       = self::get_available_providers_for_user( $user );
```
kasparsd: Use the "enabled && configured" list of providers for representing the "enabled" state.

class-two-factor-core.php
```diff
@@ -2208,13 +2209,23 @@ public static function user_two_factor_options_update( $user_id ) {
...
 			// Enable only the available providers.
-			$enabled_providers = array_intersect( $enabled_providers, array_keys( $providers ) );
-			update_user_meta( $user_id, self::ENABLED_PROVIDERS_USER_META_KEY, $enabled_providers );
+			$enabled_providers = array_intersect_key( $providers, array_flip( $enabled_providers ) );
+
+			// Ensure the enabled providers are configured and can be enabled.
+			foreach ( $enabled_providers as $provider_key => $provider ) {
+				if ( ! $provider->is_available_for_user( $user ) ) {
+					// TODO: Use the `user_profile_update_errors` filter to show these errors too.
+					unset( $enabled_providers[ $provider_key ] );
```
kasparsd: Prevent storing the provider as enabled, if it isn't fully configured.

class-two-factor-core.php
```diff
@@ -106,6 +113,7 @@ public static function add_hooks( $compat ) {
 		add_filter( 'manage_users_columns', array( __CLASS__, 'filter_manage_users_columns' ) );
 		add_filter( 'wpmu_users_columns', array( __CLASS__, 'filter_manage_users_columns' ) );
 		add_filter( 'manage_users_custom_column', array( __CLASS__, 'manage_users_custom_column' ), 10, 3 );
+		add_action( 'user_profile_update_errors', array( __CLASS__, 'action_user_profile_update_errors' ) );
```
kasparsd: Avoid displaying "Updated" message by forwarding our errors to `edit_user()` early in `wp-admin/user-edit.php`.

masteradhoc:
@kasparsd just tested this and works great.

1) went on a new profile
2) tried to add the "Authenticator App" and "Recovery Codes" without generating the backup codes or entering the TOTP
3) this message shows on top of the edit user page
image
4) also shows on the respective part
image
image

Also tested the flow when the Authenticator App gets resetted / disabled. Works also then!

LGTM!

masteradhoc (approved):
Only thing left is to recheck the copilot feedback, else all good from my side!

class-two-factor-core.php
```diff
@@ -2117,11 +2205,29 @@ private static function render_user_providers_form( $user, $providers ) {
...
+	 *
+	 * @param string $provider_key The provider key to get errors for.
+	 *
+	 * @return WP_Error[] List of errors for the provider.
+	 */
+	private static function get_provider_errors( string $provider_key ): array {
+		return array_filter(
+			self::$profile_errors,
+			static function ( WP_Error $error ) use ( $provider_key ) {
+				$error_data = $error->get_error_data(); // This currently supports only one error per instance.
```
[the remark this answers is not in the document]
kasparsd: Updated.

class-two-factor-core.php
```diff
@@ -2032,17 +2100,36 @@ private static function get_recommended_providers( $user ) {
...
+	 *
+	 * @param WP_Error[] $errors List of errors to render.
+	 */
+	private static function render_errors( array $errors ) {
+		foreach ( $errors as $error ) {
+			if ( $error->has_errors() ) {
+				wp_admin_notice(
+					implode( '</p><p>', $error->get_error_messages() ),
+					array(
+						'type'               => $error->get_error_data()['type'] ?? 'error',
```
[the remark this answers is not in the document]
kasparsd: Fixed.

kasparsd:
@masteradhoc This is now updated to address the feedback. Could you please re-test the implementation?

PR #803, opened 2026-02-21
fix: constructors must not return value

sjinks:
## What?
This pull request removes a redundant `return $this;` statement from the constructor. This aligns with PHP class construction practices.

## Why?
In PHP, constructors are [void](https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor) and, therefore, must not return a value.

## How?
Remove `return $this;`.

## Testing Instructions
N/A - there are no functional changes.

## Screenshots or screencast
N/A

### DIFF
--- a/providers/class-two-factor-provider.php
+++ b/providers/class-two-factor-provider.php
@@ -37,7 +37,6 @@ abstract class Two_Factor_Provider {
 	 * @since 0.1-dev
 	 */
 	protected function __construct() {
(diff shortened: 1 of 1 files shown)


georgestephanis (approved):
Looks good.  This was due to legacy behavior and how we were storing the instances in the static previously I believe, but didn't update with the newer version?

Either way, LGTM

PR #804, opened 2026-02-22
New Style for Backup Codes

masteradhoc:
## What?
Improve the Recovery Codes UI to a more wp.org style by:
- displaying generated codes in a two-column highlighted block,
- adding a **Copy Codes** action (comma-separated values),

## Why?
The previous recovery code display was harder to scan and used quite some space.
This change improves readability and usability, and adds a faster way to copy codes for secure storage while preserving current functionality.

## How?
- Updated the generated recovery code list rendering to use a two-column layout with tuned column gap/width.
- Added a dedicated container style for the code block.
- Added a `Copy Codes` control and clipboard logic to copy the latest generated codes as a comma-separated string.

## Testing Instructions
1. Go to a user profile with Two-Factor enabled and open the Recovery Codes section.
2. Click **Generate new recovery codes**.
3. Confirm:
   - the code block appears in two columns,
   - visual spacing/background are applied correctly,
4. Click **Copy Codes** and paste into a text editor.
   - Verify output is comma-separated codes (e.g. `12345678,87654321,...`).
5. Click **Download Codes** and verify the downloaded file still contains all generated codes.
6. Confirm the remaining codes counter updates after generation.

## Screenshots or screencast
Old Style:
image

New Style:
image

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1592,7 +1592,7 @@ class Two_Factor_Core {
 			do_action( 'login_footer' ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound -- Core WordPress action.
 			?>
 			<?php if ( $customize_login ) : ?>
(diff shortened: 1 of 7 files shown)


georgestephanis (approved):
Looks reasonable, but it seems the codes aren't rendering with a monospace font?  May be worth setting, but that's trivial.

Issue #63, opened 2015-09-17
Code Preparation: HTML 5 validation


kraftbj:
Before merge consideration, we need to run the HTML output (settings UI, etc) through validation to ensure we're meeting specs or at least can justify any exceptions.

PR #805, opened 2026-02-22
Remove unnecessary type="text/javascript" attributes from <script> tags (W3C validation warning)

masteradhoc:
Step to achieve #63 

## What?

This PR removes redundant `type="text/javascript"` attributes from inline `<script>` tags across the Two-Factor plugin to resolve W3C validation warnings.

---

## Why?

Under HTML5, `text/javascript` is the default value for `<script>` elements. Explicitly defining it is unnecessary and produces validation warnings such as:

> “The type attribute is unnecessary for JavaScript resources.”

Removing the attribute improves standards compliance and eliminates avoidable validation noise without changing behavior.

---

## How?

All instances of:

```html
<script type="text/javascript">
```

were replaced with:

```html
<script>
```

Updated files:

- `class-two-factor-core.php`
- `providers/class-two-factor-backup-codes.php`
- `providers/class-two-factor-email.php`
- `providers/class-two-factor-totp.php`

No JavaScript logic or execution flow was modified. Only the redundant attribute was removed.

---

## Testing Instructions

1. Apply the patch.
2. run pages through validator
3. make sure we dont have this error
"The type attribute is unnecessary for JavaScript resources warning."

---

## Screenshots or screencast

Not applicable (markup-only standards compliance update).

---

## Changelog Entry

> Changed - Removed unnecessary `type="text/javascript"` attributes from script tags for HTML5 compliance.

### DIFF
--- a/providers/class-two-factor-backup-codes.php
+++ b/providers/class-two-factor-backup-codes.php
@@ -186,7 +186,7 @@ class Two_Factor_Backup_Codes extends Two_Factor_Provider {
 				<a class="button button-two-factor-backup-codes-download button-secondary hide-if-no-js" href="javascript:void(0);" id="two-factor-backup-codes-download-link" download="two-factor-backup-codes.txt"><?php esc_html_e( 'Download Codes', 'two-factor' ); ?></a>
 			<p>
 		</div>
-		<script type="text/javascript">
+		<script>
 			( function( $ ) {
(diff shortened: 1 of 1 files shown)


georgestephanis (approved):
It's optional in html5 by my understanding, and if omitted is the fallback to text/javascript mime type.  Probably safe to drop, but there's also no real harm in having it.  I'm fine with either.

PR #807, opened 2026-02-22
Move class-two-factor-core.php login styles from inline to enqueued stylesheet

masteradhoc:
Step to achieve #63 

## What?
This PR removes the inline CSS block from the Two-Factor login screen and moves them to the user-edit.css stylesheet instead.

## Why?
fix the outstanding TODO

## How?
The login rendering flow now enqueues user-edit.css before outputting the login header, so the 2FA login UI receives the same styles via the stylesheet.

## Testing Instructions

1. Enable the Two-Factor plugin and configure a 2FA provider for a test user.
2. Log out, then log in as that user to reach the 2FA challenge screen.
3. Confirm the styles still apply:

### DIFF
--- a/user-edit.css
+++ b/user-edit.css
@@ -48,3 +48,29 @@
 .input.authcode::placeholder {
 	opacity: 0.5;
 }
+
+.two-factor-backup-codes-wrapper > :not(:last-child) {
+	margin-bottom: 1em;
+}
+
+.two-factor-backup-codes-list-wrap {
+	background-color: #ddd;
+	display: inline-block;
+	margin-top: 24px;
+	padding: 20px;
+}
+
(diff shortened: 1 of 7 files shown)


user-edit.css
```diff
@@ -17,3 +17,34 @@
...
+	margin-bottom: 1em;
+}
+
+.input.authcode {
+	letter-spacing: 0.3em;
+}
+
+.input.authcode::placeholder {
+	opacity: 0.5;
+}
```
masteradhoc:
```suggestion
#loginform .input.authcode {
	letter-spacing: 0.3em;
}

#loginform .input.authcode::placeholder {
	opacity: 0.5;
}
```

PR #808, opened 2026-02-23
Cleanup TOTP Provider: Remove legacy PHP 5.6 compatibility logic & duplicate autocomplete attribute

masteradhoc:
## What?
Refactors pack64() to remove outdated PHP <5.6 compatibility code while retaining 32-bit support.
+ remove duplicated autocomplete attribute

## Why?
PHP 5.6 is no longer supported. The previous implementation contained unnecessary version checks and redundant branching, increasing complexity without benefit.

## How?
- Removed version_compare logic.
- Simplified to a single early-return structure.
- Retained 32-bit fallback using manual high/low 32-bit packing.
- Kept PHP_INT_SIZE >= 8 for defensive architecture handling.

### DIFF
--- a/providers/class-two-factor-totp.php
+++ b/providers/class-two-factor-totp.php
@@ -690,11 +690,11 @@ class Two_Factor_Totp extends Two_Factor_Provider {
 	 *
 	 * @since 0.2.0
 	 *
-	 * @param mixed $value The value to be packed.
+	 * @param int $value The value to be packed.
 	 *
 	 * @return string Binary packed string.
 	 */
-	public static function pack64( mixed $value ): string {
+	public static function pack64( int $value ): string {
 		// Native 64-bit support (modern PHP on 64-bit builds).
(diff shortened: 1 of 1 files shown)


providers/class-two-factor-totp.php
```diff
@@ -655,31 +655,22 @@ public static function generate_key( $bitsize = self::DEFAULT_KEY_BIT_SIZE ) {
...
-		}
-
-		$lowmap = 0xffffffff;
-		$lower  = $value & $lowmap;
-
-		return pack( 'NN', $higher, $lower );
+	public static function pack64(int $value): string
+	{
+	    // Native 64-bit support (modern PHP on 64-bit builds)
+	    if (PHP_INT_SIZE === 8) {
```
georgestephanis: Missing whitespace / core style guides here and following.

georgestephanis (changes requested):
Some whitespace issues that need tidying for core styles.  PHPCS can probably catch them automatically?

Issue #437, opened 2022-03-24
PHP doesn't pass the WordPress coding standards


kasparsd:
The codebase currently reports several critical issues when checked against the WPCS rulesets.

```
FILE: providers/class-two-factor-email.php
----------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
----------------------------------------------------------------------
 299 | WARNING | Processing form data without nonce verification.
     |         | (WordPress.Security.NonceVerification.Recommended)
 316 | WARNING | Processing form data without nonce verification.
     |         | (WordPress.Security.NonceVerification.Recommended)
----------------------------------------------------------------------


FILE: providers/class-two-factor-backup-codes.php
----------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 1 LINE
----------------------------------------------------------------------
 307 | ERROR | Processing form data without nonce verification.
     |       | (WordPress.Security.NonceVerification.Missing)
 307 | ERROR | Processing form data without nonce verification.
     |       | (WordPress.Security.NonceVerification.Missing)
----------------------------------------------------------------------


FILE: providers/class-two-factor-fido-u2f-admin.php
----------------------------------------------------------------------
FOUND 2 ERRORS AND 4 WARNINGS AFFECTING 3 LINES
----------------------------------------------------------------------
 [24](https://github.com/WordPress/two-factor/runs/5678917495?check_suite_focus=true#step:8:24)0 | WARNING | Detected usage of a non-sanitized input variable:
     |         | $_POST['u2f_response']
     |         | (WordPress.Security.ValidatedSanitizedInput.InputNotSanitized)
 240 | ERROR   | Detected usage of a possibly undefined superglobal
     |         | array index: $_POST['u2f_response']. Use isset() or
     |         | empty() to check the index exists before using
     |         | it
     |         | (WordPress.Security.ValidatedSanitizedInput.InputNotValidated)
 [27](https://github.com/WordPress/two-factor/runs/5678917495?check_suite_focus=true#step:8:27)7 | WARNING | Processing form data without nonce verification.
      |         | (WordPress.Security.NonceVerification.Recommended)
  885 | WARNING | Processing form data without nonce
      |         | verification.
      |         | (WordPress.Security.NonceVerification.Recommended)
  885 | WARNING | Detected usage of a non-sanitized input variable:
      |         | $_REQUEST['redirect_to']
[...]

PR #818, opened 2026-02-26
Fix PHPCS and PHPStan issues across multiple files

aslamdoctor:
## Summary
- Add PHPStan bootstrap file for runtime constants (`TWO_FACTOR_DIR`, `TWO_FACTOR_VERSION`) unreachable during static analysis
- Add missing properties (`$new`, `$last_used`) to `Registration` class in `includes/Yubico/U2F.php`
- Fix PHPDoc types for `show_two_factor_login`, `process_provider`, `authentication_page`, `rename_link`, `delete_link`, and `pack64`
- Fix undefined variable bug in `wp_ajax_inline_save` where a non-matching key could be incorrectly updated
- Add input validation, sanitization (`sanitize_text_field`), and `wp_unslash` for `$_POST`/`$_REQUEST` usage in `class-two-factor-fido-u2f-admin.php`
- Remove redundant `isset($user->ID)` checks and always-true conditions
- Cast `base_convert()` result to `(int)` for array offset usage in `base32_encode()`

Reference #437

## Test plan
- [ ] Run `./vendor/bin/phpstan analyse --memory-limit=1G --no-progress` — should pass with 0 errors at level 0
- [ ] Temporarily set PHPStan level to 3 in `phpstan.dist.neon` and run analysis — should pass with 0 errors
- [ ] Temporarily set PHPStan level to 4 in `phpstan.dist.neon` and run analysis — should pass with 0 errors
- [ ] Run `./vendor/bin/phpcs --standard=WordPress providers/class-two-factor-fido-u2f-admin.php` — should pass with 0 errors/warnings
- [ ] Verify TOTP authentication and setup still works
- [ ] Verify email-based two-factor authentication still works

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.2%22%2C%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22git%3Adirectory%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Ftwo-factor.git%22%2C%22ref%22%3A%22437-php-wpcs-fixes%22%2C%22path%22%3A%22%2F%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


masteradhoc (changes requested):

providers/class-two-factor-email.php
```diff
@@ -397,7 +397,7 @@ public function pre_process_authentication( $user ) {
 	 */
 	public function validate_authentication( $user ) {
 		$code = $this->sanitize_code_from_request( 'two-factor-email-code' );
-		if ( ! isset( $user->ID ) || ! $code ) {
+		if ( ! $code ) {
```
georgestephanis: A number of these extra conditionals that you're clearing are there to account for unit tests that pass in `false` to authentication page methods --

https://github.com/WordPress/two-factor/pull/19

Are we sure they're safe to remove in the testing context as well as the real-world use context?

georgestephanis (dismissed):

masteradhoc (approved):
adjusted the PR based on the copilot feedback. LGTM.

class-two-factor-core.php
```diff
@@ -1804,9 +1804,9 @@ public static function _login_form_revalidate_2fa( $nonce = '', $provider = '',
 	 *
 	 * @since 0.9.0
 	 *
-	 * @param object  $provider        The Two Factor Provider.
-	 * @param WP_User $user            The user being authenticated.
-	 * @param bool    $is_post_request Whether the request is a POST request.
+	 * @param object|null $provider        The Two Factor Provider.
+	 * @param WP_User    $user            The user being authenticated.
+	 * @param bool       $is_post_request Whether the request is a POST request.
```
huzaifaalmesbah: The PHPDoc parameter alignment spacing is incorrect.
```suggestion
	 * @param object|null $provider        The Two Factor Provider.
	 * @param WP_User     $user            The user being authenticated.
	 * @param bool        $is_post_request Whether the request is a POST request.
```

two-factor.php
```diff
@@ -22,19 +22,17 @@
...
+if ( ! defined( 'TWO_FACTOR_DIR' ) ) {
+	define( 'TWO_FACTOR_DIR', __DIR__ . '/' );
 }
 
-/**
- * Shortcut constant to the path of this file.
- */
-define( 'TWO_FACTOR_DIR', plugin_dir_path( __FILE__ ) );
+if ( ! defined( 'TWO_FACTOR_VERSION' ) ) {
+	define( 'TWO_FACTOR_VERSION', '0.15.0' );
```
georgestephanis: Outdated.

providers/class-two-factor-email.php
```diff
@@ -401,12 +405,16 @@ public function pre_process_authentication( $user ) {
...
 	 * @since 0.1-dev
 	 *
-	 * @param WP_User $user WP_User object of the logged-in user.
+	 * @param WP_User|false $user WP_User object of the logged-in user.
 	 * @return boolean
 	 */
 	public function validate_authentication( $user ) {
+		if ( ! $user ) {
+			return false;
+		}
```
georgestephanis: Should this and other checks also be an `instanceof \WP_User` check?

georgestephanis (commented):
A couple minor stuff, but honestly I think I'm okay merging this as-is.

two-factor.php
aslamdoctor: Good catch — the guarded define was carrying the pre-bump `0.15.0` value. Updated to `0.16.0` to match the plugin header in ecba0ac.

providers/class-two-factor-email.php
aslamdoctor: Makes sense — switched all three guards in this class (`authentication_page()`, `pre_process_authentication()`, `validate_authentication()`) to `! ( $user instanceof WP_User )` in ecba0ac, matching the existing checks in `Two_Factor_Core` and the TOTP provider. This also covers truthy non-`WP_User` values, not just the `false` case from the unit tests.

masteradhoc:
Thanks for the work here @aslamdoctor and the reviews @georgestephanis @huzaifaalmesbah @dknauss - we'll get this merged now.

PR #819, opened 2026-03-03
Encrypt TOTP secrets at rest using AES-256-GCM

### LINKED ISSUE #455: Add support for encrypting TOTP secrets
Add support for encrypting TOTP secrets (if a constant is defined).

_**More details to be shared.**_

ericmann:
## Summary

Closes #455. Adds encryption-at-rest for TOTP secrets stored in `wp_usermeta` (`_two_factor_totp_key`) using AES-256-GCM via WordPress core's `sodium_compat`.

- **Fully backward-compatible**: without constants defined, behavior is completely unchanged
- **Opt-in via two `wp-config.php` constants**: `TWO_FACTOR_TOTP_ENCRYPTION_KEY` and `TWO_FACTOR_TOTP_ENCRYPTION_KEY_PREVIOUS`
- **Transparent key rotation**: move current key to `_PREVIOUS`, set a new `_KEY`, and secrets are re-encrypted on next read
- **Opportunistic encryption**: existing plaintext secrets are automatically encrypted when read (if a key is configured)
- **AES-NI hardware guard**: `sodium_crypto_aead_aes256gcm_is_available()` ensures graceful fallback to plaintext on systems without hardware support

### New: `providers/class-two-factor-totp-secret.php`

Static utility class handling all encryption/decryption logic:

- `resolve($stored_value, $user_id)` — main read path: decrypts, opportunistically encrypts plaintext, handles key rotation
- `prepare_for_storage($plaintext, $user_id)` — main write path: encrypts if key available
- `encrypt()` / `decrypt()` — AES-256-GCM with `$user_id` as additional authenticated data (prevents usermeta row-swapping attacks)
- `is_encryption_available()` / `is_encrypted()` — detection helpers
- Encrypted format: `1::<hex_nonce>:<hex_ciphertext>`

### New: `tests/providers/class-two-factor-totp-secret.php`

13 test cases covering:

1. Plaintext passthrough without encryption key (backward compat)
2. Opportunistic encryption on read
3. Encrypt/decrypt roundtrip
4. Key rotation with re-encryption
5. Decryption failure returns empty string
6. `is_encrypted()` format detection
7. Encrypted output format validation
8. Unique nonces per encryption
9. Ciphertext bound to user ID (AD mismatch fails)
10. Full set/get roundtrip with encryption
11. TOTP code validation with encrypted secret
12. Availability check with encrypted secret
13. Availability check with decryption failure

### Modified: `providers/class-two-factor-totp.php`

- `get_user_totp_key()` routes through `apply_filters( 'two_factor_totp_secret_resolve', ... )`
- `set_user_totp_key()` routes through `apply_filters( 'two_factor_totp_secret_prepare', ... )`
- Default filter callbacks registered in the constructor

### Filters (in `class-two-factor-totp.php`)

- **`two_factor_totp_secret_resolve`** — filters the TOTP secret after reading from DB (default: `Two_Factor_Totp_Secret::resolve`)
- **`two_factor_totp_secret_prepare`** — filters the TOTP secret before writing to DB (default: `Two_Factor_Totp_Secret::prepare_for_storage`)

### Actions (in `class-two-factor-totp-secret.php`)

- **`two_factor_totp_secret_encrypted`** `($user_id)` — after a secret is encrypted (on read or write)
- **`two_factor_totp_secret_decrypted`** `($user_id, $needs_reencrypt)` — after successful decryption
- **`two_factor_totp_secret_rotated`** `($user_id)` — after re-encryption during key rotation
- **`two_factor_totp_secret_decrypt_failed`** `($user_id)` — when decryption fails (useful for security monitoring)

## Setup

Add to `wp-config.php`:

```php
// Generate with: php -r "echo bin2hex(random_bytes(32));"
define( 'TWO_FACTOR_TOTP_ENCRYPTION_KEY', '<64 hex chars>' );

// Only needed during key rotation — set to the old key:
// define( 'TWO_FACTOR_TOTP_ENCRYPTION_KEY_PREVIOUS', '<64 hex chars>' );
```

## Test Plan

- [ ] Run existing tests with no encryption constants → all pass (backward compat verified)
- [ ] Run new `@group encryption` tests → all 13 pass
- [ ] Manual: define `TWO_FACTOR_TOTP_ENCRYPTION_KEY` in `wp-config.php`, set up TOTP, verify DB value starts with `1::`
- [ ] Manual: log in with TOTP using encrypted secret, verify success
- [ ] Manual: rotate key (move current to `_PREVIOUS`, set new `_KEY`), log in, verify DB value re-encrypted
- [ ] Manual: remove all key constants with encrypted DB values, verify TOTP becomes unavailable for affected users

### DIFF
--- a/providers/class-two-factor-totp-secret.php
+++ b/providers/class-two-factor-totp-secret.php
@@ -219,6 +219,15 @@ class Two_Factor_Totp_Secret {
 				$encrypted = self::encrypt( $stored_value, $user_id );
 				if ( false !== $encrypted ) {
 					update_user_meta( $user_id, '_two_factor_totp_key', $encrypted );
+
+					/**
+					 * Fires after a plaintext TOTP secret is opportunistically encrypted on read.
+					 *
+					 * @since 0.10.0
+					 *
+					 * @param int $user_id The user ID whose secret was encrypted.
+					 */
+					do_action( 'two_factor_totp_secret_encrypted', $user_id );
 				}
 			}
 			return $stored_value;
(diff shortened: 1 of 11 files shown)


providers/class-two-factor-totp-secret.php
```diff
@@ -0,0 +1,308 @@
...
+			// If encryption is available, opportunistically encrypt.
+			if ( self::is_encryption_available() ) {
+				$encrypted = self::encrypt( $stored_value, $user_id );
+				if ( false !== $encrypted ) {
+					update_user_meta( $user_id, '_two_factor_totp_key', $encrypted );
+
+					/**
+					 * Fires after a plaintext TOTP secret is opportunistically encrypted on read.
+					 *
+					 * @since 0.10.0
```
sjinks: Is this version value correct? Shouldn't it be something like 0.16.0?

masteradhoc:
FYI: https://make.wordpress.org/core/2026/08/25/proposal-a-secrets-api-for-wordpress-7-2/

georgestephanis:
Also it looks like if if decryption fails due to a bad/rotated key, it may fail open?  TOTP could change to unenforced instead of blocking the login -- which I want to make sure is an intentional decision.

PR #834, opened 2026-03-13
WordPress 7.0 release preparation

masteradhoc:
## What?
Update the PHP test matrix in the CI workflow to prepare for WordPress 7.0 and drop support for WordPress 6.8.
Update also other necessary files so plugin is tested against the newest WP Version.

## Why?
WordPress 6.8 is no longer supported by two-factor. WordPress 6.9 is the new previous major version, and WordPress 7.0 (covered by latest) is the upcoming release. The test matrix needs to reflect the currently supported versions. Additionally, PHP 7.2 and 7.3 are not compatible with the latest WordPress versions and should be excluded from those test runs to avoid false failures.

## How?
Adjusting necessary test files and plugin headers

## Testing Instructions
1. Verify the Actions test matrix runs against latest, trunk, and 6.9
2. Confirm no test jobs are triggered for PHP 7.2 or 7.3 against latest or trunk
3. Confirm 6.8 no longer appears in any test job names
4. check the plugin headers


.github/workflows/test.yml
```diff
@@ -80,14 +80,17 @@ jobs:
...
                   - trunk
-                  - '6.8'
+                  - '6.9'
                 exclude:
                   - php: '7.3'
                     wp: trunk
                   - php: '7.2'
                     wp: trunk
-                  - php: '8.5'
-                    wp: '6.8'
```
georgestephanis: I assume as we're no longer testing against 6.8, we don't need to exclude it with php 8.5 due to an inherent version conflict

georgestephanis (approved):

PR #848, opened 2026-03-23
Fix email resend during 2FA rate limiting

### LINKED ISSUE #847: Email resend bypasses retry delay in 2FA challenge flow
## Summary

## Why this matters

## Verified code path

## Potential impact

## Suggested fix

Either:

## Related context

dknauss:
## Summary

In `process_provider()` (line 1692), `pre_process_authentication() `is called before the rate-limit check (line 1702). So a rate-limited user clicking "Resend Code" can bypass the rate limit and get a fresh token emailed — defeating the purpose of rate limiting.

This patch adds a guard in `providers/class-two-factor-email.php` so resend requests do not generate a fresh code while the user is already rate-limited. (It lets the request fall through to the existing error path rather than adding a second error.) It also adds a regression test in `tests/class-two-factor-core.php`. 

## Specific Changes

- Block email resend requests while the user is inside the existing two-factor retry delay.
- Let the request fall through to `Two_Factor_Core::process_provider()` so it returns the normal `two_factor_too_fast` error instead of generating a fresh code.
- Add a regression test covering the resend path during rate limiting.

## Testing
- `php -l providers/class-two-factor-email.php`
- `php -l tests/class-two-factor-core.php`
- `vendor/bin/phpcs -q providers/class-two-factor-email.php tests/class-two-factor-core.php` (shows two pre-existing warnings in `tests/class-two-factor-core.php` unrelated to this change)

Closes #847

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1688,18 +1688,13 @@ class Two_Factor_Core {
 			);
 		}
 
-		// Allow the provider to re-send codes, etc.
-		if ( true === $provider->pre_process_authentication( $user ) ) {
-			return false;
-		}
-
-		// If it's not a POST request, there's no processing to perform.
-		if ( ! $is_post_request ) {
-			return false;
-		}
-
-		// Rate limit two factor authentication attempts.
+		// Rate limit two factor authentication attempts, including pre-processing (e.g. resend).
 		if ( true === self::is_user_rate_limited( $user ) ) {
+			// Invalidate any provider token to prevent reuse after rate limiting.
+			if ( method_exists( $provider, 'delete_token' ) ) {
+				$provider->delete_token( $user->ID );
+			}
+
 			$time_delay = self::get_user_time_delay( $user );
 			$last_login = get_user_meta( $user->ID, self::USER_RATE_LIMIT_KEY, true );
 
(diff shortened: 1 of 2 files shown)


dknauss:
Further thoughts on the security concern around the old token remaining valid for its full 15-minute TTL (`class-two-factor-email.php:159`)... An attacker who intercepted/guessed token A could still use it even after the user is rate-limited. The rate limit blocks new code generation but doesn't invalidate the existing code. My commit 80a84c76840eabd160bb9e15b7170930e3683b1d doesn't address this. 

**Recommendation:** Invalidate the existing token when rate-limit is hit, or at minimum on each failed validation attempt.

dknauss:
> This is a great suggestion. I think your proposal to fix this in the plugin core would be a better approach since it would apply equally to all providers.
> 
> Would you be able to create a pull request that implements that?

Yep, I'll do that.

dknauss:
New PR: wordpress/two-factor#859

dknauss:
**Follow-up on the above:** I implemented this recommendation in [PR #917](https://github.com/WordPress/two-factor/pull/917) (originally #859) but then reverted it after testing revealed a worse problem.

Deleting the token when the rate-limit fires causes `login_html()` → `authentication_page()` to immediately regenerate and email a new code on every blocked POST — because `authentication_page()` checks for a token and sends one if none is found. The resend button itself, the verify button, every blocked submission: each one silently triggered a new email. The flooding vector just moved one step later in the call chain.

The replay concern in the original recommendation is also covered by the existing TTL: the default `two_factor_email_token_ttl` and the max rate-limit cap are both 15 minutes, so a captured token expires at roughly the same time the lockout lifts. Token deletion is unnecessary and counterproductive here.

The resend button visibility during lockout is addressed separately in [PR #921](https://github.com/WordPress/two-factor/pull/921).

PR #858, opened 2026-03-28
fix misleading notice #855

### LINKED ISSUE #855: Misleading notice
### Describe the bug

We are using the `two_factor_providers` filter to remove certain providers, including `Two_Factor_Backup_Codes`. However, we are still seeing the following notice, which in this case is misleading. As far as I can tell, there doesn’t appear to be a straightforward way to remove or suppress it.

'To prevent being locked out of your account, consider enabling a backup method like Recovery Codes in case you lose access to your primary authentication method.'

### Steps to Reproduce

Use the `two_factor_providers` filter to remove certain providers. We got the above notice that is misleading and there isn't any easy way to remove it.

Here is an example of the providers we have removed:

```
add_filter( 'two_factor_providers', function( $providers ) {
  unset( $providers['Two_Factor_FIDO_U2F'] );
  unset( $providers['Two_Factor_Backup_Codes'] );
  unset( $providers['Two_Factor_Dummy'] );
  return $providers;
});
```

Image

### Environment information

_No response_

Yes

Yes

masteradhoc:
## What?
Show a generic fallback notice when the `Two_Factor_Backup_Codes` provider has been removed via the `two_factor_providers` filter, instead of always referencing "Recovery Codes" regardless of whether that provider is available.

## Why?
When sites use the `two_factor_providers` filter to remove `Two_Factor_Backup_Codes`, the notice still tells users to enable "Recovery Codes" — a method that doesn't exist in their setup. This is misleading and confusing.

Fixes #855.

## How?
The existing condition already correctly uses the filtered `$providers` array to determine whether there are multiple providers available (`count( $providers ) > 1`). The fix adds a simple `isset( $providers['Two_Factor_Backup_Codes'] )` check on the message string itself:

- If `Two_Factor_Backup_Codes` is present → show the existing specific message referencing Recovery Codes.
- If `Two_Factor_Backup_Codes` is absent → show a generic message encouraging the user to enable any additional method.

No new filters or structural changes are needed.

## Testing Instructions
1. Install the Two-Factor plugin.
2. Add the following to your theme's `functions.php` or a mu-plugin:
```php
add_filter( 'two_factor_providers', function( $providers ) {
    unset( $providers['Two_Factor_Backup_Codes'] );
    return $providers;
});
```
3. Enable a single primary 2FA method (e.g. TOTP) for a user.
4. Visit the user's profile page.
5. **Before this fix:** the notice references "Recovery Codes" even though that provider is unavailable.
6. **After this fix:** the notice shows the generic message: *"To prevent being locked out of your account, consider enabling an additional two-factor method in case you lose access to your primary authentication method."*
7. Remove the filter and repeat — confirm the original specific message referencing Recovery Codes still appears.


dknauss:
@masteradhoc The dynamic check on the warning notice is the right approach, but as I noted [on the issue](https://github.com/WordPress/two-factor/issues/855#issuecomment-4239765001), there's a second hard-coded "Recovery Codes" mention at [`class-two-factor-core.php:2241`](https://github.com/WordPress/two-factor/blob/master/class-two-factor-core.php#L2241) that needs the same treatment.

Happy to push a commit to your branch if you'd like. Here's what the second fix would look like:

```php
<?php
if ( isset( $providers['Two_Factor_Backup_Codes'] ) ) {
	esc_html_e( 'Configure a primary two-factor method along with an additional authentication method to avoid being locked out if you lose access to your primary method. Methods marked as recommended are more secure and easier to use.', 'two-factor' );
} else {
	esc_html_e( 'Configure a primary two-factor method along with an additional authentication method to avoid being locked out if you lose access to your primary method. Methods marked as recommended are more secure and easier to use.', 'two-factor' );
}
?>
```

And a minor suggestion — the message could be simplified like this:

```php
__( 'To prevent being locked out of your account, consider enabling an additional authentication method in case you lose access to your primary authentication method.', 'two-factor' )
```

To make it absolutely clear this is a "secondary" or "additional" *authentication* method, those terms are better than "backup method," which could be misunderstood as referring to a system backup. 

Also, the "methods marked as recommended" text seems to add a vague and not directly relevant suggestion. The user can see clearly enough what methods are marked as recommended.

masteradhoc:
Thanks @dknauss! 

1) I just added a commit to handle the second notice as well. Can you give it a test?

2) I do find the notice clear and well. Inside two-factor the recommended methods are clearly marked in the UI, in my opinion. What do you think?
image

I'd also prefer to avoid changing strings unnecessarily — every modification invalidates existing translations across all locales.

dknauss:
I tested the latest commit in Studio, and the second notice is now handled too. It works as expected and reads fine for me! 

Not a blocker, but the missing test coverage would be nice to include, so I went ahead and did that here: 

- Test branch: [dknauss/two-factor:codex/pr-858-tests](https://github.com/dknauss/two-factor/tree/codex/pr-858-tests)
- Compare view: [dknauss/two-factor/compare/7c47c7f...codex/pr-858-tests](https://github.com/dknauss/two-factor/compare/7c47c7f...codex/pr-858-tests)

This branch adds coverage in `tests/class-two-factor-core.php` for the two notice variants:
- Recovery Codes wording when `Two_Factor_Backup_Codes` is available.
- No mention of Recover Codes when they are filtered out.

Tested locally with Studio, wp-env, Codex, and Claude.

masteradhoc:
Thanks @dknauss i just added your work to this PR. Appreciate you taking a look at this again :)

georgestephanis (approved):

PR #877, opened 2026-04-12
Fail closed when CSPRNG is unavailable during nonce generation

dknauss:
## Summary

This PR replaces the weak fallback in `create_login_nonce()` with a hard failure.

The existing fallback uses `wp_hash()` with inputs including `$user_id`, `wp_rand()`, and `microtime()`. That defensive path dates back to the PHP 5.x / `random_compat` era. On the plugin’s current minimum PHP version (7.2+), `random_bytes()` should use the OS CSPRNG and should not fail under normal conditions.

If secure randomness is unavailable, failing closed is safer than generating a weaker nonce. Both existing call sites already handle a `false` return with `wp_die()`.

## Change

```diff
try {
    $login_nonce['key'] = bin2hex( random_bytes( 32 ) );
} catch ( Exception $ex ) {
-   $login_nonce['key'] = wp_hash( $user_id . wp_rand() . microtime(), 'nonce' );
+   return false;
}


masteradhoc:
@dknauss any reason to close it?

dknauss:
Whoops, that was unintentional — it must have gotten swept up in some automated cleanup I was doing across several repos. Restoring/reopening.

georgestephanis (approved):

masteradhoc (approved):

PR #911, opened 2026-06-16
Add Playground PR preview workflows

fellyph:
## Summary

- Add fork-safe WordPress Playground PR Preview workflows for Two Factor pull requests.
- Build the plugin with the existing `npm run build` process, package `dist/` as a slug-shaped `two-factor.zip`, and publish a Playground preview button after the read-only build workflow succeeds.
- Use a custom Blueprint so the preview logs in and opens the Two Factor settings page.

## Why

Two Factor needs generated preview files because the release build copies the QR code dependency from `node_modules` into `dist/`. The two-workflow setup keeps pull request code in a read-only build workflow, then lets the trusted `workflow_run` publish workflow expose the ZIP and update the PR description.

## Testing

- Parsed both workflow files as YAML.
- Parsed the embedded Blueprint JSON.
- Ran `npm ci`.
- Ran the configured build/ZIP steps and verified `build/two-factor.zip` extracts to a single `two-factor/` directory.
- Ran `npm run lint:js`.
- Ran `npm run lint:css`.
- Ran `composer validate --strict`.
- Ran `composer lint-compat`.
- Ran `composer lint-phpstan`.

PHPUnit was not run locally because `wp-env` could not connect to Docker in this environment.

### DIFF
--- a/.github/workflows/pr-playground-preview-publish.yml
+++ b/.github/workflows/pr-playground-preview-publish.yml
@@ -1,0 +1,42 @@
+name: PR Playground Preview Publish
+
+on:
+  workflow_run:
+    workflows:
+      - PR Playground Preview Build
+    types:
+      - completed
+
+permissions:
+  actions: read
+  contents: write
+  pull-requests: write
+
+jobs:
+  publish:
+    permissions:
+      actions: read
+      contents: write
+      pull-requests: write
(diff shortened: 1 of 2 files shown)


masteradhoc:
@fellyph would be great to have this feature. ping me when the PR is ready for review.

PR #915, opened 2026-06-30
Add network-level provider settings for WordPress Multisite

### LINKED ISSUE #894: Add network-level provider settings for WordPress Multisite

I am using Two-Factor on a WordPress Multisite network.

The plugin works on Multisite, but the current settings model can be confusing in a network environment.

From what I understand:

- User 2FA configuration is stored at the user level, which makes sense because users are shared across the network.
- However, the enabled provider settings appear to be stored per subsite.
- This means one subsite could allow TOTP and backup codes, another subsite could disable TOTP, and another subsite could allow only email codes.
- As a result, the same user may have a network-wide 2FA configuration, but the available providers can vary depending on the subsite context.

Since 2FA protects the shared user account rather than a specific subsite, this can feel inconsistent for Multisite administrators.

### Proposed Solution

It would be helpful to add network-level provider settings for Multisite installations.

Possible improvements:

1. Add a Network Admin settings page for Two-Factor provider settings.
2. Store enabled provider settings as network options when the plugin is network activated.
3. Allow Super Admins to choose whether subsites can override the network-level provider settings.
4. Optionally add a simple network-level policy for requiring 2FA for Super Admins, administrators, or selected roles.

This would keep the plugin lightweight while making its Multisite behavior more consistent and predictable.

### Designs

No mockup needed.

A possible UI location could be:

Network Admin → Settings → Two-Factor

Possible settings:

- Enabled providers for the network:
  - TOTP
  - Email codes
  - Backup codes
  - Other providers

- Allow subsites to override provider settings:
  - Yes / No

- Optional enforcement:
  - Do not enforce
  - Require 2FA for Super Admins
  - Require 2FA for Super Admins and site administrators
  - Require 2FA for selected roles

I considered managing the provider settings separately on each subsite, but this can become inconsistent and difficult to maintain on a Multisite network.

I also considered using custom code or an MU plugin to force providers network-wide, but that feels fragile and may break if plugin internals change in a future update.

A dedicated Network Admin setting would be cleaner, easier to understand, and more consistent with how Multisite user management works.

Yes

faisalahammad:
## Summary

This PR adds a Network Admin settings page for Two-Factor provider configuration. When the plugin is network-activated, a Super Admin can choose which providers are available across the network and decide whether subsites may override that list. The change keeps the existing site-level settings screen intact for non-network mode or when no network policy is configured.

Fixes #894

### `class-two-factor-core.php`
- Added network option constants: `ENABLED_PROVIDERS_NETWORK_OPTION_KEY` and `NETWORK_ALLOW_SITE_OVERRIDE_OPTION_KEY`.
- Extended `uninstall()` to delete the new network options with `delete_site_option()`.

### `two-factor.php`
- Added `two_factor_is_network_mode()` with a `two_factor_network_mode` filter so the network logic can be unit-tested without a Multisite install.
- Network settings page is registered under `network_admin_menu` only when the plugin is network-activated.
- Added `two_factor_get_site_enabled_providers_option()` and rewrote `two_factor_get_enabled_providers_option()` to return the effective provider list: network option first, with subsite intersection when override is allowed.
- Updated the provider and user-provider filters to use the effective list.

### `settings/class-two-factor-network-settings.php` (new)
- Network Admin settings page that saves provider list and the override toggle with nonces and capability checks.
- Provider keys are validated against the registered providers before being saved.

### `settings/class-two-factor-settings.php`
- When the network has taken over provider settings:
  - Override disabled: the page is read-only, shows a network-managed notice, and hides the Save button.
  - Override enabled: shows a narrowing notice and disables checkboxes for providers not enabled at the network level.
- When the network has not configured a policy, the page behaves as before.

### `tests/class-two-factor-network-settings.php` (new)
- Covers network mode detection, effective option logic, provider filtering, uninstall cleanup, and the site settings page read-only/override notices.

## Testing

**Test 1: Network-managed provider list**
1. Network-activate the plugin.
2. Go to Network Admin → Settings → Two-Factor.
3. Disable "Email Codes" and save.
4. On a subsite, edit a user profile.
5. Result: Email Codes is not available; TOTP and Backup Codes remain available.

**Test 2: Subsite override disabled**
1. On the network settings page, leave "Allow subsites to override" unchecked.
2. On a subsite, go to Settings → Two-Factor.
3. Result: a notice appears, all checkboxes are disabled, and the Save button is hidden.

**Test 3: Subsite override enabled**
1. On the network settings page, enable only Email and TOTP, and check "Allow subsites to override".
2. On a subsite, go to Settings → Two-Factor.
3. Result: Backup Codes is disabled/unchecked; the site can narrow the list to TOTP only.

**Automated tests**
```bash
npm run composer -- test
npm run lint:php
npm run lint:phpstan
```
All 203 tests pass, PHPStan reports no errors, and no new PHPCS errors are introduced in the changed files.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.2%22%2C%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22git%3Adirectory%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Ftwo-factor.git%22%2C%22ref%22%3A%22fix%2F894-network-provider-settings%22%2C%22path%22%3A%22%2F%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -38,6 +38,24 @@ class Two_Factor_Core {
 	const ENABLED_PROVIDERS_OPTION_KEY = 'two_factor_enabled_providers';
 
 	/**
+	 * The network-wide enabled providers option key.
+	 *
+	 * @since 0.17.0
+	 *
+	 * @type string
+	 */
+	const ENABLED_PROVIDERS_NETWORK_OPTION_KEY = 'two_factor_network_enabled_providers';
+
+	/**
+	 * The network option key that controls whether subsites can override provider settings.
+	 *
+	 * @since 0.17.0
+	 *
+	 * @type string
+	 */
+	const NETWORK_ALLOW_SITE_OVERRIDE_OPTION_KEY = 'two_factor_network_allow_site_override';
+
+	/**
 	 * The user meta nonce key.
 	 *
 	 * @type string
@@ -203,6 +221,12 @@ class Two_Factor_Core {
 			self::ENABLED_PROVIDERS_OPTION_KEY,
 		);
 
+		// Network options are stored in sitemeta on Multisite and in wp_options on single-site.
+		$network_option_keys = array(
+			self::ENABLED_PROVIDERS_NETWORK_OPTION_KEY,
+			self::NETWORK_ALLOW_SITE_OVERRIDE_OPTION_KEY,
+		);
+
 		$providers = self::get_default_providers();
 
 		/** This filter is documented in the get_providers() method */
@@ -241,6 +265,12 @@ class Two_Factor_Core {
 		if ( ! empty( $option_keys ) ) {
 			foreach ( $option_keys as $option_key ) {
 				delete_option( $option_key );
+			}
+		}
+
+		if ( ! empty( $network_option_keys ) ) {
+			foreach ( $network_option_keys as $option_key ) {
+				delete_site_option( $option_key );
 			}
 		}
 
@@ -1186,7 +1216,7 @@ class Two_Factor_Core {
 
 			foreach ( $backup_providers as $backup_provider_key => $backup_provider ) {
 				$backup_link_args['provider'] = $backup_provider_key;
-				$links[] = array(
+				$links[]                      = array(
 					'url'   => self::login_url( $backup_link_args ),
 					'label' => $backup_provider->get_alternative_provider_label(),
 				);
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -167,6 +167,7 @@ class Two_Factor_Core {
 
 		// Add Settings link to plugin action links.
 		add_filter( 'plugin_action_links_' . plugin_basename( TWO_FACTOR_DIR . 'two-factor.php' ), array( __CLASS__, 'add_settings_action_link' ) );
+		add_filter( 'network_admin_plugin_action_links', array( __CLASS__, 'add_network_settings_action_link' ), 10, 4 );
 
 		$compat->init();
 	}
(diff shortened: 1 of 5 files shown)


acer2021lg:
Thanks for working on this PR. This matches the Multisite use case I described in #894.

From my perspective as a network admin, the most important behavior is that providers disabled in Network Admin should not be re-enabled by individual subsites. If “Allow subsites to override” is disabled, the network provider list should be the effective policy for all subsites. Even if subsite override is enabled, subsites should only be able to narrow the network provider list, not expand it.

I noticed two small points while looking at the implementation:

1. In a network-activated setup, the plugin “Settings” action link appears to still point to the regular site-level settings page. Should it point to the Network Admin Two-Factor settings page instead?

2. If a Super Admin disables all providers at the network level, the setting can be saved as an empty provider list. I’m not sure whether that is intentional, but since the impact is network-wide, maybe it would be worth adding a warning or some protection against accidental misconfiguration.

Thanks again for taking this on.

faisalahammad:
Hi @acer2021lg,

Thank you for the feedback. Both points are addressed:

1. **Settings action link**: the link is now network-aware. In network-activated mode, it points to  for users with . I also registered the  handler so the link appears on the Network Admin plugins screen. Subsite admins still see the site-level settings page, which shows the read-only network-managed notice.

2. **Empty provider list**: the network settings page now rejects saving an empty provider list. It shows an error notice and preserves the existing configuration, so a Super Admin cannot accidentally disable all providers across the network.

Thanks again.

PR #917, opened 2026-07-03
Move rate-limit check before provider pre-processing, preserve email token

### LINKED ISSUE #847: Email resend bypasses retry delay in 2FA challenge flow

dknauss:
> **Note:** This is a rebase of [PR #859](https://github.com/WordPress/two-factor/pull/859), which was accidentally closed without comment on June 14. The branch has been rebased onto current master. Apologies for the confusion — the milestone, reviewer assignments, and discussion thread are in the original.

## Problem

Two-factor has a rate-limit that makes you wait longer after each wrong code you enter, if you keep failing retries, so it becomes a short-term ban or lockout. The lockout period starts at 2 seconds, then 4, 8, 16... up to 15 minutes if you keep failing repeated login attempts. 

**The bug:** clicking "Resend Code" remains available and operational for rate-limited users during the lockout period. Clicking "Resend Code" while locked out is possible because the resend logic runs before the rate-limit check. 

**Security and UX Impact:**
- An attacker can deliberately lock out and flood a known user with useless emailed codes.
- A valid user who locks themselves out can do the same thing to themself: if they keep clicking the "Resend Code" and trying to log in again, even with a valid code and the right credentials they will be blocked and the lockout period will be extended. 

## Solution

PR [#917](https://github.com/WordPress/two-factor/pull/917) simply moves the rate-limit check to run first, so Resend is blocked during a lockout just like a code submission is. 

- Runs the rate-limit check before any provider can resend or validate on POST requests. No provider-specific changes.
- Reorders `process_provider()` so the rate-limit gate runs before `pre_process_authentication()`, blocking all providers from resending or rotating state while rate-limited. Previously, the email provider could bypass the rate limit via the "Resend Code" button because `pre_process_authentication()` ran first.
- GET requests (page loads) are excluded from rate-limiting, so the login form renders normally when the rate limit expires.

Implements the core-level approach [suggested by @kasparsd in #848](https://github.com/WordPress/two-factor/pull/848#issuecomment-2755338046).

## Security Context

**Summary:** Moving the rate-limit gate before `pre_process_authentication()` closes the resend bypass. GET requests are excluded to avoid triggering token regeneration on page reload. Pre-existing concurrency races ([#510](https://github.com/WordPress/two-factor/issues/510)) are not worsened.

### Token handling during rate limiting

The email token is intentionally preserved when the rate limit fires. An earlier version of this PR deleted the token on rate-limit, reasoning that a captured code should not outlive the brute-force threshold. In practice this caused `authentication_page()` — called by `login_html()` immediately after the `WP_Error` return — to see no token and regenerate one, sending a new email on every blocked POST. That recreated the flooding vector the rate-limit was meant to stop.

The replay concern is already covered by the token TTL: the default `two_factor_email_token_ttl` and the max rate-limit cap are both 15 minutes, so a captured token expires at the same time the lockout lifts.

### Provider switching during rate limiting

Rate-limit state is per-user, not per-provider (`_two_factor_last_login_failure`, `_two_factor_failed_login_attempts`). If a user triggers the rate limit on TOTP then switches to the email provider, the rate-limit gate still fires. This is intentional — the account is under scrutiny regardless of which provider the failures came from.

### GET requests excluded

The `$is_post_request` early return is kept above the rate-limit check. Without this, a page reload while rate-limited would trigger `authentication_page()` on a GET, causing repeated token emails on every reload. The resend button is a POST form submission, so it is still gated.

### Concurrent request races

The read-modify-write race on the failure counter ([acknowledged by @dd32 in #510](https://github.com/WordPress/two-factor/issues/510)) and the TOCTOU window in `validate_token()` are pre-existing and not worsened by this change.

## Specific changes

- **Reorder `process_provider()`**: `$is_post_request` check first, then `is_user_rate_limited()`, then `pre_process_authentication()` — gating all provider POST actions equally.
- **Tests** (6 total):
  - Email resend blocked while rate-limited.
  - Email token preserved on rate-limited validation attempt (token deletion causes `authentication_page()` to re-send immediately — regression test added in email provider suite).
  - Email token preserved on provider switch while rate-limited.
  - GET request while rate-limited passes through without side effects.
  - Revalidation flow respects rate limiting equally (`_login_form_validate_2fa` and `_login_form_revalidate_2fa` share the same rate-limit state, per [#510](https://github.com/WordPress/two-factor/issues/510)).
  - `authentication_page()` does not send a new email when re-rendered after a rate-limited POST.

## Test plan

- [ ] Verify rate-limited users cannot trigger email resend via POST
- [ ] Verify rate-limited users see `two_factor_too_fast` error on both resend and validation POST requests
- [ ] Verify email token is preserved when rate limit fires (no re-send on blocked POST)
- [ ] Verify page reload (GET) while rate-limited renders the form normally
- [ ] Verify provider switch while rate-limited still blocks the attempt
- [ ] Verify legitimate users can retry with the same token before hitting the rate limit
- [ ] Run `vendor/bin/phpunit tests/class-two-factor-core.php tests/providers/class-two-factor-email.php`

Closes #847
Supersedes #848
Supersedes #859

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.2%22%2C%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22git%3Adirectory%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Ftwo-factor.git%22%2C%22ref%22%3A%22rate-limit-gate%22%2C%22path%22%3A%22%2F%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

---

The companion PR [#921](https://github.com/WordPress/two-factor/pull/921) has a [WordPress Playground demo](https://playground.wordpress.net/?blueprint-url=https://gist.githubusercontent.com/dknauss/2d9cc3e09b16cbbd2ff0072a91e45226/raw/927410265ea7f658a7741b9e02be17f4f1212d33/two-factor-921-rate-limit-messaging.json) that renders the locked-out 2FA login this gate produces — the Resend button is suppressed and only the rate-limit message shows.

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1848,11 +1848,6 @@ class Two_Factor_Core {
 
 		// Rate limit two factor authentication attempts, including pre-processing (e.g. resend).
 		if ( true === self::is_user_rate_limited( $user ) ) {
-			// Invalidate any provider token to prevent reuse after rate limiting.
-			if ( method_exists( $provider, 'delete_token' ) ) {
-				$provider->delete_token( $user->ID );
-			}
-
 			$time_delay = self::get_user_time_delay( $user );
 			$last_login = get_user_meta( $user->ID, self::USER_RATE_LIMIT_KEY, true );
 
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1841,18 +1841,18 @@ class Two_Factor_Core {
 			);
 		}
 
-		// Allow the provider to re-send codes, etc.
-		if ( true === $provider->pre_process_authentication( $user ) ) {
-			return false;
-		}
-
 		// If it's not a POST request, there's no processing to perform.
 		if ( ! $is_post_request ) {
 			return false;
 		}
 
-		// Rate limit two factor authentication attempts.
+		// Rate limit two factor authentication attempts, including pre-processing (e.g. resend).
 		if ( true === self::is_user_rate_limited( $user ) ) {
+			// Invalidate any provider token to prevent reuse after rate limiting.
+			if ( method_exists( $provider, 'delete_token' ) ) {
+				$provider->delete_token( $user->ID );
+			}
+
 			$time_delay = self::get_user_time_delay( $user );
 			$last_login = get_user_meta( $user->ID, self::USER_RATE_LIMIT_KEY, true );
 
@@ -1864,6 +1864,11 @@ class Two_Factor_Core {
 					human_time_diff( $last_login + $time_delay )
 				)
 			);
+		}
+
+		// Allow the provider to re-send codes, etc.
+		if ( true === $provider->pre_process_authentication( $user ) ) {
+			return false;
 		}
 
 		// Ask the provider to verify the second factor.
--- a/tests/class-two-factor-core.php
+++ b/tests/class-two-factor-core.php
@@ -782,7 +782,6 @@ class Test_ClassTwoFactorCore extends WP_UnitTestCase {
 		$provider = Two_Factor_Email::get_instance();
 
 		$provider->generate_token( $user->ID );
-		$original_token = $provider->get_user_token( $user->ID );
 
 		update_user_meta( $user->ID, Two_Factor_Core::USER_FAILED_LOGIN_ATTEMPTS_KEY, 1 );
 		update_user_meta( $user->ID, Two_Factor_Core::USER_RATE_LIMIT_KEY, time() );
--- a/tests/class-two-factor-core.php
+++ b/tests/class-two-factor-core.php
@@ -818,11 +818,11 @@ class Test_ClassTwoFactorCore extends WP_UnitTestCase {
 
 		$this->assertWPError( $result );
 		$this->assertSame( 'two_factor_too_fast', $result->get_error_code() );
-		$this->assertFalse( $provider->user_has_token( $user->ID ), 'Token is invalidated when rate limited' );
+		$this->assertTrue( $provider->user_has_token( $user->ID ), 'Token is preserved when rate limited to avoid triggering re-send on form re-render' );
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


dknauss:
@masteradhoc — could you add this to milestone 0.17.0 and assign reviewers as you did on the original #859? 

I got caught by this trap today — it's really more than a security issue, it's a highly confusing and frustrating gotcha for any user who locks themselves out with a temporary ban after multiple failed login attempts. It ties into some UX issues I've also added, mainly #920 but also #918 and #919.

dknauss:
**Note for reviewers familiar with the original [PR #859](https://github.com/WordPress/two-factor/pull/859):** the approach here has changed significantly since that PR was open.

The original version also deleted the email token when the rate limit fired, on the reasoning that a captured code should not remain valid after brute-force detection. That was removed after testing revealed it recreated the problem we were solving: `login_html()` calls `authentication_page()` immediately after the `WP_Error` return to re-render the form, and `authentication_page()` regenerates and emails a new token whenever none is found — so every blocked POST silently sent a fresh code, including every "Resend Code" click. The flooding path just moved one step later in the call chain.

The fix is now limited to the ordering change in `process_provider()`: rate-limit check before `pre_process_authentication()`. Token state is left alone. A regression test in the email provider suite covers the re-render path to guard against reintroducing the deletion.

PR #922, opened 2026-07-03
fix: reword mixed-audience login failure notice to be informational

### LINKED ISSUE #919: Failed-attempt warning addresses two mutually exclusive audiences simultaneously
## Problem

The notice shown after any failed 2FA attempt ([`class-two-factor-core.php` lines 1027–1028](https://github.com/WordPress/two-factor/blob/master/class-two-factor-core.php#L1027)):

> WARNING: Your account has attempted to login %d times without providing a valid two factor token. The last failed login occurred %s ago. **If this wasn't you, you should reset your password.**

The message speaks to two mutually exclusive people at the same time:

- **The legitimate account holder** — addressed by "your account" and "you should reset your password"
- **A hypothetical attacker reading the form** — implied by "if this wasn't you"

In practice, the person reading this message is almost always the legitimate user, who just mistyped their verification code. Telling them "if this wasn't you, reset your password" in response to their own failed attempt is alarming and misleading. It implies they may have been compromised when the most likely explanation is a typo.

If the goal is to alert the user to unauthorized access attempts, the message should be specific about what that means and when it applies — not appended as a hedge to every failed attempt.

## Impact

- Legitimate users who mistype their code are told to reset their password unnecessarily
- The dual-audience framing makes the message unclear to both audiences
- Affects all providers, not just email

## Suggested direction

Separate the two concerns:

1. Show the failure count and timestamp as informational context (no alarming language for the common case of a typo)
2. Reserve the "reset your password" prompt for cases where the failure count or pattern actually suggests unauthorized access — or remove it from this message entirely and rely on the existing compromised-password reset flow

## Related

- [`Two_Factor_Core::maybe_show_last_login_failure_notice()`](https://github.com/WordPress/two-factor/blob/master/class-two-factor-core.php#L1017)
- [`Two_Factor_Core::should_reset_password()`](https://github.com/WordPress/two-factor/blob/master/class-two-factor-core.php) — the existing flow that handles the compromised-password case separately

dknauss:
Companion to [PR #917](https://github.com/WordPress/two-factor/pull/917) and [PR #921](https://github.com/WordPress/two-factor/pull/921).

Closes #919.

## What changed

`maybe_show_last_login_failure_notice()` in `class-two-factor-core.php`.

**Before:**
> WARNING: Your account has attempted to login %1$s times without providing a valid two factor token. The last failed login occurred %2$s ago. If this wasn't you, you should reset your password.

**After:**
> %1$s failed verification code attempts on this account. The last attempt was %2$s ago. If you did not make these attempts, someone else may know your password. Change your password after you log in.

### Why each change

**Remove "WARNING:"** — same rationale as [PR #921](https://github.com/WordPress/two-factor/pull/921): a severity prefix for a temporary informational notice is alarming and inappropriate.

**Reframe "if this wasn't you," and keep concrete advice** — This notice appears on the 2FA form, which is only reached after a correct password submission, and the failed-attempt counter increments only past that point (`process_provider()` runs after `verify_login_nonce()`). So if the account holder did not make these attempts, someone else submitted their correct password — and a password change is the correct remedy, because it invalidates the password the other party already has. The new copy drops the alarming `WARNING:` framing and the awkward "if this wasn't you," but states the implication plainly and gives the concrete action, deferred to after login because the password can't be changed mid-flow.

**"attempted to login … without providing a valid two factor token" → "failed verification code attempt on this account"** — shorter, plain-language phrasing that doesn't require the user to parse "two factor token" to understand what happened.

## Tests

- `test_maybe_show_last_login_failure_notice` — updated string assertions to match the new text, pinning both the singular and plural forms.
- `test_login_failure_notice_language_is_calm_and_informational` — asserts the absence of `WARNING` and `wasn't you`; asserts the concrete guidance (`someone else may know your password` / `Change your password after you log in`) and the rendered time (`The last attempt was … ago`) are present.

## Not addressed here

This notice is only shown when there is no rate-limit error already displayed (see the `elseif` condition in `login_html()`). The rate-limit error message itself was addressed in [PR #921](https://github.com/WordPress/two-factor/pull/921).

## Try it in Playground

[Try this change in WordPress Playground](https://playground.wordpress.net/?blueprint-url=https://gist.githubusercontent.com/dknauss/d4c2ea836ee261a4103e9b5f16b2ff5b/raw/abd72109f731b8aee96242d15b0bcd1757730d9c/two-factor-922-notice-mixed-audience.json) — boots the 2FA login showing the reworded failure notice.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.2%22%2C%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22git%3Adirectory%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Ftwo-factor.git%22%2C%22ref%22%3A%22ux%2Fnotice-mixed-audience%22%2C%22path%22%3A%22%2F%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1022,10 +1022,10 @@ class Two_Factor_Core {
 			echo '<div id="login_notice" class="message"><strong>';
 			printf(
 				esc_html(
-					/* translators: 1: number of failed login attempts, 2: time since last failed attempt */
+					/* translators: 1: number of failed verification code attempts, 2: human-readable time since the last attempt, e.g. "5 minutes" */
 					_n(
-						'WARNING: Your account has attempted to login %1$s time without providing a valid two factor token. The last failed login occurred %2$s ago. If this wasn\'t you, you should reset your password.',
-						'WARNING: Your account has attempted to login %1$s times without providing a valid two factor token. The last failed login occurred %2$s ago. If this wasn\'t you, you should reset your password.',
+						'%1$s failed verification code attempt on this account. The last attempt was %2$s ago. If you did not make this attempt, review your account security after logging in.',
+						'%1$s failed verification code attempts on this account. The last attempt was %2$s ago. If you did not make these attempts, review your account security after logging in.',
 						$failed_login_count,
 						'two-factor'
 					)
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -1024,8 +1024,8 @@ class Two_Factor_Core {
 				esc_html(
 					/* translators: 1: number of failed verification code attempts, 2: human-readable time since the last attempt, e.g. "5 minutes" */
 					_n(
-						'%1$s failed verification code attempt on this account. The last attempt was %2$s ago. If you did not make this attempt, review your account security after logging in.',
-						'%1$s failed verification code attempts on this account. The last attempt was %2$s ago. If you did not make these attempts, review your account security after logging in.',
+						'%1$s failed verification code attempt on this account. The last attempt was %2$s ago. If you did not make this attempt, someone else may know your password. Change your password after you log in.',
+						'%1$s failed verification code attempts on this account. The last attempt was %2$s ago. If you did not make these attempts, someone else may know your password. Change your password after you log in.',
 						$failed_login_count,
 						'two-factor'
 					)
--- a/tests/class-two-factor-core.php
+++ b/tests/class-two-factor-core.php
@@ -796,8 +796,10 @@ class Test_ClassTwoFactorCore extends WP_UnitTestCase {
 		$contents = ob_get_clean();
 
 		$this->assertNotEmpty( $contents );
-		$this->assertStringNotContainsString( '1 times', $contents );
-		$this->assertStringContainsString( 'failed verification code attempt', $contents );
+		// A single failure uses the singular form; assert the plural is absent, since
+		// "attempt" alone also matches "attempts".
+		$this->assertStringContainsString( '1 failed verification code attempt on this account', $contents );
+		$this->assertStringNotContainsString( 'failed verification code attempts', $contents );
 
 		// 5 failed login attempts 5 hours ago - User should be informed.
 		$five_hours_ago = time() - 5 * HOUR_IN_SECONDS;
@@ -808,7 +810,7 @@ class Test_ClassTwoFactorCore extends WP_UnitTestCase {
 		$contents = ob_get_clean();
 
 		$this->assertNotEmpty( $contents );
-		$this->assertStringContainsString( 'failed verification code attempts', $contents );
+		$this->assertStringContainsString( '5 failed verification code attempts on this account', $contents );
 		$this->assertStringContainsString( human_time_diff( $five_hours_ago ), $contents );
 	}
 
@@ -820,7 +822,7 @@ class Test_ClassTwoFactorCore extends WP_UnitTestCase {
 	public function test_login_failure_notice_language_is_calm_and_informational() {
 		$user = $this->get_dummy_user();
 		update_user_meta( $user->ID, Two_Factor_Core::USER_FAILED_LOGIN_ATTEMPTS_KEY, 3 );
-		update_user_meta( $user->ID, Two_Factor_Core::USER_RATE_LIMIT_KEY, time() - 60 );
+		update_user_meta( $user->ID, Two_Factor_Core::USER_RATE_LIMIT_KEY, time() - MINUTE_IN_SECONDS );
 
 		ob_start();
 		Two_Factor_Core::maybe_show_last_login_failure_notice( $user );
--- a/tests/class-two-factor-core.php
+++ b/tests/class-two-factor-core.php
@@ -832,9 +832,10 @@ class Test_ClassTwoFactorCore extends WP_UnitTestCase {
 	 * @covers Two_Factor_Core::maybe_show_last_login_failure_notice()
 	 */
 	public function test_login_failure_notice_language_is_calm_and_informational() {
-		$user = $this->get_dummy_user();
+		$user           = $this->get_dummy_user();
+		$one_minute_ago = time() - MINUTE_IN_SECONDS;
 		update_user_meta( $user->ID, Two_Factor_Core::USER_FAILED_LOGIN_ATTEMPTS_KEY, 3 );
-		update_user_meta( $user->ID, Two_Factor_Core::USER_RATE_LIMIT_KEY, time() - MINUTE_IN_SECONDS );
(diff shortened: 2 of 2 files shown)


playground/blueprint.json
masteradhoc, line 1: Can you please remove this file?

package-lock.json
masteradhoc, line 1: Can we remove the changes on this file and rather move it to a separate PR?

masteradhoc (changes requested):
Thanks for the PR @dknauss - LGTM. Can you check my two pending comments?

package-lock.json
dknauss: Removed — the `package-lock.json` change came from a Dependabot bump on the fork's master that had not yet landed in upstream. Rebasing the branch onto current upstream/master dropped it cleanly.

playground/blueprint.json
dknauss: Correction to my note above: I've removed the Playground demo from this PR entirely rather than inlining it — a demo pointing at my fork branch is workflow-specific noise that doesn't belong in the PR body. The PR is now just the notice-wording change and its tests.

dknauss: Yes, done -- sorry about that. We've been playgrounding all the things, and Claude gets overzealous about it. Tried to fit in too many tasks that morning.

tests/class-two-factor-core.php
```diff
@@ -797,8 +797,7 @@ public function test_maybe_show_last_login_failure_notice() {
 
 		$this->assertNotEmpty( $contents );
 		$this->assertStringNotContainsString( '1 times', $contents );
-		$this->assertStringContainsString( 'attempted to login', $contents );
-		$this->assertStringContainsString( 'without providing a valid two factor token', $contents );
+		$this->assertStringContainsString( 'failed verification code attempt', $contents );
```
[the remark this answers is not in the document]
dknauss: Right — `1 times` can't appear in the reworded notice, so that assertion was dead. Worth noting `failed verification code attempt` was also weak, since it matches the plural `attempts` too. Now asserting the rendered singular (`1 failed verification code attempt on this account`) and that the plural is absent, with the plural case pinned the same way.

tests/class-two-factor-core.php
```diff
@@ -809,10 +808,31 @@ public function test_maybe_show_last_login_failure_notice() {
...
 
+	/**
+	 * Test that the login failure notice uses calm, informational language.
+	 *
+	 * @covers Two_Factor_Core::maybe_show_last_login_failure_notice()
+	 */
+	public function test_login_failure_notice_language_is_calm_and_informational() {
+		$user = $this->get_dummy_user();
+		update_user_meta( $user->ID, Two_Factor_Core::USER_FAILED_LOGIN_ATTEMPTS_KEY, 3 );
+		update_user_meta( $user->ID, Two_Factor_Core::USER_RATE_LIMIT_KEY, time() - 60 );
```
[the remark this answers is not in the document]
dknauss: Switched to `MINUTE_IN_SECONDS`.

tests/class-two-factor-core.php
```diff
@@ -821,10 +822,31 @@ public function test_maybe_show_last_login_failure_notice() {
...
+		$this->assertStringContainsString( '5 failed verification code attempts on this account', $contents );
 		$this->assertStringContainsString( human_time_diff( $five_hours_ago ), $contents );
 	}
 
+	/**
+	 * Test that the login failure notice uses calm, informational language.
+	 *
+	 * @covers Two_Factor_Core::maybe_show_last_login_failure_notice()
+	 */
+	public function test_login_failure_notice_language_is_calm_and_informational() {
```
masteradhoc, line 831: @dknauss 
None of these actually check that %2$s rendered as something like "1 minute ago." Is that intended?

class-two-factor-core.php
```diff
@@ -1022,10 +1022,10 @@ public static function maybe_show_last_login_failure_notice( $user ) {
 			echo '<div id="login_notice" class="message"><strong>';
 			printf(
 				esc_html(
-					/* translators: 1: number of failed login attempts, 2: time since last failed attempt */
+					/* translators: 1: number of failed verification code attempts, 2: human-readable time since the last attempt, e.g. "5 minutes" */
 					_n(
-						'WARNING: Your account has attempted to login %1$s time without providing a valid two factor token. The last failed login occurred %2$s ago. If this wasn\'t you, you should reset your password.',
-						'WARNING: Your account has attempted to login %1$s times without providing a valid two factor token. The last failed login occurred %2$s ago. If this wasn\'t you, you should reset your password.',
+						'%1$s failed verification code attempt on this account. The last attempt was %2$s ago. If you did not make this attempt, review your account security after logging in.',
+						'%1$s failed verification code attempts on this account. The last attempt was %2$s ago. If you did not make these attempts, review your account security after logging in.',
```
masteradhoc: @dknauss didnt test it on my side _yet_ - but im not sure how clear "review your account security after logging in" is. Can we be more clear on what we actually would suggest the user?

dknauss: @masteradhoc Your question sent me back to trace when this notice actually fires, and it turned up a mistake in my own reasoning. (It's quite a confusing trail to keep track of!)

The failed-attempt counter only increments after `verify_login_nonce()` passes in `process_provider()` — meaning this screen is only reachable after a correct password submission. So if the account holder didn't make those attempts, someone else submitted their correct password. "Review your account security" wasn't just vague, it was not appropriately flagging the urgency of this situation.

The right remedy is a password change, since that invalidates the password the other party (an assumed attacker) already has. 

New wording: "If you did not make these attempts, someone else may know your password. Change your password after you log in." 

(Deferred to after login because they can't change it mid-flow.) I also corrected the PR description, which had argued this backwards — it claimed a reset "doesn't help when the password is already known," which is exactly wrong. Pushed in ad1ea7f.

tests/class-two-factor-core.php
dknauss: Right — the test seeded the timestamp a minute back but never checked the rendered output. Added an assertion that the string contains `"The last attempt was " . human_time_diff( $one_minute_ago ) . " ago"`, so the `%2$s` interpolation is actually exercised. Pushed in ad1ea7f.

masteradhoc (approved):
LGTM! Let's wait for the review of @georgestephanis

georgestephanis (approved):

PR #933, opened 2026-07-22
Fix: Users list table fatals (wp_die) when a user's 2FA provider is deregistered

### LINKED ISSUE #932: Bug: Users list table dies (wp_die) when a user's only 2FA provider has been deregistered
### Summary
If an admin removes a 2FA provider from the registered provider list (e.g. via a `two_factor_providers` filter) while a user still has that provider as their *only* enabled method, the wp-admin **Users** list table fatals mid-render via `wp_die()`. This truncates the page and breaks other admin UI (e.g. Screen Options stops responding).

See https://wordpress.org/support/topic/truncated-users-list-when-previous-2fa-method-unavailable/

### Root cause
`get_available_providers_for_user()` returns a `WP_Error` when a user's enabled provider no longer exists and Email isn't available as a fallback:

```php
} else {
    return new WP_Error( 'no_available_2fa_methods', ... );
}
```

`get_primary_provider_for_user()` then calls `wp_die()` on that error:

```php
} elseif ( is_wp_error( $available_providers ) ) {
    wp_die( $available_providers );
}
```

This is invoked per-row from `manage_users_custom_column()` while rendering the Users list table, so one user's stale data kills the whole admin page for anyone viewing it.

### Expected behavior
A single user's invalid/legacy 2FA state should never be able to `wp_die()` the Users list table. The row should show a graceful status (e.g. "legacy/unavailable method") instead.

### Suggested fix
- Remove the `wp_die()` call from `get_primary_provider_for_user()`; return `null`/`WP_Error` instead and let callers handle it.
- In `manage_users_custom_column()`, check for `null`/`WP_Error` before calling `->get_label()` and render a safe fallback string.
- Audit other callers of `get_primary_provider_for_user()` for the same unguarded assumption.

### Steps to Reproduce

1. Have a user enable Email as their only 2FA method.
2. Remove `Two_Factor_Email` from the registered providers (filter out `two_factor_providers`).
3. Visit wp-admin → Users.
4. Page output truncates when the table reaches that user's row.

_No response_

### Environment information

_No response_

Yes

Yes

masteradhoc:
## What?
Fixes #932

Prevents the wp-admin Users list table from fataling (`wp_die()`) when a user's only enabled 2FA provider has been deregistered.

## Why?
`get_primary_provider_for_user()` calls `wp_die()` when `get_available_providers_for_user()` returns a `WP_Error` (which happens when a user's saved provider no longer exists and Email isn't available as a fallback). This is called per-row from `manage_users_custom_column()`, so one user's stale data breaks the entire Users list page for any admin viewing it, including unrelated UI like Screen Options.

## How?
- Removed the `wp_die()` call in `get_primary_provider_for_user()`; it now returns `null` on error instead of halting execution.
- Updated `manage_users_custom_column()` to check for `null`/`WP_Error` before calling `->get_label()`, rendering a safe "legacy/unavailable method" string instead.
- Audited other callers of `get_primary_provider_for_user()` to confirm none assume a valid object is always returned.

## Use of AI Tools
AI assistance: Yes
Tool(s): Claude
Model(s): Claude Sonnet 5
Used for: Root cause analysis and initial fix suggestions; final implementation and tests reviewed and edited by me.

## Testing Instructions
1. Enable Email 2FA for a test user.
2. Remove `Two_Factor_Email` from registered providers via a `two_factor_providers` filter.
3. Visit wp-admin → Users.
4. Confirm the list renders fully (no truncation) and the affected user's row shows a fallback status instead of a fatal error.

## Screenshots or screencast
| Before | After |
| ------ | ----- |
|        |       |

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.2%22%2C%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22git%3Adirectory%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Ftwo-factor.git%22%2C%22ref%22%3A%22user-list-wp-die%22%2C%22path%22%3A%22%2F%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -2116,10 +2116,6 @@ class Two_Factor_Core {
 			return $output;
 		}
 
-		if ( ! self::is_user_using_two_factor( $user_id ) ) {
-			return sprintf( '<span class="dashicons-before dashicons-no-alt">%s</span>', esc_html__( 'Disabled', 'two-factor' ) );
-		}
-
 		$provider = self::get_primary_provider_for_user( $user_id );
 
 		if ( is_wp_error( $provider ) ) {
--- a/class-two-factor-core.php
+++ b/class-two-factor-core.php
@@ -744,7 +744,9 @@ class Two_Factor_Core {
 	 *
 	 * @param int|WP_User        $user Optional. User ID, or WP_User object of the the user. Defaults to current user.
 	 * @param null|string|object $preferred_provider Optional. The name of the provider, the provider, or empty.
-	 * @return null|object The provider
+	 * @return null|object|WP_Error The provider, null if none is available, or a WP_Error if the user has
+	 *                               provider(s) enabled that are no longer registered (see
+	 *                               Two_Factor_Core::get_primary_provider_for_user()).
 	 */
 	public static function get_provider_for_user( $user = null, $preferred_provider = null ) {
 		$user = self::fetch_user( $user );
@@ -789,6 +791,10 @@ class Two_Factor_Core {
 		$primary_provider    = get_user_meta( $user->ID, self::PROVIDER_USER_META_KEY, true );
 		$available_providers = self::get_available_providers_for_user( $user );
 
+		if ( is_wp_error( $available_providers ) ) {
+			return null;
+		}
+
 		if ( ! empty( $primary_provider ) && ! empty( $available_providers[ $primary_provider ] ) ) {
 			return $primary_provider;
 		}
@@ -802,7 +808,10 @@ class Two_Factor_Core {
 	 * @since 0.2.0
 	 *
 	 * @param int|WP_User $user Optional. User ID, or WP_User object of the the user. Defaults to current user.
-	 * @return object|null
+	 * @return object|null|WP_Error Provider instance, null if the user has none configured, or a WP_Error if the
+	 *                               user has provider(s) enabled that are no longer registered. Callers that render
+	 *                               shared admin UI (e.g. list tables) must not `wp_die()` on the WP_Error case, since
+	 *                               that would break the page for everyone, not just the affected user.
 	 */
 	public static function get_primary_provider_for_user( $user = null ) {
 		$user = self::fetch_user( $user );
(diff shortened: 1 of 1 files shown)


chubes4 (changes requested):
Fuzzed this against a disposable path-multisite network (seed `9338501`: 5,000 randomized provider states, 500 transition sequences, 1,500 cross-site steps). The stale Users-row behavior generally remains fail-closed and no provider/meta leakage or cross-blog bleed appeared, but one PR-path fatal remains:

- `class-two-factor-core.php:1144-1152` calls `array_diff_key()` before checking whether the second availability lookup returned `WP_Error`. If a provider is deregistered between resolution calls, `login_html()` throws `TypeError: array_diff_key(): Argument #1 ($array) must be of type array, WP_Error given`. Move the `WP_Error` guard ahead of `array_diff_key()` and retain fail-closed login behavior.

Minimized replay: make the selected provider available for the initial provider lookups, remove it via `two_factor_providers` before the subsequent availability lookup, then invoke `login_html()`.

Current CI/lint blockers also need resolution: PHPStan flags the new `is_wp_error()` branch against the current return annotation, and PHPCS rejects the new `wp_die( WP_Error )` calls as unescaped output. The original Users-list fix otherwise survived the state campaign, including subsequent-row rendering and malicious stale provider values. Happy to rerun seed `9338501` after the guard/type fixes.

masteradhoc:
Thanks @chubes4 - ready for another review when you are ;)

chubes4 (changes requested):
Second pass against `4fde20d` covered 50,001 randomized provider states across three seeds, 5,000 transition sequences/15,000 cross-site steps, PHP 7.4 + WP 6.9, PHP 8.3 + trunk, PHP 8.5/current, and browser rendering with 64 mixed users. The prior `login_html()` `array_diff_key(WP_Error)` TypeError is fixed and CI is green, but provider-filter churn exposed a remaining fail-open authentication schedule:

- With successive `two_factor_providers` snapshots `[present, absent, present]`, the initial supported lookup sees the persisted enabled provider, the availability lookup returns an empty array, and the later enabled lookup sees it again. `is_user_using_two_factor()` then returns false (`class-two-factor-core.php:702-735`, `:822-875`), allowing `filter_authenticate()` at `:936-953` to leave cookies/REST/XML-RPC password auth unrestricted. Reproduced on PHP 7.4/WP 6.9 and PHP 8.3/trunk. Tracked in #943.

The authentication decision needs one consistent provider snapshot or another durable fail-closed signal for the entire request. Add a regression covering this exact call schedule; fixing only UI/error callers is insufficient.

Adjacent pre-existing malformed-meta crashes also remain: enabled-provider scalar/object values (#941) and primary-provider object/array offsets (#944). They may stay separately scoped, but tests for #933 should include stale-row and fail-closed authentication behavior because the PR changes a public return contract and currently adds none.

No cross-blog bleed, malicious-label leakage, unrelated-row truncation, or extra page errors appeared. Browser assertions passed across main/alpha/network user lists and stale profiles. Worst bounded render probe (1,000 rows × 512 providers) remained approximately O(rows × providers), ~18-19s/50MB. Happy to rerun seeds `9338501-9338503` after the auth snapshot fix.

georgestephanis (approved):
Good to ship, could use a regression test in a pinch?

PR #958, opened 2026-08-12
Fix empty backup code when appending codes for a user with none

### LINKED ISSUE #956: generate_codes() with 'method' => 'append' stores an empty code when the user has none
Generating backup codes with `'method' => 'append'` for a user who has none stores an extra, empty code. The empty entry can never be validated and is never removed, so `codes_remaining_for_user()` permanently reports one more code than the user actually has.

Line references are against `master` at 6245be6. Reproduced on WordPress 7.0.3 with Two Factor 0.16.0 installed from wordpress.org.

## Reproduction

```php
$user  = get_user_by( 'login', 'someone' ); // no backup codes yet
$codes = Two_Factor_Backup_Codes::get_instance()->generate_codes( $user, array( 'method' => 'append' ) );

count( $codes );                                              // 10 — correct
count( get_user_meta( $user->ID, '_two_factor_backup_codes', true ) ); // 11
Two_Factor_Backup_Codes::codes_remaining_for_user( $user );   // 11
```

Index `0` of the stored array is an empty string; the ten real hashes follow it.

## Root cause

`providers/class-two-factor-backup-codes.php:319-321`

```php
// Append or replace (default).
if ( isset( $args['method'] ) && 'append' === $args['method'] ) {
    $codes_hashed = (array) get_user_meta( $user->ID, self::BACKUP_CODES_META_KEY, true );
}
```

`get_user_meta( ..., true )` returns `''` when the key is absent, and `(array) ''` evaluates to `array( '' )` rather than `array()`. The ten generated hashes are then appended to an array that already contains one empty string.

Worth noting the rest of the class already guards this correctly — `codes_remaining_for_user()` at `:400-403` and `validate_code()` at `:491-493` both check `is_array()` before use. Line 320 is the only place that relies on a bare cast.

## Why it matters

The empty entry is permanent. `validate_code()` at `:490-501` iterates the stored hashes and only calls `delete_code()` on a match; `wp_check_password( $code, '' )` never matches, so nothing ever removes it.

That leads to two user-facing consequences once the real codes are spent:

**A user with zero usable codes is still offered backup codes at login.** `is_available_for_user()` at `:214-220` returns `true` whenever `codes_remaining_for_user()` is above zero. With only the empty entry left, the count is `1`, so `Two_Factor_Backup_Codes` stays in `get_available_providers_for_user()` and is presented as a recovery option on the two-factor prompt. Someone locked out of their primary factor selects it, and no code they hold can ever work.

**The low-codes warning states the wrong number.** The notice at `:157-170` reports `$count`, so a user with one real code left is told they have two, and a user with none is told they have one — advising them they still have a way in when they do not.

## Suggested fix

```php
if ( isset( $args['method'] ) && 'append' === $args['method'] ) {
    $existing     = get_user_meta( $user->ID, self::BACKUP_CODES_META_KEY, true );
    $codes_hashed = is_array( $existing ) ? $existing : array();
}
```

This matches the `is_array()` guard the class already uses elsewhere.

A migration for already-affected users may be worth considering separately — filtering empty values out of `codes_remaining_for_user()` would correct the count for existing installs, though it would not remove the stored entry.

## Scope

No caller inside the plugin passes `method => append` — `rest_generate_codes()` at `:345-356` builds `$args` from the request and the REST route does not expose it. So this only affects code calling `generate_codes()` directly, which is how I hit it: an add-on generating a first set of codes for a user during an enrollment flow, passing `append` defensively so that an existing set could never be clobbered. On a user with no codes, that defensive flag is what produces the extra entry.

faithcoder:
## What?
Generating backup codes with `'method' => 'append'` for a user who has no existing codes no longer stores an extra, empty code that can never be validated or removed.

Fixes #956

## Why?
`generate_codes()` cast `get_user_meta( ..., true )` directly to an array. For a user with no stored codes that meta value is `''`, and `(array) ''` evaluates to `array( '' )` rather than an empty array, so an empty string was permanently appended ahead of the real hashes. `codes_remaining_for_user()` then reports one more code than the user actually holds, causing two problems:

- A user with zero usable codes is still offered recovery codes at login.
- The low-codes warning reports one code too many.

## How?
Replaced the bare `(array)` cast with an `is_array()` guard before assigning the existing hashes, matching the pattern already used by `codes_remaining_for_user()` and `validate_code()`. Added a regression test for the append-with-no-existing-codes case.

## Use of AI Tools
AI assistance: Yes
Tool(s): OpenCode
Model(s): deepseek-v4-pro
Used for: Root-cause analysis, fix implementation, and test authoring; the final diff and validation were reviewed by the contributor.

## Testing Instructions
1. `npm run composer -- test -- --filter Tests_Two_Factor_Backup_Codes`
2. `npm test` (full suite)
3. Manual reproduction from the issue:

```php
$user  = get_user_by( 'login', 'someone' ); // no backup codes yet
$codes = Two_Factor_Backup_Codes::get_instance()->generate_codes( $user, array( 'method' => 'append' ) );

count( $codes );                                              // 10
count( get_user_meta( $user->ID, '_two_factor_backup_codes', true ) ); // 10 (was 11)
Two_Factor_Backup_Codes::codes_remaining_for_user( $user );   // 10 (was 11)
```

## Screenshots or screencast
Not applicable (no UI changes).

### DIFF
--- a/providers/class-two-factor-backup-codes.php
+++ b/providers/class-two-factor-backup-codes.php
@@ -317,7 +317,8 @@ class Two_Factor_Backup_Codes extends Two_Factor_Provider {
 
 		// Append or replace (default).
 		if ( isset( $args['method'] ) && 'append' === $args['method'] ) {
-			$codes_hashed = (array) get_user_meta( $user->ID, self::BACKUP_CODES_META_KEY, true );
+			$existing     = get_user_meta( $user->ID, self::BACKUP_CODES_META_KEY, true );
+			$codes_hashed = is_array( $existing ) ? $existing : array();
 		}
 
 		$code_length = $this->get_backup_code_length( $user );
--- a/tests/providers/class-two-factor-backup-codes.php
+++ b/tests/providers/class-two-factor-backup-codes.php
@@ -202,6 +202,33 @@ class Tests_Two_Factor_Backup_Codes extends WP_UnitTestCase {
 	}
 
 	/**
+	 * Verify appending codes for a user with no existing codes does not store an empty entry.
+	 *
+	 * @covers Two_Factor_Backup_Codes::generate_codes
+	 * @covers Two_Factor_Backup_Codes::codes_remaining_for_user
+	 */
+	public function test_generate_codes_append_with_no_existing_codes() {
+		$user = new WP_User( self::factory()->user->create() );
+
+		$codes = $this->provider->generate_codes(
+			$user,
+			array(
+				'number' => Two_Factor_Backup_Codes::NUMBER_OF_CODES,
+				'method' => 'append',
+			)
+		);
+
+		$this->assertCount( Two_Factor_Backup_Codes::NUMBER_OF_CODES, $codes );
+
+		$backup_codes = get_user_meta( $user->ID, Two_Factor_Backup_Codes::BACKUP_CODES_META_KEY, true );
+
+		$this->assertIsArray( $backup_codes );
+		$this->assertCount( Two_Factor_Backup_Codes::NUMBER_OF_CODES, $backup_codes );
+		$this->assertNotContains( '', $backup_codes );
+		$this->assertEquals( Two_Factor_Backup_Codes::NUMBER_OF_CODES, $this->provider->codes_remaining_for_user( $user ) );
+	}
+
+	/**
 	 * Test backup code length filter.
 	 */
 	public function test_backup_code_length_filter() {


masteradhoc (approved):
LGTM! Thanks to @dknauss for the detailed Issue and the PR @faithcoder.

providers/class-two-factor-backup-codes.php
```diff
@@ -317,7 +317,8 @@ public function generate_codes( $user, $args = array() ) {
 
 		// Append or replace (default).
 		if ( isset( $args['method'] ) && 'append' === $args['method'] ) {
-			$codes_hashed = (array) get_user_meta( $user->ID, self::BACKUP_CODES_META_KEY, true );
+			$existing     = get_user_meta( $user->ID, self::BACKUP_CODES_META_KEY, true );
+			$codes_hashed = is_array( $existing ) ? $existing : array();
 		}
```
masteradhoc, line 321: The patch stops new empty entries but leaves every already-affected user with a phantom code that still counts toward codes_remaining_for_user() and still offers a dead Backup Codes option at login. We'll have to decide if thats OK or we want to fix this for everyone

dknauss: Good point. It's important to cover already-affected users; otherwise, the phantom empty-string entry continues to inflate `codes_remaining_for_user()` and exposes an unusable Backup Codes option at login. We can address this without a site-wide migration by excluding empty entries when counting and filtering the existing hashes before appending, which cleans the stored metadata on its next write. Regression coverage could `seed _two_factor_backup_codes` with array( '' ) and verify a zero remaining count, an unavailable provider, and no empty entry after append.

PR #972, opened 2026-09-07
Update PHPStan to 2.x and exclude includes/ from analysis

masteradhoc:
## Why

CI was emitting an upgrade warning on every `lint:phpstan` run:

> You're running an old version of PHPStan. The last release in the 1.12.x series with new features and bugfixes was released on July 17th 2025 [...] Upgrade today to PHPStan 2.2 or newer.

The run also failed with a single error in `includes/function.login-footer.php` (`if.alwaysFalse`).

## What

- **Bump to PHPStan 2.x** — `szepeviktor/phpstan-wordpress` goes from `^1.3` to `^2.0`, which pulls in `phpstan/phpstan` 2.2.13 (was 1.12.34) and `php-stubs/wordpress-stubs` 7.1.0 (was 6.9.4). PHPStan itself stays a transitive dependency of the extension.
- **Drop `includes/` from the analysed paths** — those files are taken over from WordPress core and are already checked upstream. This also resolves the `if.alwaysFalse` failure without touching code that is intentionally kept close to core.
- **Set `treatPhpDocTypesAsCertain: false`** — PHPStan 2 infers types more aggressively and started reporting `alreadyNarrowedType` on defensive runtime checks. Runtime values in WordPress are not guaranteed to match the documented PHPDoc types, so the defensive checks stay and the check is relaxed instead.
- **Remove a redundant guard in the settings screen** — `class_exists( 'Two_Factor_Core' ) && method_exists( 'Two_Factor_Core', 'get_providers' )` is statically always true (the class is loaded unconditionally by `two-factor.php`). This was the one error not covered by `treatPhpDocTypesAsCertain`. The `is_array()` fallback on the return value is kept.

No baseline entries, `@phpstan-ignore` comments, or type casts were added.

## Testing

- `composer lint-phpstan` — no errors
- `composer lint` (PHPCS) — 22/22 clean

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.2%22%2C%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22git%3Adirectory%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Ftwo-factor.git%22%2C%22ref%22%3A%22phpstan-update%22%2C%22path%22%3A%22%2F%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
     "phpcompatibility/phpcompatibility-wp": "3.0.0-alpha2",
     "phpunit/phpunit": "^8.5|^9.6",
     "spatie/phpunit-watcher": "^1.23",
-    "szepeviktor/phpstan-wordpress": "^1.3",
+    "szepeviktor/phpstan-wordpress": "^2.0",
     "wp-coding-standards/wpcs": "^3.3",
     "yoast/phpunit-polyfills": "^4.0"
   },
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "9857e125bdb14775c411bd7b933570ab",
+    "content-hash": "da01ead1426a7dbac3a656eab2d7d74d",
     "packages": [],
     "packages-dev": [
         {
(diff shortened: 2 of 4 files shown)


masteradhoc:
after merging the PR #974 of @georgestephanis this is green now. ready for a review!

PR #980, opened 2026-09-14
Only clear the login nonce once it has expired

georgestephanis:
Fixes the destructive half of `verify_login_nonce()`: an unrecognized nonce no longer discards the pending one.

## The behavior

`verify_login_nonce()` currently deletes the stored login nonce on *any* verification failure, including a simple mismatch:

```php
if ( $hashes_match && time() < $login_nonce['expiration'] ) {
    return true;
}

// Require a fresh nonce if verification fails.
self::delete_login_nonce( $user_id );
```

Its one caller, `validate_login_form_2fa()`, runs that check *before* `process_provider()` — which is where rate limiting lives. So the destructive path is the only one in the flow with nothing throttling it, and it needs no authentication to reach. Anyone can end another user's in-progress 2FA login by submitting a garbage nonce with a guessed user ID. The victim isn't locked out, but their half-finished login dies and they start over at the password prompt.

## Why the delete was there

Not as a considered defense. The original implementation collapsed both failure cases into one branch:

```php
if ( $nonce != $login_nonce['key'] || time() > $login_nonce['expiration'] ) {
    self::delete_login_nonce( $user_id );
    return false;
}
```

Deleting an **expired** nonce is reasonable garbage collection. Deleting on **mismatch** came along for the ride. The 2021 hashing refactor (f3232d9) split the condition apart and added the comment "Require a fresh nonce if verification fails" — describing the existing behavior rather than arguing for it.

## Why dropping it costs nothing

**It isn't brute-forceable.** `create_login_nonce()` uses `bin2hex( random_bytes( 32 ) )` — 256 bits, inside a ten minute window.

**The nonce alone isn't sufficient.** Even a correct one only reaches the second factor prompt; `process_provider()` still demands the real TOTP code, email code, or backup code. A guessed nonce puts an attacker exactly where knowing the password would.

**Rotation already covers it.** On a failed second factor, `validate_login_form_2fa()` calls `create_login_nonce()` again and overwrites the stored value. The "one attempt per prompt" property delete-on-mismatch would supposedly provide is already provided by rotation.

**There's no replay surface to protect.** The stored hash binds `user_id` and `expiration`, so a nonce can't be reused for another user or have its window extended.

## The change

```php
if ( $hashes_match ) {
    if ( time() < $login_nonce['expiration'] ) {
        return true;
    }

    // Expired nonces can never succeed again; clear them.
    self::delete_login_nonce( $user_id );

    return false;
}

// A value we never issued. Leave the pending nonce alone.
return false;
```

`verify_login_nonce()` has exactly one caller and it treats every non-`true` return identically, so nothing downstream shifts.

## Tests

`test_invalid_nonce_deletes_valid_nonce` locks in the behavior being removed, so it inverts:

- Renamed to `test_invalid_nonce_preserves_valid_nonce`; the second assertion flips to `assertTrue`, plus a usermeta check that the pending nonce survived.
- New `test_expired_nonce_is_deleted` covers the retained delete path.
- `test_can_verify_login_nonce` drops a now-stale "Must create a new one since incorrect nonces deletes them" workaround.

## Stacked on #973

This branches off #973 and reuses its `log_login_nonce_failure()` call sites, so the reason codes stay accurate across the split. **Review and merge #973 first** — the diff shown here includes its commits.

---

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.2%22%2C%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22git%3Adirectory%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Ftwo-factor.git%22%2C%22ref%22%3A%22fix%2Flogin-nonce-preserve-on-mismatch%22%2C%22path%22%3A%22%2F%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


georgestephanis:
Good catch — adopted.

Gating the expiry cleanup on `hashes_match` was wrong, and the case you describe is the common one: someone starts a login, abandons it, and the stale nonce then sits in usermeta until their next password success overwrites it. Worse, every probe against that user logged `mismatch` instead of `expired`, so the reason code was actively misleading about what was in the database.

Checking expiry first is also just simpler, and it means the hash is never computed for a nonce that cannot succeed:

```php
if ( time() >= $login_nonce['expiration'] ) {
    self::log_login_nonce_failure( $user_id, 'expired' );
    self::delete_login_nonce( $user_id );

    return false;
}

$unverified_hash = self::hash_login_nonce( $unverified_nonce );

if ( $unverified_hash && hash_equals( $login_nonce['key'], $unverified_hash ) ) {
    return true;
}

self::log_login_nonce_failure( $user_id, 'mismatch' );

return false;
```

Worth being explicit that this does not weaken what the PR is for. Deleting an **expired** nonce is not destructive — it was already dead, and no request could have used it. The behavior being removed is deleting a **live** nonce because someone presented the wrong value against it, which is what let an unauthenticated request end another user's in-progress login. Expiry cleanup and mismatch preservation are independent, and checking expiry first is what keeps them that way.

Coverage added for exactly the case you named: a new `'past expiration, any value'` data set asserting the reason is `expired` rather than `mismatch` when the key is wrong, and `test_expired_nonce_is_deleted` now runs a second pass with a garbage key to confirm cleanup does not depend on the caller knowing it.

masteradhoc (approved):
LGTM!

